PR #20, opened 2016-01-28
Allow server client requests

aeschli:


daviwil:
Awesome!  I've been hoping this might happen :)

PR #44, opened 2016-06-13
Add ability to change documentSelector dynamically

mrmckeb:
As per discussion here with @dbaeumer:
https://github.com/Microsoft/vscode-eslint/pull/74#issuecomment-223235549

This seems to work during testing locally. Once this is approved (pending changes), I'll update the PR on the ESLint extension to use this instead of restarting the client.

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -430,6 +430,11 @@ export class LanguageClient {
 		}
 	}
 
+	public changeDocumentSelector(documentSelector: string | string[]): void {
+		this._languageOptions.documentSelector = documentSelector;
+		this._syncExpression = this.computeSyncExpression();
+	}
+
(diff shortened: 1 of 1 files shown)


mrmckeb:
Unfortunately this doesn't work (it appeared as though it did, but that was an issue with my compiler locally - I was running an old version).

I've tried a number of ways of achieving this, and I'm somewhat stumped. Are you able to offer some guidance here, @dbaeumer? Any help would be greatly appreciated.

dbaeumer:
@mrmckeb you are on the right path. Since I have a request to support more than 'one' language in the protocol I created a branch to start an initial implementation. It is https://github.com/Microsoft/vscode-languageserver-node/tree/dbaeumer/languageProvider. Let me know if this works for me. Then I will add a first limited support for language providers to the client.

dbaeumer:
@mrmckeb this is now supported in the new version of the library published today. Closing.

mrmckeb:
Thanks @dbaeumer, again great work.

Issue #57, opened 2016-07-14
Move additional types to `vscode-languageserver-types`


david-driscoll:
Would it be possible to move some more interfaces over to types so that they may be shared?
https://github.com/Microsoft/vscode-languageserver-node/blob/master/server/src/protocol.ts#L128

I can understand anything referencing json-rpc shouldn't be shared, but it would be great to see `ServerCapabilities` and friends available for creating implementations using TypeScript.  As it stands I have to make a copy of `protocol.ts`.

It would also be great to have a methods namespace as well, to make those constants and available.

dbaeumer:
@david-driscoll these got move to a separate npm module called vscode-languageserver-protocol. Please ping if this doesn't do the trick for you.

PR #58, opened 2016-07-14
moved more types into common types package

david-driscoll:
This is a proposal for #57.

Feel free to let me know what may / may not need to change.  :tada:

### DIFF
--- a/client/src/protocol.ts
+++ b/client/src/protocol.ts
@@ -18,176 +18,37 @@ import {
 		DocumentHighlight, DocumentHighlightKind,
 		SymbolInformation, SymbolKind,
 		CodeLens, CodeActionContext,
-		FormattingOptions,
-	} from 'vscode-languageserver-types';
-	
-/**
- * A parameter literal used in requests to pass a text document and a position inside that
- * document.
- */
-export interface TextDocumentPositionParams {
-	/**
-	 * The text document.
-	 */
-	textDocument: TextDocumentIdentifier;
+		FormattingOptions, Methods,
+		ClientCapabilities, CodeActionParams, CodeLensOptions,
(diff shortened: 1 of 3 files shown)


hediet:
Can we please have a single source of every type (including request and notification type specifications) that is mentioned in the language server protocol specification and separate them from types not mentioned there?
Right now, it is just incredibly hard to generate code for other languages, as the specification is very inconsistent and thus hard to parse (its typescript even has syntactical errors) and the types here are mixed with implementation types that have nothing to do with the protocol (like `TextEditChange` or `TextDocument`).
Your pull request makes it even harder, as it adds another indirection with all member in the `Methods` class - getting the method out of `{ get method() { return 'textDocument/hover'; } };` is already hard enough when using the typescript parser API.

dbaeumer:
The idea of the types package is to have the types which are independent of the protocol in a separate package. So we should not move protocol specific types in there. @david-driscoll why do you need these types in a separate package. Do you want to implement your own client in TS ?

@hediet the idea for this is to have a JSON schema in the protocol repository which then can be used to generate 'interfaces' for other languages.

@aeschli any additional comment.

david-driscoll:
Yeah, I'm implementing the client for Atom.  Almost all of the types are reusable at the client level.  I hate to have to copy and paste things, but that's what I have at the moment.

aeschli:
+1 for a JSON schema description of all type protocol messages & types.

Maybe we should think of a language-service-module package, containing protocol types/messages plus a Connection implementation that can be used for server and client implementations, without any VSCode specific bindings. That would help David as well as other language client implementations that want to implement a client in NodeJS.

dbaeumer:
All these types are now available in the protocol npm module which is independent of server or client.

PR #59, opened 2016-07-15
Test Protocol client

iyyappam:
This test client is used to validate the communication channel and protocol between this client and language server. The client can send request and notification to server based on the language protocol and expect response for the request. Server can also send specific notification to client. Any custom protocol in addition to the language server protocol can be built on to od existing channel.

Mocha can used to run tests for this protocol client.

### DIFF
--- a/testprotocolclient/.npmignore
+++ b/testprotocolclient/.npmignore
@@ -1,0 +1,9 @@
+.vscode/
+lib/test/
+lib/*.map
+src/
+test/
+.eslintrc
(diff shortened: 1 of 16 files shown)


maxm007:
Hi

Will this pull request be merged at some point?  Could really use this.

Max

dbaeumer:
I will close this PR. Most of it got implemented by having the new npm protocol module which only depends on node and can be integrated into test runners like mocha.

PR #83, opened 2016-08-28
Print child process stderr to output channel

### LINKED ISSUE #82: How to get stderr output of a language server?
@dbaeumer 
I justed implemented my language server and did not find a way to get the stack trace or logs it outputs on stderr when it crashes.

sqs:
Prior to this change, child process stderr was discarded, which makes
it hard to debug language server processes. (You can make them log to
a separate file or use the window/logMessage method, but in general
it's hard to guarantee that all output, including crash stack traces,
goes through a user-defined stream in a program.)

This change appends child process stderr output to the output channel.
Clients can suppress this behavior by setting the server's
Executable.stdio 3rd element to be "ignore" (per the Node.js spec).

Fixes #82

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -980,6 +980,9 @@ export class LanguageClient {
 				return Promise.reject<IConnection>(`Launching server using command ${command.command} failed.`);
 			}
 			this._childProcess = process;
+			if (process.stderr) {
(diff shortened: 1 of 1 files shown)


dbaeumer:
@sqs thanks for the PR.

PR #98, opened 2016-09-17
sendRequest fail if not listening

AArnott:
Rejects calls to `sendRequest` when `listen()` hasn't been called already. This immediately identifies a common mistake for a new _client_ that doesn't realize that `listen()` is important to call, even though no incoming RPC calls are expected.

### DIFF
--- a/jsonrpc/src/main.ts
+++ b/jsonrpc/src/main.ts
@@ -447,6 +447,12 @@ function createMessageConnection<T extends MessageConnection>(messageReader: Mes
 		}
 	}
 
+	function throwIfNotListening() {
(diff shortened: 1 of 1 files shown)


dbaeumer:
Makes sense. Need to understand why a test case fails now. Might call sendRequest without listening.

AArnott:
Woops. I should have run the tests.

dbaeumer:
Since the code changed in this area I replayed the changes by hand.

dbaeumer:
Tests only need to call listen before send :-)

AArnott:
Thanks

PR #110, opened 2016-10-19
refactor: rename directories and add install.sh script

mbana:
Trying to simplify the module(s) to make it easier to use:
1. cleaning up the scripts,
2. renaming dirs,

Todo:
1. reduce size of classes,
2. add samples, along with scripts to start them,

It's not so clear on where to start - hopefully this will improve over time.

Thanks,


dbaeumer:
@mbana I do like contributions but it would be preferable if something like this is discussed in an issue first. What is the benefit of renaming all this. To match the module names. We decided to not to repeat these. If we want to have this then IMO we should move each npm module into its own repository. But again this should be discussed first. 

I feel sorry but I will close the PR. It has too many conflicts and I think if we do change this we should use separate git repositories.

mbana:
ok thanks fine.

PR #112, opened 2016-10-25
export './protocol' from client as Protocol.

mbana:
Would be nice if the definitions of the `Protocol` were available outside of the `Client`.


dbaeumer:
Agree. But this is a little bit more work since parts of the types are now in vscode-languageserver-types and we should not have two definitions. I will clean this up for 3.0.

PR #117, opened 2016-10-27
JSON-RPC over WebSocket: please comment

mbana:
I'd like to go ahead and implement this but before I do that, I'd really appreciate any feedback:
- The different modules need to be refactored a bit to to exclude un-needed transports.
- Somehow coordinate the work?

All in all, if that works, it will allow the browser to communicate with the `server` using the `client` module without much trouble. I hope so :), at least.

---

**Edit:**
See `jsonrpc/scripts/websocket_start.sh` for starting.
If you want to press ahead, I can start making it more prod ready; using Docker for builds etc. so they "always" etc.
Soz, about the Unix scripts. It's my dev machine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+
+# yarn lock file
+yarn.lock
+
 # Logs
 logs
 *.log
--- a/client.tests/package.json
+++ b/client.tests/package.json
@@ -21,7 +21,10 @@
 		"vscode": "^1.0.0",
 		"vscode-languageserver-types": "2.0.0-alpha.1",
 		"@types/node": "^6.0.45",
-		"@types/mocha": "^2.2.32"
+		"@types/mocha": "^2.2.32",
+        "@types/socket.io": "^1.4.27",
+        "@types/socket.io-client": "^1.4.27",
+        "socket.io": "^1.0.6"
 	},
 	"scripts": {
 		"postinstall": "node ./node_modules/vscode/bin/install",
--- a/client/package.json
+++ b/client/package.json
@@ -1,34 +1,35 @@
 {
-	"name": "vscode-languageclient",
-	"description": "VSCode Language client implementation",
-	"version": "3.0.0-alpha.3",
-	"author": "Microsoft Corporation",
-	"license": "MIT",
-	"engines": {
-		"vscode": "^1.6.0"
-	},
-	"repository": {
-		"type": "git",
-		"url": "https://github.com/Microsoft/vscode-languageserver-node.git"
-	},
-	"bugs": {
-		"url": "https://github.com/Microsoft/vscode-languageserver-node/issues"
-	},
-	"main": "./lib/main.js",
-	"typings": "./lib/main",
-	"devDependencies": {
-		"typescript": "^2.0.3",
-		"vscode": "^1.0.0",
-		"@types/node": "^6.0.42"
-	},
-	"dependencies": {
-		"vscode-jsonrpc": "^3.0.0-alpha.2",
-		"vscode-languageserver-types": "^2.0.0-alpha.1"
-	},
-	"scripts": {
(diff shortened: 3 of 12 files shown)


dbaeumer:
I like the idea of having more transports, but instead of adding them all to the JSON-RPC npm libnrary I would suggest that we make the transport pluggable and let users control this from the outside. The json-rpc library already supports this. See: 

```ts
export function createMessageConnection(reader: MessageReader, writer: MessageWriter, logger: Logger): MessageConnection;
```

The only thing that is necessary is to support this in the client as well. Would you be willing to look into this. The web socket transport should then go into its own npm module.

mbana:
thanks for the feedback.

when i implemented it, i saw that various readers and writers were defined in `vscode-jsonrpc`. so i followed the convention.

yes, i'm at the moment making the `monaco-go` talk language server protocol via a websocket. i'll post an update when i have a _draft_ version working. the `vscode` version  roughly works, i haven't had a chance to test it thoroughly as i am concerned with `monaco` support.

here's what i am thinking in terms of implementation:

* five packages `vscode-languageclient` (node impl) & `vscode-jsonrpc` (node impl) `vscode-languageclient-browser` & `vscode-jsonrpc-browser`, `vscode-languageclient-abstract` & `vscode-jsonrpc-abstract`
* The node impl will marshall what it can to the `vscode-languageclient-abstract` & `vscode-jsonrpc-abstract` so the `*-browser` can reuse what's needed. existing code won't break, i hope, since the package name will still be the same.
* `vscode-languageclient-abstract` & `vscode-jsonrpc-abstract` will contain most of the implementation unless something platform-specific is required.

i think it's easier to see it working first then get feedback?

dbaeumer:
Let me think about this. I think with TS having structural typings we don't need so many new projects / npm modules. I think we can make that work easier.

mbana:
@dbaeumer, hi,

if you can please take a look at this, https://github.com/mbana/vscode-languageserver-node/commit/683eaf95fff0c526de504411be8bf2185cc9f383, commit and possibly the one before it.

I've done the following:

* `vscode-jsonrpc-common`: new package with all the common code.
* `vscode-jsonrpc`: the node impl, the default one, imports what it needs and defines new readers and writers.
* `vscode-jsonrpc-browser`: like above, but adds the websocket readers and writers.

compiles fine and tests pass.

i'll do likewise for `vscode-languageclient` tomorrow.

i have local `Dockerfile`s for building all the libs in an isolated container. i haven't tested them enough though, so this will have to wait.

dbaeumer:
@mbana I am very busy getting the 3.0 version of the libraries out and I will not have time right now to look at the changes. Sorry. I did quickly look at our changes and the look massive to me. They even start to pull in things (like yarn, tslint, ....) which can't come in via a simple PR. 

I am still very confident that we can come up with something easier than common, node and browser version. Since TS is structural typed we don't need common super types. 

I would suggest that you open a regular issue and then we discuss how we can break this down in simpler steps.

mbana:
@dbaeumer, :), pls. do let me know when u're free.

* do you know of any good examples that make use of the structural typing that i could look at.
* i'll start with the jsonrpc package if that's ok with you? or do you have another suggestion?

would this allow one to simply include the `vscode-languageclient` without the module trying to load native node modules, say, child_process, or the user to browersify the package with shims.

there is a build step that uses requirejs in https://github.com/Microsoft/monaco-css and it has a few restrictions, such as es5 only - `vscode-languageclient` uses es6. i don't really what to modify the monaco-* build files.

dbaeumer:
A good explanation of structural typing can be found here: https://www.typescriptlang.org/docs/handbook/type-compatibility.html. It basically says that two type are assignment compatible if their members are assignment compatible even if they don't stand in an inheritance relationship.

Issue #88, opened 2016-09-03
Monaco integration


akosyakov:
I want to connect monaco editor with a language server by means of this project.

I've managed to make use of `jsonrpc` module to create a client connection and hoped to reuse code and protocol converters, but it seems that types expected by vscode and monaco are not compatible, e.g. monaco.IRange vs vscode.Range.

I'm aware that monaco is assembled from vscode. Would it be possible to eliminate the difference?

dbaeumer:
@akosyakov in principal yes, however this is quite some work. Besides different names there are semantic differences as well (that is why the names are different at the end). For example in Monaco position informations are one based, whereas in the VS Code API they are zero based.

georgewfraser:
It looks like [sourcegraph](https://sourcegraph.com) is using the language server protocol with monaco.

mbana:
would be awesome if web sockets were used. it might then be easy to make a pure brower-based client without much trouble, i hope?

akosyakov:
@mbana i would prefer that transport should be interchangeable, as actually it is a case with vscode integration.

dbaeumer:
OK. The right approach would be to make the converters pluggable into the language client. There is already an interface for this in the corresponding file (e.g. Converter interface type). The converters also expose old legacy functions. When we make this pluggable we should remove the old function and break (e.g. version 3.0).

mbana:
@dbaeumer and @akosyakov,

so i tried out another approach, which although quite painful requires little code change in modules within `vscode-languageserver-node`. At the moment it's making a connection, see below, I just need to keep stubbing out more code.

```sh
LanguageClient:langserver-antha -  [Trace - 9:21:35 PM] Received response 'initialize - (0)' in 16ms. Request failed: invalid root path "": must be file:/// URI (0).
(index):137 LanguageClient:langserver-antha -  [Error - 9:21:35 PM] Server initialization failed.
(index):137 LanguageClient:langserver-antha -  Error: invalid root path "": must be file:/// URI
    at new ResponseError (http://127.0.0.1:8080/node_modules/vscode-languageclient/node_modules/vscode-jsonrpc/lib/main.js:371:32)
    at handleResponse (http://127.0.0.1:8080/node_modules/vscode-languageclient/node_modules/vscode-jsonrpc/lib/main.js:988:48)
[...]

PR #118, opened 2016-10-29
Refactor client

mbana:
Refactoring the client a bit
- moving classes to separate files.
- created classes for the `Converter`s.

We're also doing a `WebSocketReader`. If my employer is happy to push the changes back, I'll create a PR for that as well. That said, the `client` lib will be structured differently though so as to only allow the importer to choose a specific set of transports required.

Many thanks,


akosyakov:
@mbana Is it about Monaco integration: https://github.com/Microsoft/vscode-languageserver-node/issues/88? How does it help? I see that converters still depend on vscode types: https://github.com/Microsoft/vscode-languageserver-node/pull/118/files#diff-e05feeb2a7817f9fe64dd047e2c12b00R1.

Is not having `WebSocketReader` and `WebSocketWriter` enough to change transport? Why not?

dbaeumer:
@mbana I agree with @akosyakov: why does this help and what is the improvement. The client already exports both converters as properties.

mbana:
i was trying to understand the structure of the various projects and along the way i started to move classes etc. to separate files. i guess it saves time by having most of the implementation in one file?

`vscode-languageserver-types`: not sure i understand. why is this an issue? the package seems mostly interfaces... please explain.

dbaeumer:
@mbana What do you mean by

> vscode-languageserver-types: not sure i understand. why is this an issue? the package seems mostly interfaces... please explain.

mbana:
i wasn't sure what @akosyakov  meant.

that said, why this an issue? `vscode-languageserver-types` doesn't include `vscode` so that's fine. it's only `vscode-languageclient` that does which is understandable - it marshalls to and fo a LSP. couldn't every LSP client need to do this?

PR #121, opened 2016-11-09
Fixes #120: CancellationToken.is does not check for undefined

### LINKED ISSUE #120: bad jasonrpc shutdown: CancellationToken.is breaks on undefined token
jasonrpc: [CancellationToken.is](https://github.com/Microsoft/vscode-languageserver-node/blob/master/jsonrpc/src/cancellation.ts#L40) does not check for `undefined` before dereferencing a property, while languageclient [calls `shutdown`](https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/main.ts#L187) with an `undefined` token:
```typescript
shutdown: () => connection.sendRequest(ShutdownRequest.type, undefined),
```

siegebell:


dbaeumer:
This is already fix in master using the following code:

```ts
	export function is(value: any): value is CancellationToken {
		let candidate = value as CancellationToken;
		return candidate && (candidate === CancellationToken.None
			|| candidate === CancellationToken.Cancelled
			|| (Is.boolean(candidate.isCancellationRequested) && Is.defined(candidate.onCancellationRequested)));
	}

```

dbaeumer:
Thanks for the test case which I merged in by hand.

PR #123, opened 2016-11-15
Fix Client.stop promise

kentor:
The promise should wait for `connection.shutdown()`, not just
`resolveConnection()`. Also flatten the promise to match the type
signature of the method.

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -932,17 +932,19 @@ export class LanguageClient {
 		this.state = ClientState.Stopping;
 		this.cleanUp();
 		// unkook listeners
-		return this.resolveConnection().then(connection => {
-			connection.shutdown().then(() => {
-				connection.exit();
-				connection.dispose();
-				this.state = ClientState.Stopped;
-				this._connection = null;
-				let toCheck = this._childProcess;
-				this._childProcess = null;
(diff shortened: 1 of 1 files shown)


kentor:
not sure why test is failing

mbana:
Yeh, I've noticed the CI system is not compiling the PR but rather the master branch.
In one of my PRs I moved the err'ing file, `messageReader.ts`, elsewhere yet I still get a similar error in my PR: https://travis-ci.org/Microsoft/vscode-languageserver-node/builds/175949270. I guess this (not compiling the PR) was intentended.

I guess something like below might help:

* Reproduce build env locally. If you figure out how pls do share.
* Open each project as separate windows in vscode, this way build errors are shown.

dbaeumer:
This is fixed on master in the following way:

```ts
		return this.resolveConnection().then(connection => {
			return connection.shutdown().then(() => {
				connection.exit();
				connection.dispose();
				this.state = ClientState.Stopped;
				this._connection = null;
				let toCheck = this._childProcess;
				this._childProcess = null;
				// Remove all markers
				this.checkProcessDied(toCheck);
			})
		});
```

Any objection to close this PR

kentor:
👍 either way works

PR #124, opened 2016-11-15
Fix compile errors

mbana:


dbaeumer:
@mbana there is not need to add ./node_modules/ to scripts. npm doesn this automatically. 

And I don't see any compile errors on master which would require your code changes. Can you please explain why you are suggesting those.

I like your script. However to merge this in I need Windows scripts as well :-)

mbana:
ok thanks. i was just being careful; i think i may have run into an issue in the past with a global installation overriding the package-specific one. soz, please close then.

compile errors:
i'm sure i was getting compile errors are those lines.
in the future, i will capture the error when raising PRs.

build:
can i kindly suggest compiling in Docker - since they are no native dependencies required? - and running the tests on the various platforms. you probably already do this.
any chance of adding pre-commit hooks, that way compile errors (in theory) in master won't happen.

thanks,

mbana:
errors i got before the fix are below.
any idea what's up?

```sh
file: 'file:///Users/mbana/dev/vscode/mbana/vscode-languageserver-node/client/src/main.ts'
severity: 'Error'
message: 'Type 'Promise<{}>' is not assignable to type 'Thenable<void>'.
  Types of property 'then' are incompatible.
    Type '{ (onfulfilled?: (value: {}) => {} | PromiseLike<{}>, onrejected?: (reason: any) => {} | PromiseL...' is not assignable to type '{ <TResult>(onfulfilled?: (value: void) => TResult | Thenable<TResult>, onrejected?: (reason: any...'.
      Types of parameters 'onfulfilled' and 'onfulfilled' are incompatible.
        Type '(value: void) => any' is not assignable to type '(value: {}) => {} | PromiseLike<{}>'.
          Types of parameters 'value' and 'value' are incompatible.
            Type '{}' is not assignable to type 'void'.'
at: '1580,10'
source: 'ts'
```

```sh
file: 'file:///Users/mbana/dev/vscode/mbana/vscode-languageserver-node/client/src/main.ts'
severity: 'Error'
message: 'Type 'Promise<{}>' is not assignable to type 'Thenable<void>'.'
at: '1593,10'
source: 'ts'
```

LN 1579-160:

```javascript
	private handleRegistrationRequest(params: RegistrationParams): Thenable<void> {
		return new Promise((resolve, reject) => {
			params.registrations.forEach((element) => {
				const handler = this._registeredHandlers.get(element.method);
				element.registerOptions.documentSelector = element.registerOptions.documentSelector || this._clientOptions.documentSelector;
				if (handler) {
					handler.register(element);
				}
			});
			resolve();
		})
	}

	private handleUnregistrationRequest(params: UnregistrationParams): Thenable<void> {
		return new Promise((resolve, reject) => {
			params.unregisterations.forEach((element) => {
				const handler = this._registeredHandlers.get(element.method);
				if (handler) {
					handler.unregister(element.id);
				}
			});
			resolve();
		})
	}
```

dbaeumer:
There are automatic builds happen on every checkin. They are here: https://travis-ci.org/Microsoft/vscode-languageserver-node

I am not sure why the build fails for you. I assume an older version of a TS compiler.

Issue #86, opened 2016-09-01
Treat null as undefined


felixfbecker:
JavaScript is very special with the seperation between `null` and `undefined`. In many other languages such as PHP, if a property has not been set, it is `null`. Consequently, not-set properties will also JSON-encode to `null`. There is no way to find out if a property is `null` because it has been set to `null` or because the property was not set.

My PHP language server responds with this response to the `initialize` request:
```json
Content-Type: application/vscode-jsonrpc; charset=utf8
Content-Length: 493

{
   "id":0,
   "result":{
      "capabilities":{
         "textDocumentSync":1,
         "hoverProvider":null,
         "completionProvider":null,
         "signatureHelpProvider":null,
         "definitionProvider":null,
         "referencesProvider":null,
         "documentHighlightProvider":null,
         "documentSymbolProvider":true,
         "workspaceSymbolProvider":null,
         "codeActionProvider":null,
         "codeLensProvider":null,
         "documentFormattingProvider":null,
         "documentRangeFormattingProvider":null,
         "documentOnTypeFormattingProvider":null,
         "renameProvider":null
      }
   },
   "error":null,
   "jsonrpc":"2.0"
}
```
For a human reading this, it is clear that the request was successful, `result` is defined and `error` is not defined. But the language client explicitely checks if `error !== undefined` [here](https://github.com/Microsoft/vscode-languageserver-node/blob/master/jsonrpc/src/main.ts#L297). This makes VS Code treat the request as a failed, and then fail with `Cannot read property 'code' of null`, because it tries to read the `code` property of the error.

It makes it impossible to implement the language server protocol with classes in any language that does not have a dedicated `undefined` type. Please just treat `null` as `undefined`.

`value != undefined` will be `true` for both `null` and `undefined`.

dbaeumer:
@felixfbecker we do treat undefined and null 'equal' in the protocol.I even think that I state somewhere that you should send null not undefined since undefined doesn't exist in other languages. On our server side we even have code that explicitly exchanges undefined with null. And the client code always treats undefined and null as 'equal'

However in the case you pointed out I explicitly tried to followed the JSON RPC spec which states:

```
result
   This member is REQUIRED on success.
   This member MUST NOT exist if there was an error invoking the method.
   The value of this member is determined by the method invoked on the Server.
error
   This member is REQUIRED on error.
   This member MUST NOT exist if there was no error triggered during invocation.
[...]

PR #128, opened 2016-11-19
Treat null as undefined

### LINKED ISSUE #127: Cannot read property 'range' of null: CompletionItem.textEdit cannot be null
The protocol defines that all properties on `CompletionItem` except `label` are optional. However, sending a completion item with only a `label` results in this exception:
```
Cannot read property 'range' of null: TypeError: Cannot read property 'range' of null
    at asTextEdit (C:\Users\felix\git\OpenSource\vscode-php-intellisense\node_modules\vscode-languageclient\lib\protocolConverter.js:108:46)
    at C:\Users\felix\git\OpenSource\vscode-php-intellisense\node_modules\vscode-languageclient\lib\protocolConverter.js:101:67
    at set (C:\Users\felix\git\OpenSource\vscode-php-intellisense\node_modules\vscode-languageclient\lib\protocolConverter.js:89:13)
    at asCompletionItem (C:\Users\felix\git\OpenSource\vscode-php-intellisense\node_modules\vscode-languageclient\lib\protocolConverter.js:101:9)
    at Array.map (native)
    at asCompletionResult (C:\Users\felix\git\OpenSource\vscode-php-intellisense\node_modules\vscode-languageclient\lib\protocolConverter.js:82:26)
    at process._tickCallback (internal/process/next_tick.js:103:7)
```
That is because [here](https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/protocolConverter.ts#L182) `textEdit` is passed to `asTextEdit`, which passes `textEdit.range` to `asRange`. `textEdit` can be `null` though, which leads to the exception.

The same happens with `asCommand`.

I highly recommend you to take a look at `strictNullChecks` - it prevents these bugs at compile-time!

felixfbecker:
Fixes #127

### DIFF
--- a/client/src/protocolConverter.ts
+++ b/client/src/protocolConverter.ts
@@ -164,7 +164,7 @@ export function createConverter(uriConverter?: URIConverter): Converter {
 	}
 
 	function set<T>(value: T, func: () => void): void {
-		if (is.defined(value)) {
+		if (is.defined(value) && !is.nil(value)) {
 			func();
 		}
 	}


dbaeumer:
This is IMO not a correct fix for the  problem. If the value is null the property should be present. This is how non JS language work and the protocol defines it that way.

I will look into #127 separately. Any objection to close.

felixfbecker:
> If the value is null the property should be present. This is how non JS language work and the protocol defines it that way.

In non-JS languages the protocol structures are typically implemented as a class, and if the property is `null`, it means the absence of a value. 
Maybe I don't understand you `set()` helper function, but as I understand it, it conditionally calls the provided callback only if the passed value is "there". Your usage suggests that it is used to check first if a value for the property was provided, and only then convert that value to your custom classes. So, if you want to treat `null` as meaning the absence of a value, just like `undefined` (you said so [here](https://github.com/Microsoft/vscode-languageserver-node/issues/86#issuecomment-245612694)), the `set()` function should check if the property is `null` too.

Again - sorry if I misunderstood the use of the helper function, I though this was the most generic way to solve it, because it doesn't only appear with `TextEdit`, but also with `Command` etc.

dbaeumer:
@felixfbecker I reworked all this for the upcoming 3.0 release of the node libraries. The set function is gone and I try to be more preceise with null and undefined. I even enable strictNull and documented the API with undefined | null if allowed.

For converting protocol objects into code objects a null value now gets mapped to undefined in this direction since VS Code's API accepts both and it makes the code more clear. However on the protocol layer itself I sometimes explicitly set thing to null if the spec requires this. For example a response must either contain a result or an error. If a request does return undefined in a node server I will set the result to null to honor this.

I will close the PR since the set function is gone.

PR #148, opened 2017-01-22
vscode-languageclient: make LanguageClient platform-independent

sqs:
Platform-specific routines remain in main.ts, but consumers can now import client.ts without requiring any Node or Electron dependencies. This makes it possible to use LanguageClient in platforms other than Node+Electron.

Consumers of client.ts merely need to provide a `() => Thenable<{reader: StreamMessageReader, writer: StreamMessageWriter}>`.

The main.ts API remains unchanged and should be 100% backward-compatible. The client.tests tests pass.

The large size of this diff is misleading. The changes were in 2 steps:

1. Move the LanguageClient from main.ts to a new file client.ts
2. Create a LanguageClient in main.ts that extends the "base" LanguageClient in client.ts, and move Node/Electron stuff from client.ts to overrided methods in main.ts's LanguageClient

No new "real" code was added.

I would love to get feedback on this diff early, so I created this PR. Happy to go back to the drawing board if this is not something you are willing to merge.


client/src/client.ts
```diff
@@ -0,0 +1,2022 @@
...
+			this._connectionPromise = this.createConnection();
+		}
+		return this._connectionPromise;
+	}
+
+	private initialize(connection: IConnection): Thenable<InitializeResult> {
+		this.refreshTrace(connection, false);
+		let initOption = this._clientOptions.initializationOptions;
+		let initParams: InitializeParams = {
+			processId: process.pid,
```
felixfbecker: If I understand you correctly, client.ts is supposed to work without Node, right?
In that case this must be something like
```ts
processId: typeof process !== 'undefined' ? process.pid : undefined`
```
`processId` is optional in the LSP spec.

Probably even better would be to pass that option to the constructor, and let the `main.ts` Client set default it to `process.pid`.

felixfbecker:
Btw, allowing to pass Promises/Thenables as parameters is considered an antipattern. It just makes code more complicated, the _callee_ should let the promise resolve before calling the function.

client/src/client.ts
mbana: i am very much in favour of this PR.
for the tests, maybe, try to bootup the new client in the browser so see if it loads.
that way you might be able to errors like the above fairly early.

dbaeumer:
@sqs nice work. Could you look into signing the CLA ?

sqs:
@dbaeumer Done, thanks!

client/src/client.ts
sqs: Thanks, I forgot to remove this. (I do have it running in the browser successfully, but I configured webpack to inject a global `process` definition, so I missed this.)

msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

client/src/client.ts
```diff
@@ -1146,7 +1146,7 @@ export class LanguageClient {
 		this.refreshTrace(connection, false);
 		let initOption = this._clientOptions.initializationOptions;
 		let initParams: InitializeParams = {
-			processId: process.pid,
+			processId: typeof process !== 'undefined' ? process.pid : 0,
```
felixfbecker: Should be `null`, not `0`
https://github.com/sourcegraph/language-server-protocol/blob/master/protocol.md#initialize

sqs: The LSP spec says `null` is OK, but the TypeScript type definition forbids `null`. I will leave it as `0` for now so it compiles here.

sqs:
@dbaeumer The only outstanding thing I'm aware of is the InitializeParams `processId`. This PR contains this change:

```diff
-                       processId: process.pid,
+                       processId: typeof process !== 'undefined' ? process.pid : 0,
```

I can either change the TypeScript definition of InitializeParams to allow `processId: null` (which conforms to the LSP spec), or leave this at 0. I have no preference.

Otherwise, let me know if you'd like to see any changes. Thanks! :+1:

dbaeumer:
@sqs I would change the protocol.ts to have | null for the `processId`. A couple of other comments:

- I would name `streamOpener` `streamProvider`. This is more in line with the naming we have in VS Code.
- I would name `LanguageClient` `BaseLanguageClient` in client.ts and make it abstract
- I would add a file `browser.ts` (I am open for a different name. Agree that browser is not optimal) which exports a `LanguageClient` extending BaseLanguageClient.  

This better ensures that we can further develop that design/

And again thanks for the work you are doing here.

dbaeumer:
@sqs: there seems to be another bigger thing to do here. The json-rpc library currently used by the client.ts file heavily depends on node as well. So we also need to restructure the json-rpc library to ensure that the client part is not relying on any node specific code through the json-rpc library.

felixfbecker:
@dbaeumer I implemented the first three changes. I am not to sure what benefit a browser.ts would provide if it is just this?
```ts
export class BrowserLanguageClient extends LanguageClient {}
```
Let me know if you still want me to add it.

I would look at jsonrpc next and do a separate PR for that if that is okay.

dbaeumer:
@felixfbecker thanks for continuing on this. 

I would still introduce a second file `browser.ts` (as said I am open for a different name). Otherwise non npm clients need to instantiate BaseLanguageClients which looks strange IMO.

felixfbecker:
@dbaeumer done

rothfels:
@felixfbecker browser.ts imports main.ts, which depends on the `child_process` module. You'll need to rework that.

mbana:
is webpack less strict than requirejs?
with requirejs it traverses the entire import tree. if something tries to import something it doesn't know about it complains - `child_process`.
globals, need to be treated differently.

do you have a special build config? would be nice to take a look.

felixfbecker:
Sorry, this was a typo. browser.ts was supposed to reexport client.ts, not main.ts. Fixed now.

felixfbecker:
@dbaeumer anything else?

dbaeumer:
@felixfbecker sorry, I was busy with some emergency fixes and will be out on vacation next week. Will comment on the PR when I am back. I you don't hear anything please ping.

felixfbecker:
ping @dbaeumer

dbaeumer:
@felixfbecker thanks for the ping. The thing I am not sure about yet is if there is work needed on jsonrpc as well. vscode-languageclient depends on vscode-jsonrpc which is highly node dependent. When you will use vscode-languageclient outside of node how do you guys consume this. Will the dependency be ignored? I just want to make sure we do something and it will have ripples down to vscode-jsonrpc which we then need to ripple in the client again.

dbaeumer:
One additional thing I stumbled over. The Language Client heavily depends on VS Code API as well. How would the client be used outside an Node / Electron app. VS Code API fully requires Node / Electron.

felixfbecker:
@dbaeumer vscode-jsonrpc only pulls in types from node afaik, not values. This PR makes it significantly easier to use in a non-node environment and works for our needs (we have this branch working in the browser with VS Code without Electron/Node). Other use cases might require even more decoupling but we can address that when it comes

dbaeumer:
I am still a little puzzled here: the client always supported to pass in a function as a server to create the server process. In this case no child_process code was executed. Would there have been an easier solution then to make it work in your environment. Looks like you only want to avoid executing node code.

felixfbecker:
@dbaeumer It would not _execute_ child_process code, but the child_process module was still imported and used (eg the _value_ was imported and referenced in the code, not only the type).

mbana:
@felixfbecker as mentioned that's what ends up happening. a bit confusing at first - i agree.
what i ended up doing is stubbing all the imports and their functions.

i wonder if there is a tool that would actually do this automatically, now that i come to think of it.
it seems mechincal: look all the `import` and find the browser based version, if not, stub all the functions of the module.

irc, one needs to be double-check, even if the import fails the script (in the browser) might still work. however, i guess its better to fix this...

felixfbecker:
@mbana Browserify will automatically give you browser shims for all kinds of Node modules like `url`, `events`, `http`, `buffer` - maybe even `child_process` with WebWorkers. But these shims are often huge, and it is much better to actually provide platform-independent modules

dbaeumer:
@felixfbecker but there is a lot of other code that imports node code and references it. In addition there is all the VS Code specific code that it executed (all the type conversion code). What if VS Code decides to reference child process in its code which I reference through the converters. 

Don't get me wrong I am all in for making the library more useful in other contexts but I am not sure I understand the goal right now. Can you give me an example where the `client/src/client.ts` is used given all the VS Code and other node dependencies this file drags in.

felixfbecker:
@dbaeumer you can go to https://sourcegraph.com/github.com/Microsoft/vscode-languageserver-node/-/blob/client/src/main.ts right now and it is using this branch :)
If you open the dev tools and search for `lsp` in the network tab you will find a WebSocket connection that speaks LSP to a TypeScript language server on the backend.

The goal of this PR is not to remove the dependency on VS Code, it is just to make it easier to run it in a browser.

If VS Code decides to reference child_process somewhere we modify it to use WebWorkers instead :)

sqs:
Hi @dbaeumer, what can we do to improve this PR? This patch is working well in production and makes it much easier to use the library in other contexts. Is @felixfbecker's example sufficient? Thanks!

dbaeumer:
We had endgame the last week and I needed to implement features for VS Code tasks. This is why the PR is on hold right now.

Here is what I think should be done since this is not really making the library platform independent:

- have a client file with basically the code as proposed with an abstract LanguageClient
- have a main file to not break anything

DO NOT include a browser.ts file since this gives the impression that the library actually supports running in the browser and we would need to fix things if this is broken.

The current PR has merge conflicts. If one of you can make the changes then I will merge them in. Otherwise I will have a look in the next 1 - 2 weeks.

sqs:
@dbaeumer Thanks for the pointers. I just updated the PR.

dbaeumer:
Tried to merge but I already changed the code (again :-). So I went down the path of doing the change myself on master. It is basically what you guys suggested with some tweaks:

- in client the class is called BaseLanguageClient and is abstract.
- since we need subclasses anyways I made the creation of the reader/writer a method to implement instead of a function passed to the BaseLanguageClient.
- main.js behaves as before.

I pushed the change to master. @sqs can you please have a look if this satisfies your needs. Please open separate issues if not

PR #160, opened 2017-02-10
Enable source maps

nicksnyder:
For debugging :)

### DIFF
--- a/client/src/tsconfig.json
+++ b/client/src/tsconfig.json
@@ -3,7 +3,7 @@
 		"target": "es6",
 		"module": "commonjs",
 		"moduleResolution": "node",
-		"sourceMap": false,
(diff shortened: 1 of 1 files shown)


dbaeumer:
@nicksnyder I didn't enable this since it would require to may be publish the source maps and the corresponding TS files. Are you still interested in this ?

nicksnyder:
I was able to build everything locally so I personally don't need this anymore. I am don't know if it is normal for libs to publish source maps or not so I will leave it up to you

dbaeumer:
I will discuss this with other devs. I know that in the past this caused problems during debugging if source maps are present but the original source is not. But now that we can inline the source into the source map things might be easier to handle.

PR #163, opened 2017-02-16
MessageBuffer.append: length of string in bytes

dgreisen:
messageBuffer.append creates a buffer based on the length of the string to be written to it. `str.length` does not return the correct length of the string in bytes when the string contains multi-byte unicode characters. Buffer.byteLength does return the proper length, so switched to this method.


dbaeumer:
Good catch.

PR #164, opened 2017-02-17
Support reference streaming

nicksnyder:
This PR implements the necessary changes in vscode-languageclient and vscode-jsonrpc to support streaming the response for "Find all references" requests as described in https://github.com/Microsoft/vscode/issues/20010 (please read this issue for complete context).

This PR depends on (and should not be merged before):
- [ ] LSP protocol proposal https://github.com/Microsoft/language-server-protocol/pull/182
- [ ] vscode.d.ts proposal https://github.com/Microsoft/vscode/pull/20346 (required before CI can pass)

**Implementation details**

The vscode-jsonrpc layer applies JSON patches and forwards the cumulative partial result so higher layers don't need to care about the JSON patch format.

Methods in vscode-languageclient may choose to forward this as-is or perform performance optimizations to only send new data since the data gets serialized over RPC. For streaming references I implemented this performance optimization.

Todos:
- [ ] Use capabilities as defined in spec
- [ ] Update implementation to expect null final result
- [ ] Forward JSON patch all the way to UI?
- [ ] Don't forward incomplete locations

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -10,7 +10,7 @@ import ChildProcess = cp.ChildProcess;
 import {
 	workspace as Workspace, window as Window, languages as Languages, commands as Commands,
 	TextDocumentChangeEvent, TextDocument, Disposable, OutputChannel,
-	FileSystemWatcher, DiagnosticCollection, Uri,
+	FileSystemWatcher, DiagnosticCollection, Uri, ProgressCallback,
 	CancellationToken, Position as VPosition, Location as VLocation, Range as VRange,
 	CompletionItem as VCompletionItem, CompletionList as VCompletionList, SignatureHelp as VSignatureHelp, Definition as VDefinition, DocumentHighlight as VDocumentHighlight,
 	SymbolInformation as VSymbolInformation, CodeActionContext as VCodeActionContext, Command as VCommand, CodeLens as VCodeLens,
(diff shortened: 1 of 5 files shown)


dbaeumer:
First quick question: why do we need JSON patch for this. If It is real streaming then there is no need to patch something. I would rather prefer not to pull in a library here.

nicksnyder:
The LSP proposal has an explanation so if you disagree, let's discuss there:
https://github.com/Microsoft/language-server-protocol/pull/182

Can you clarify why you object to a dependency? Even if we chose to stream parse JSON, I would want to pull in a library for that.

If you don't want any dependencies at all then we would need to implement one of these things (JSON patch or a streaming parser). Anyway, the reasons why we propose JSON patch can be discussed in the issue above.

dbaeumer:
I will be out next week on vacation. Will comment when I am back. Please ping if you don't hear anything.

dbaeumer:
I am not against a dependency also managing it is always overhead. 

I see the technical beauty of a patch but from an application point of view I don't see a benefit (yet). Patching means that I patch something that I already have. If I think about find references then VS code will receive n matches and hands them off to the UI. Then another batch and hands them off. Why would we patch the original first response with additional items. Re-handing them off to the UI simply causes more work for the UI plus memory overhead. But could be I am missing something :-) My brain is already half on vacation.

nicksnyder:
Enjoy your vacation!

> If I think about find references then VS code will receive n matches and hands them off to the UI. Then another batch and hands them off. Why would we patch the original first response with additional items.

This is an approach that we considered and would certainly work.

There are two benefits of JSON patch

1. This streaming proposal will work for any endpoint without subsequent changes to LSP
2. There may be objects where the server can send back useful data for the UI to display and then later update the object with information that is only needed on click (and more expensive to compute). For example, the UI might need name/kind for SymbolInformation but not location until the user actually clicks.

Would rather go down the simpler but less generic path and implement streaming for LSP methods one at a time?

> Re-handing them off to the UI simply causes more work for the UI plus memory overhead. But could be I am missing something.

Sure, the UI thread has to more work to re-render the UI more frequently, but we think the work is worth it for the user to see results sooner.

cc @felixfbecker

PR #167, opened 2017-02-17
Fix jsonrpc README example

### LINKED ISSUE #166: Params in vscode-jsonrpc README example is not working
```
yarn install vscode-jsonrpc
```

```js
// test.js
var rpc = require('vscode-jsonrpc');
  
let connection = rpc.createMessageConnection(
  new rpc.StreamMessageReader(process.stdin),
  new rpc.StreamMessageWriter(process.stdout)
);

var cp = require('child_process');
let childProcess = cp.spawn('node', ['dummy.js']);
let notification = { method: 'format' };
connection.listen();
connection.sendNotification(notification, 'Hello World');
```

```js
// dummy.js
process.stdin.pipe(process.stdout);
```

```js
$ node test.js
Content-Length: 47

{"jsonrpc":"2.0","method":"format","params":[]}
```

As you can see, the params array is empty. I'm still debugging what's going on but just wanted to let you know.

vjeux:
Fixes #166

### DIFF
--- a/jsonrpc/README.md
+++ b/jsonrpc/README.md
@@ -20,11 +20,9 @@ let connection = rpc.createMessageConnection(
 	new rpc.StreamMessageReader(childProcess.stdout),
 	new rpc.StreamMessageWriter(childProcess.stdin));
 
-let notification: rpc.NotificationType<string> = { method: 'testNotification' };
-
 connection.listen();
 
-connection.sendNotification(notification, 'Hello World');
+connection.sendNotification('testNotification', 'Hello World');
 ```
 
 The server side looks very symmetrical:


nmote:
I just encountered the same issue with the example code, but found another fix that works:

`let notification = { method: 'testNotification', numberOfParams: 1 };`

Not sure which one is preferred.

dbaeumer:
Actually the right fix here is to use the new classes instead of object literals.
```ts
let notification = new NotificationType<string, void>('testNotification');
```

I will make the necessary changes.

nmote:
What if you are using this with stock Node, not TypeScript?

dbaeumer:
If you use it with plain JS then there is no need to define a NotificationType. You can directly send a message using connection.sendNotification('testNotification', ....);

PR #172, opened 2017-02-21
client: Pass all arguments to ExecuteCommandRequest

smarter:
Previously, only the first argument was passed.


dbaeumer:
Good catch.

Issue #179, opened 2017-03-03
nodePath not working (vscode-tslint)


maxdeviant:
You can read the original issues description in [vscode-tslint #173](https://github.com/Microsoft/vscode-tslint/issues/173).

Whenever I try to set a `nodePath` for vscode-tslint, it always ends up using the global node_modules path.

You can see that even though the NODE_PATH has a value, it still continues on to call `resolveGlobalNodePath`.

```
[Trace - 10:24:16 AM] NODE_PATH value is: c:\Projects\CloudTestGrading\node_modules\@remark\typescript\node_modules
[Trace - 10:24:17 AM] 'npm config get prefix' value is: C:\Program Files\nodejs
[Trace - 10:24:17 AM] NODE_PATH value is: C:\Program Files\nodejs\node_modules
```

From [`server/src/files.ts`](https://github.com/Microsoft/vscode-languageserver-node/blob/master/server/src/files.ts#L206-L223):

```ts
export function resolveModulePath(workspaceRoot: string, moduleName: string, nodePath: string, tracer: (message: string, verbose?: string) => void): Thenable<string> {
	if (nodePath) {
		if (!path.isAbsolute(nodePath)) {
			nodePath = path.join(workspaceRoot, nodePath);
		}
		return resolve(moduleName, nodePath, nodePath, tracer).then((value) => {
                        // This line seems to be the culprit
			if (value.indexOf(path.normalize(nodePath)) === 0) {
				return value;
			} else {
				return Promise.reject<string>(new Error(`Failed to load ${moduleName} from node path location.`));
			}
		}).then(undefined, (_error: any) => {
                        // We are passing a custom nodePath that _should_ be valid,
                        // so we do not expect this to be called.
			return resolve(moduleName, resolveGlobalNodePath(tracer), workspaceRoot, tracer);
		});
	} else {
		return resolve(moduleName, resolveGlobalNodePath(tracer), workspaceRoot, tracer);
	}
}
```

maxdeviant:
##### Environment Info:
**OS:** Windows 7
**Node:** v7.7.1
**VSCode:** v1.10.1
**vscode-tslint:** v0.8.1

maxdeviant:
I have been trying to assemble a test case to illustrate the problem, but unfortunately there isn't any test coverage for the `server` code, and it looks like all of the resolve functions are pretty tightly coupled to the outside world.

dbaeumer:
[...]

Issue #173, opened 2017-02-22
WorkspaceEdit is incompatible with the protocol


KalitaAlexey:
https://github.com/Microsoft/vscode-languageserver-node/blob/master/types/src/main.ts#L366
```ts
export interface WorkspaceEdit {
	changes: TextDocumentEdit[];
}
```
In the protocol it described as follows:
```ts
interface WorkspaceEdit {
    /**
     * Holds changes to existing resources.
     */
    changes: { [uri: DocumentUri]: TextEdit[]; };
}
```

Also [asWorkspaceEdit](https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/protocolConverter.ts#L418) is wrong.

This is cause of the problem in my extension.

It worked on `"vscode-languageclient": "^2.6.3"`, but it doesn't on `"vscode-languageclient": "3.0.4"`.

gorkem:
@dbaeumer https://github.com/Microsoft/vscode-languageserver-node/commit/75c63564a3a678d8fc39dd63117b8e0be645cbe7 seems to have broken the [asWorkspaceEdit](https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/protocolConverter.ts#L418). This is now a blocker for https://github.com/redhat-developer/vscode-java/issues/99 since we can not move to the 3.x version of the library.

dbaeumer:
@gorkem I will have a look. I know that we discussed this but I need to refresh my memory.

dbaeumer:
Reading through the code I remember why we changed this for the 3.0 version of the protocol. The reason was that the old `WorkspaceEdit` was a map from URI to edit. The problem was that this didn't allow to specify on which version of a document an edit is applicable. This is why we changed this from a map to an array where the `TextDocumentEdit` contains a `textDocument` property pointing to a `VersionedTextDocumentIdentifier`.

I know this is breaking but to support proper version checking when applying text edits this was the best way to go forward. Now users of the protocol have to think about this.

So bottom line is that the spec is incorrect. I will fix this. Would you be able to adopt this?

gorkem:
I am concerned that this will break all the existing implementations, not all clients are able to move as quick.  I think we need a migration path on this. Perhaps we can leave the changes as a deprecated field and use something like `edits` for the `TextDocumentEdit`. 

I think we should consider https://github.com/Microsoft/language-server-protocol/issues/41 before making changes to `WorkspaceEdit`on the protocol.

[...]

PR #180, opened 2017-03-06
WIP: Fix nodePath issue

maxdeviant:
This PR aims to fix the issue described in #179 and [vscode-tslint (#173)](https://github.com/Microsoft/vscode-tslint/issues/173).

### DIFF
--- a/server/src/files.ts
+++ b/server/src/files.ts
@@ -212,9 +212,17 @@ export function resolveModulePath(workspaceRoot: string, moduleName: string, nod
 			if (value.indexOf(path.normalize(nodePath)) === 0) {
 				return value;
 			} else {
-				return Promise.reject<string>(new Error(`Failed to load ${moduleName} from node path location.`));
+				const message = `Failed to load ${moduleName} from node path location.`;
+				if (tracer) {
+					tracer(message);
+					tracer(`Node path location: ${nodePath}`);
+				}
+				return Promise.reject<string>(new Error(message));
 			}
 		}).then(undefined, (_error: any) => {
+			if (tracer) {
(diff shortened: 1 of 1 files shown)


maxdeviant:
I added some trace statements around what I think is the problematic area to see if it can shed some light on the issue.

I do not have an environment set up where I can test all of this locally against the `vscode-tslint` extension, so I'm not really sure how to best proceed.

server/src/files.ts
```diff
@@ -209,12 +209,23 @@ export function resolveModulePath(workspaceRoot: string, moduleName: string, nod
 			nodePath = path.join(workspaceRoot, nodePath);
 		}
 		return resolve(moduleName, nodePath, nodePath, tracer).then((value) => {
+			if (tracer) {
+				tracer(`Resolved module path is ${value}`);
+			}
 			if (value.indexOf(path.normalize(nodePath)) === 0) {
```
maxdeviant: I think the issue lies with this `indexOf` check. It could be something as simple as `c:/` vs `C:/` in the path name.

Between this trace statement and the one in the `else` case, it should be hopefully be pretty clear what is going wrong.

dbaeumer:
I will close the PR since it only added logging. I implemented a correct path comparison depending on the OS the server is running.

maxdeviant:
@dbaeumer Thank you so much! Will the fix be in the next version of VS Code?

dbaeumer:
No, since VS Code doesn't ship the vscode-tslint extension. We first need to publish an official version of the libraries (version 3.2.0) and then vscode-tslint need to pull the new version in and deliver a new version.

PR #189, opened 2017-04-04
Add return value to getEnvironment().

trixnz:
This fixes a bug where `env` was not being propagated to spawned server instances.

I was debugging why my user-defined `env` object was not being passed down to my spawned language server and noticed the missing return value, resulting in `undefined` being passed down the chain.


dbaeumer:
@trixnz great catch. Thanks!

PR #193, opened 2017-04-10
Detect and copy npm-shirkwrap file if present

XVincentX:
This PR, when a shrinkwrap file is provided as `argv[5]` and this exist, will make sure that file gets copied so I can lock down my dependencies in both client and server installations and have a 100% reproducible environment in my local development experience.

This should become a best practise, worth mentioning in the documentation.

### DIFF
--- a/server/bin/installServerIntoExtension
+++ b/server/bin/installServerIntoExtension
@@ -57,5 +57,12 @@ var dest = path.join(extensionServerDirectory, 'package.json');
 console.log('Copying package.json to extension\'s server location...');
 fs.writeFileSync(dest, fs.readFileSync(packageFile));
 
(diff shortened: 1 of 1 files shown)


dbaeumer:
Great idea.

dbaeumer:
I reviewed the code and a couple of comments:

- to keep things consisted I think the shrink wrap file should be resolved as well (as is the package.json on line 23)
- the dest var must be adjusted. Right now the shrink wrap file is copied over the package.json file. Is this on purpose.

XVincentX:
@dbaeumer You were right on both comments. I should have fixed these now.

dbaeumer:
Thanks. Will merge.

XVincentX:
Thanks!

PR #195, opened 2017-04-16
client: Fix unintended argument passing

ishitatsuyuki:
This is probably working for most time since it's referencing the same object, but not for the undefined case.

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -1557,7 +1557,7 @@ export class LanguageClient {
 			let command: Executable = <Executable>json;
 			let options = command.options || {};
 			options.cwd = options.cwd || Workspace.rootPath;
-			let process = cp.spawn(command.command, command.args, command.options);
+			let process = cp.spawn(command.command, command.args, options);
 			if (!process || !process.pid) {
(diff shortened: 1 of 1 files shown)


dbaeumer:
@ishitatsuyuki can you elaborate where this fails in your opinion. The idea is that if no options are provided the cwd will be the workspace rootPath which will be undefined if opened on a single file.

ishitatsuyuki:
I have no idea what you're talking about. Whether `options` is defined has no relation to opened workspace (it's up to the extension's code).

This is something that I came across when I'm checking the code for debugging another problem. Thus, I have no clear failure case for this, just that the code doesn't look like what it should be.

dbaeumer:
Can you then elaborate what you think the code should do and why it doesn't behave that way. IMO the code does the right thing.

ishitatsuyuki:
If options is undefined, the local variable `options` and `command.options` points to different things. The former has more complete data, which is probably intended to be passed.

dbaeumer:
The idea behind the current code is to make sure that cwd points to the workspace root if not provided. I am closing the PR since the PR changes this behavior.

PR #216, opened 2017-06-10
Only convert `undefined` argument to null (fixes #215)

### LINKED ISSUE #215: Falsy values are translated to `null`
Message values like `0` or `''` get translated to `null`.

svenefftinge:
Falsy parameter values were translated to `null`. This change fixes that.


dbaeumer:
@svenefftinge very good catch.

dbaeumer:
One little thing I changed. We usually use void 0 for undefined.

PR #220, opened 2017-06-19
Fix rename ResponseError not being passed onto VSCode

Gama11:
I was trying to return a `ResponseError` from the rename method in the Haxe language server to display an error in the UI (like the TypeScript language server does in some siutations, [example](http://i.imgur.com/M5ioIWz.png)).

However, no matter what error code I used, the end result was always the same, VSCode displays `"No result."` and the language client just logs it to the output channel:

This is probably not a proper solution, and there might be other methods that are affected similarly, but it _does_ fix my issue:

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -2317,7 +2317,7 @@ export abstract class BaseLanguageClient {
 				this._p2c.asWorkspaceEdit,
 				(error: ResponseError<void>) => {
 					this.logFailedRequest(RenameRequest.type, error);
-					Promise.resolve(new Error(error.message));
+					throw error.message;
 				}
 			);
 		};


dbaeumer:
@Gam11: thanks for the PR. However I fixed this differently since we usually avoid to throw with promises. The problem was that the code used Promise#resolve instead of Promise#reject.

nadako:
That fix doesn't really help with the original issue - VS Code still shows "No result" instead of the error message from the server.

dbaeumer:
Seems to be a VS Code issue. Opened https://github.com/Microsoft/vscode/issues/37808

Gama11:
This is actually still an issue with VSCode 1.21.0 and vscode-languageclient 4.0.0:



I'm not sure which of the two is at fault, but this error shows up in the dev console after a failed rename:



I gave your test case from https://github.com/Microsoft/vscode/issues/37808 a try, and that works as expected.

PR #222, opened 2017-06-24
Add DocumentSelectorFactory

bmewburn:
Hello,
This pull request adds a factory method to `LanguageClientOptions` to allow different `DocumentSelector` values depending on the method that is being registered. This is useful for language servers that may provide services for multiple languages but may not want every capability registered for every language.

An example is providing PHP completions when in HTML language mode but avoiding registering a PHP specific formatter. VSCode only attempts to use a single format provider leading to issues over which registered formatter to choose. Avoiding registering an inferior PHP formatter for HTML solves this. See https://github.com/Microsoft/vscode/issues/670#issuecomment-310822851.

```Typescript
export interface DocumentSelectorFactory {
	(method:string): DocumentSelector | string[];
}

export interface LanguageClientOptions {
	documentSelector?: DocumentSelector | string[];
	documentSelectorFactory?:DocumentSelectorFactory;
	synchronize?: SynchronizeOptions;
	diagnosticCollectionName?: string;
	outputChannelName?: string;
	revealOutputChannelOn?: RevealOutputChannelOn;
	/**
	 * The encoding use to read stdout and stderr. Defaults
	 * to 'utf8' if ommitted.
	 */
	stdioEncoding?: string;
	initializationOptions?: any | (() => any);
	initializationFailedHandler?: InitializationFailedHandler;
	errorHandler?: ErrorHandler;
	middleware?: Middleware;
	uriConverters?: {
		code2Protocol: c2p.URIConverter,
		protocol2Code: p2c.URIConverter
	};
}
```

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -437,8 +437,13 @@ export interface Middleware {
 	resolveDocumentLink?: (link: VDocumentLink, token: CancellationToken, next: ResolveDocumentLinkSignature) => ProviderResult<VDocumentLink>;
 }
 
+export interface DocumentSelectorFactory {
+	(method:string): DocumentSelector | string[];
+}
+
 export interface LanguageClientOptions {
 	documentSelector?: DocumentSelector | string[];
+	documentSelectorFactory?:DocumentSelectorFactory;
 	synchronize?: SynchronizeOptions;
 	diagnosticCollectionName?: string;
 	outputChannelName?: string;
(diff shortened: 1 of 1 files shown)


dbaeumer:
@bmewburn that makes sense but with the work I did the last weeks this should be implemented differently. All code to hook the protocol to the VS Code API is now in so called features. Each feature has a register method which takes a document selector. Currently the client still creates is default features But I would suggest that we add an option to create a client blank and let the consumer create the features. The consumer can then register them for different document selectors.

The code changes I did around this are in master. Let me know if you want to look into this.

bmewburn:
Thanks for reviewing, time permitting I'll take a look. Though, I'm thinking that perhaps my PR here was more of a stop gap measure to get the html language server and my php server to play nicely together. I'm still going to run into problems in vscode where only a single formatter is chosen. I might be better spending time modifying the client to work with multiple language servers to get this working nicely.

PR #231, opened 2017-08-06
Remove reference to a non-existent range property

rcjsuen:
The documentation for `textEdit` in `CompletionItem` refers to a `range` property but this property doesn't actually exist on `CompletionItem`.


dbaeumer:
@rcjsuen Thanks. Unfortunately the PR removes to much. I fixed it by hand and pushed.

PR #233, opened 2017-08-08
Use the client's code2ProtocolConverter to convert fileSystemWatcher events

### LINKED ISSUE #228: LanguageClient doesn't respect uriConverter.code2Protocol option when sending DidChangeWatchedFilesParams
This is related to https://github.com/Microsoft/vscode-languageserver-node/issues/105.

VS Code sends a non-standard URI on Windows. Instead of sending file:///c:/Us... it sends file:///c%3A/Us... (notice the %3A). We can modify this behavior by providing a [`uriConverter`](https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/client.ts#L455) to `LanguageClientOptions`.  And this works for _almost_ all the different requests that include a URI.

**Where it fails is the `DidChangeWatchedFilesParams`.**

When I receive a `DidChangeWatchedFilesParams` it doesn't go through my custom uriConverter and still has the %3A, necessitating my server to do the conversion for this case.

Is this working-as-design? Or is this an oversight? If you look at the [`Converter` interface](https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/codeConverter.ts#L14) there is no accommodation for `DidChangeWatchedFilesParams`.

Based on the [protocol](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#workspace_didChangeWatchedFiles), I would expect that a client consistently sends the URI in a particular format for all requests that include a URI.

vazexqi:
If the client has a code2ProtocolConverter, we should use it to convert the uri from fileSystemWatcher events.

Fixes #228.

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -579,7 +579,7 @@ namespace DynamicFeature {
 }
 
 function ensure<T, K extends keyof T>(target: T, key: K): T[K] {
-	if (target[key] === void 0) {
+if (target[key] === void 0) {
 		target[key] = {} as any;
 	}
 	return target[key];
@@ -1036,7 +1036,7 @@ class FileSystemWatcherFeature implements DynamicFeature<DidChangeWatchedFilesRe
 
 	private _watchers: Map<string, Disposable[]> = new Map<string, Disposable[]>();
 
-	constructor(private _notifyFileEvent: (event: FileEvent) => void) {
+	constructor(private _client: BaseLanguageClient, private _notifyFileEvent: (event: FileEvent) => void) {
 	}
 
 	public get messages(): RPCMessageType {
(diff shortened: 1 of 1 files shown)


client/src/client.ts
```diff
@@ -808,7 +808,7 @@ class FileSystemWatcherFeature implements FeatureHandler<DidChangeWatchedFilesRe
 
 	private _watchers: Map<string, Disposable[]> = new Map<string, Disposable[]>();
 
-	constructor(private _notifyFileEvent: (event: FileEvent) => void) {
+	constructor(protected _client: BaseLanguageClient, private _notifyFileEvent: (event: FileEvent) => void) {
```
vazexqi: Do you prefer this to be protected or private?

dbaeumer:
@vazexqi thanks for the PR. I replayed the changes by hand since the code changed due to the multi workspace folder work in the LSP.

PR #259, opened 2017-10-06
Do not assume that a client has workspace capabilities

### LINKED ISSUE #257: LSP client that doesn't support any workspace capabilities will cause an error if the server is using workspace/workspaceFolders
If you create your server with `let connection = createConnection(ProposedFeatures.all);` and use a client that connects to it with only very rudimentary capabilities, an error will be thrown.
```TypeScript
function initialize(id: number) {
  send(id, "initialize", {
    rootPath: process.cwd(),
    processId: process.pid,
    capabilities: {
      textDocument: {
        completion: {
          completionItem: {
            snippetSupport: true
          }
        }
      }
    }
  });
}
```

```JavaScript
{
  jsonrpc: '2.0',
  id: 1,
  error: {
    code: -32603,
    message: 'Request initialize failed with message: Cannot read property \'workspaceFolders\' of undefined'
  }
}
```
This is because workspaceFolders.proposed.ts incorrectly assumes that `clientCapabilities.workspace` is set.
https://github.com/Microsoft/vscode-languageserver-node/blob/9059f83cd8fb25b4a0fe5429540c562aa899dc6e/server/src/workspaceFolders.proposed.ts#L23-L24

However, according to the [protocol](https://github.com/Microsoft/language-server-protocol/blob/b04dc06202bd0cbca06cd4b0b4a640aa46e8a081/protocol.md), that capability is optional as it is flagged with a `?`.
```TypeScript
interface ClientCapabilities {
	/**
	 * Workspace specific client capabilities.
	 */
	workspace?: WorkspaceClientCapabilities;

	/**
	 * Text document specific client capabilities.
	 */
	textDocument?: TextDocumentClientCapabilities;

	/**
	 * Experimental client capabilities.
	 */
	experimental?: any;
}
```

rcjsuen:
Fix #257 by checking that the client actually has workspace capabilities defined before trying to read it.

https://github.com/Microsoft/vscode-languageserver-node/blob/9059f83cd8fb25b4a0fe5429540c562aa899dc6e/server/src/workspaceFolders.proposed.ts#L23-L24

### DIFF
--- a/server/src/workspaceFolders.proposed.ts
+++ b/server/src/workspaceFolders.proposed.ts
@@ -21,7 +21,7 @@ export const WorkspaceFoldersFeature: WorkspaceFeature<WorkspaceFolders> = (Base
 		private _unregistration: Thenable<Disposable>;
 		public initialize(capabilities: ClientCapabilities): void {
 			let workspaceCapabilities = (capabilities as Proposed.WorkspaceFoldersClientCapabilities).workspace;
-			if (workspaceCapabilities.workspaceFolders) {
+			if (workspaceCapabilities && workspaceCapabilities.workspaceFolders) {
 				this._onDidChangeWorkspaceFolders = new Emitter<Proposed.WorkspaceFoldersChangeEvent>();
 				this.connection.onNotification(Proposed.DidChangeWorkspaceFoldersNotification.type, (params) => {
 					this._onDidChangeWorkspaceFolders.fire(params.event);


rcjsuen:
Looks like @dbaeumer fixed this with cbf3d48c89bd5909fd88447d3a1c7a82c91629d7 so this pull request isn't needed anymore.

dbaeumer:
@rcjsuen I apologize. Overlooked the PR.

Issue #245, opened 2017-08-29
Is socket support works correct?


destitutus:
Language server answers nothing if `--socket` option used. Is implementation broken?

```bash
node -v
v6.11.2
```

Ways to repeat. Install Language server, for example `npm install -g ocaml-language-server`
Run it via command line `ocaml-language-server --socket=3000`
Connect via telnet to server
```bash
$ telnet
telnet> toggle crlf
Will send carriage returns as telnet <CR><LF>.
telnet> open localhost 3000
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
Content-Length: 58

{"jsonrpc":"2.0","id":2,"method":"initialize","params":{}}
```

It replays nothing on correct request.
If I change line https://github.com/Microsoft/vscode-languageserver-node/blob/master/server/src/main.ts#L1460-L1464

to
```js
let server = net.createServer(socket => {
                server.close();
                output.on('data', function (d) {
                        socket.write(d)
                })
                socket.on('data', function (d) {
                        input.write(d)
                })
            }).listen(port);
```

it replay with correct response

```bash
$ telnet
telnet> toggle crlf
Will send carriage returns as telnet <CR><LF>.
telnet> open localhost 3000
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
Content-Length: 58

{"jsonrpc":"2.0","id":2,"method":"initialize","params":{}}
Content-Length: 152

{"jsonrpc":"2.0","id":0,"method":"window/showMessageRequest","params":{"type":2,"message":"Cannot find merlin binary at \"ocamlmerlin\".","actions":[]}}
```

Is it known issue?

dbaeumer:
@aeschli can you please answer. You originally put that in.

aeschli:
Sorry, I'm not really the node streams expert. So maybe this never worked?

PavelSosin:
It looks like ALL LSP implementations based on this package have the same issue!

rcjsuen:
I've tried connecting to the example server in the past and also got nowhere. Please see Microsoft/vscode-languageserver-node-example#37..

AHaberl:
So will there be a fix for this?

dbaeumer:
Fixed. Published as 3.5.0-next.1

Please note that socket connections are only supported when both the client and server run on the same machine.

dbaeumer:
@AHaberl usually a PR helps as well :-)

[...]

PR #260, opened 2017-10-07
Use PassThrough to pipe data to and from sockets

rcjsuen:
The original code seems to have been piping the content in the wrong direction. The data that we read from the socket should be the input and the output's data should be written to the socket.
https://github.com/Microsoft/vscode-languageserver-node/blob/a8584e62075f327b240e10a444c5e723718d8c4c/server/src/main.ts#L1483-L1487


dbaeumer:
Closing. See comments in https://github.com/Microsoft/vscode-languageserver-node/issues/245#issuecomment-336054699

PR #261, opened 2017-10-08
Add proposed window/progress extension

Xanewok:
This proposes a simple `window/progress` server -> client notification that's been described in https://github.com/Microsoft/language-server-protocol/pull/245. I tried to follow [contributing.md](https://github.com/Microsoft/language-server-protocol/blob/master/contributing.md) as closely as possible, however it's possible I missed something. Additionally I'm not a TS/JS dev, so if there are any concerns for the client implementation, I'm more than open to changing it accordingly.

I think it's a simple but often requested feature and it'd be good to land it, at least with a simple initial version to gather proposal and improve upon it.

cc @felixfbecker this proposes the https://github.com/sourcegraph/language-server-protocol/commit/8986486a61f50bfe8ed449d1fd0b8e2f42656a5a.


ljw1004:
I know you've done your work solely for VSCode. I'm thinking about how it would be incorporated into LSP for use by other editors as well. In editors where "progress" is indicated by a spinner on the status bar with text-only tooltips, how do you think the "title+message+percent" should be surfaced?

progress

This is a screenshot where two different LSP-providers are each providing their own progress report.

The reason I ask is because I'm inclined towards an API where progress is solely reported by a single plain-text string, without breaking it up into title/message/percent.

Some providers will find that a percentage number is a poor way of presenting the information. For instance, an C++ indexing provider might report `indexed 5/20 files`, but then change it to `6/35 files` as it discovers more information. In this case a percentage number wouldn't be useful (it would fluctuate up and down and would more reflect how much of the scope had yet been discovered rather than how much of the work had been done). But the pair of numbers is very informative.

I think that with your proposed split into message/percent, we'll end up in a world where 50% of providers report their progress with a custom piece of text inside the message like `5/20 files`, and others report it with a percentage number following your scheme, and this lack of standardization will mean they're not actually useful anymore.

felixfbecker:
It should definitely work for all kinds of editors and it is up to the editor to display a spinner, a progress bar or show only the text. If the language server wants to update the message with `indexing file x/y` that is totally fine and it can do that in addition to the percentage. The percentage is only to power a determined progress indicator, if the editor has a UI element for that, it can use it, if it uses an indeterministic progress indicator, it doesn't have to use it. An earlier version of this proposal had the percentage split into `current` and `total`, but I don't think there is value in having these as separate programmatic-inspectable fields.

ljw1004:
@felixfbecker Imagine if this PR goes ahead in its current form. Then I think few LSP servers will emit both `8/16 files` in their message and also a `50%` progress indicator because (in the current PR) it will render as `indexed 8/16 files (50%)` which looks kind of irritating.

I think the idea "it's up to the client to decide" only goes so far. Because LSP servers will alter themselves to look good on the dominant client.

Xanewok:
@ljw1004 while I recognize how you may feel about it, I think this minimal solution finds a good balance between giving freedom for the LSP clients and providing more ability to programatically inspect the given value. I think it's worth noting that separating current/total and the percentage is sometimes desired:
* sometimes the units of work are not equal, e.g. 14/30 might mean 70% of total work is done
* `processing headers: 16781/321681 (5.2%)` allows to quickly spot order of magnitude difference at a glance
* from a UX point of view it might be feasible to indicate progress being done even when processed/total ratio doesn't change
* you don't always want to expose the exact current/total ratio to the user - by providing only a single progress measure this forces the client to always display the exact amount in the message or always use it for a progress bar (specifying if the client can omit the percentage/value and in what context can it use it would warrant additional flags in the params, which I don't think is ideal).

That's why I think providing both optional message and optional percentage allows for more and is concise and simple enough for most cases.

Maybe we could emphasize that `percentage` can be consumed by the LSP client to display the progress using UI?

ljw1004:
@Xanewok Another possibility is to indicate that the percentage should be rolled into the message (1) always, (2) if no other progress indication is available, (3) never. But that's over-engineering.

Actually I'm more concerned about `title` at the moment. I read through the source code but couldn't figure out how the title is rendered. Maybe a screenshot?

felixfbecker:
I agree with @ljw1004 that it is not good to put the percentage into the title automatically in this "reference implementation". The percentage should only be used to update a deterministic progress indicator which VS Code has in other places:

https://github.com/Microsoft/vscode/blob/24d7998be78fcd340ac5c4f6e5a7a660a26b24b9/src/vs/workbench/services/progress/browser/progressService2.ts#L70

Xanewok:
@ljw1004 yeah, I was just editing my response about the title.

I agree with you that it may in fact be too specific here. I left it here, since it was in the original proposal and VSCode is capable of using it. Here's a screen:
window-progress

@felixfbecker agreed, I saw only `withProgress<R>` API with `Progress<{ message?: string; }>` that's why I wrapped it into the message. I'll try using the variant with `number`, thanks for pointing that out.

Xanewok:
@felixfbecker actually I'm not sure about the exposed API, not much of a TS/VSCode dev myself, mind pointing out how can I also include the percentage there?

[vscode.d.ts API](https://github.com/Microsoft/vscode/blob/6e7a575339fcb38503d1a9d2778d7d69475dd8e2/src/vs/vscode.d.ts#L4680) doesn't seem to expose the percentage, while the [implementation](https://github.com/Microsoft/vscode/blob/35b91522e736711680a558a7d801f30b5b69b039/src/vs/workbench/api/node/extHost.api.impl.ts#L386) has it?
I'm confused :cry:

keegancsmith:
It seems the number progress was initially meant for use in the now deprecated `withScmProgress`. I assume when that is removed they will remove support in the implementation.

Given the number is contentious maybe we should exclude it for now? I only included it in the linked PR because at the time I proposed it, I was matching the functionality of vscode. Allowing developers to report back numbers for progress gives me unpleasant dreams back to file copy dialogs in Windows XP.

felixfbecker:
Would be fine for me to implement it with only supporting indeterministic progress indicators for now, we can add it later as a backwards-compatible change

ljw1004:
The next question is: once we remove `percent` from the API, how about we do something with `title` and `message`?

In the screenshot you appeared to show two different UXs:
* In the status bar it prints `message`
* In the datatip it prints `${title} - ${message}`

First, the ability to display the text in the status-bar seems *extraordinary*. In our environment we commonly have several progress messages all at once, e.g. `lint path_to_file` and `hh_server is initializing` and `indexing files`. To fit into the status bar, (1) it will look pretty goofy if there are three things adjacent to each other, (2) they basically have to condense themselves down into just one single word.

I propose removing the current `title` and `message` and replacing them with two other fields:

```
// message is a mandatory string for this progress operation. Editors may display it in
// datatips. Example: 'Indexing file 5/25 (5%)'
message: string;

// shortMessage is mandatory too. Some (not all) editors will use this when displaying
// progress in space-limited situations such as the status bar. Example: 'Indexing (5%)'
shortMessage: string;
```

My thinking is that to fit into the status-bar we LSP-server-authors will have to craft our messages with extraordinary care and compactness. And even to fit into a datatip we still need to craft our messages carefully to be brief enough and actionable enough. Crafting those two things effectively is difficult and we shouldn't be hamstrung.

Xanewok:
Should the percentage in the displayed message in the reference
implementation not be included there, then?

Also about providing a title in the params, what are your opinions? Do you
think it can be actually useful or is it too specific?

In the specific case of Rust Language Server we have 2 operations, one
executed after another: Compiling a diagnostic build and processing
analysis data, so I assume we could stick a "Compiling" and "Processing"
titles for those appropriately, but since the messages will probably also
include that information, I think it seems redundant, and in general case I
can't still think of a good use case for that.

felixfbecker:
@ljw1004 I don't see a difference in content between `shortMessage` and `title`, but the implementation in this PR should not do any string concatenation of any properties. And I don't think messages should be required. If an LS wants to just indicate that it is doing _some_ work it should be able to omit the messages, it would be silly to require them to set an empty string or whatever workaround would be needed.

ljw1004:
@felixfbecker - you anticipated my next suggestion, which is that the LS itself should be responsible for putting the name of the LS into the message. In the PR, the tooltip is a concatenation of (1) LS id, (2) title, (3) message, (4) percent.

I think the LS should be able to decide how to display the ID itself.

Given this, if message is completely empty, the user won't even know which LS or other service it was that produced the spinner.

Xanewok:
@ljw1004 thanks for the condensed status bar suggestion!
In this case I think the `title`/`message`/`percentage` is a good combo: the clients can decide in which format they should present the message.

* clients with small amount of space and no spinner/UI, such as Vim, can display it as "Title" or "Title (Percentage%)"
* editors with some progress bar or other notification UI can also use the passed percentage number value to drive the UI and display only "Title" or "Message" (or however)
* clients, not the servers, should care whether they are capable of displaying the longer messages or the shorter ones, hence why if, e.g., there's only a single progress reported it can use the longer version, but if the server sends 2 additional progresses it can condense each and display only a title (with a text percentage or UI progress or without it completely).

We can change the names of the passed values and/or update documentation to better reflect its intended usage, but with these values any given LSP client has all the appropriate information to present it in a convenient and descriptive manner.

Regarding the LS ID, I think it's entirely up to the client how or if at all, does it show the origin of the progress indication and the indication itself.

In this reference client implementation, the VSCode `windows.withProgress` API automatically prepends the language ID and I think:
* that' outside the scope of the LSP progress reporting
* this would require more tinkering and not using the `window.withProgress` API (this would probably require more engineering, creating a custom IStatusBarItem, gluing it to the LS client, changing the default tooltip shown etc.).

I have no idea how to do that, so if we were to get rid of the default `<languageId>: ` prefix, any implementation tips are welcome!

ljw1004:
@Xanewok I think it will be hard to pull off. You try it... write down five example progress reports with title+message+percentage divided up in such a way that (1) title+percent looks good, (2) title+message+percent looks good, (3) Title alone looks good, (4) Message alone looks good.

I don't think it can be done!

I myself tinkered with my progress reports off and on over a month to make them as compact and actionable as possible. Here are the ones I came up with:

* hh_server is initializing [naming] [53 seconds]
* [Hack] preparing to check edits
* [Hack] checking edits
* [Hack] checking entire project
* Starting Hack language service
* Stopping Hack language service
* Flow server is busy (/users/ljw/code/proj1)

I don't know how I'd divide those into languageId/title/message/percent effectively.

Xanewok:
@ljw1004 I hope I didn't come off as rude, I certainly didn't intend to and I hope I wasn't received as such! 😞 

I'm a bit lost at this point - not sure if we're discussing the protocol itself or the reference VSCode implementation.
From the protocol and server's responsibilities point of view, I think sending title/message/percent (each optional!) is simple and sufficient, while the presentation side, the client, should decide how it uses the information.

Now onto what a client might do...
I think messages like 'Starting/Stopping XYZ language Service' should be outside of the protocol and presented by the client on its own, not expecting the server to do so, since it's very general and tied to what the client does, not the server.
Without factoring in the languageId, I think messages like: (each part is optional)
* Initializing/hh_server is initializing [naming]/X%
* Indexing/indexing headers 312/6582/X%
* Checking/checking {edits, project}/X%
* Waiting (for server)/Flow server is busy (/users/ljw/code/proj1)

might be okay? In these cases you might present it as Initializing, Initializing (hh_server is initializing [naming]), hh_server is initializing each optionally with (X%) stuck at the end, or optionally using UI, and I think it looks good enough.

As I said, clients such as Vim may only use the title itself, since their space is limited. GUI editors could detect if they need to condense the message and only display a title (optionally with a percentage/spinner?) and possibly provide a full name on hover (e.g. somewhat what VSCode also does).

I would leave languageId out of it, since sometimes these progress reports can be presented in a separate window per each LS, or even have their own space at the status bar (e.g. VSCode) and so I think it's only fair for an extension to group related progresses and use only a single language id in a text format, if it needed - for example a Vim frontend might display:
```
Rust: Compiling/Processing       Hack: Checking         Flow: Waiting for server
```
if it had 4 tasks running for 3 different LSs, in this example.

ljw1004:
@Xanewok you've been the model of politeness and a very pleasant person to discuss with!

Following your reference implementation, this is what people will see when they hover over the progress indicator:

```
[hack] Initializing - hh_server is initializing [naming] (X%)
[hack] Indexing - indexing headers 312/6582 (X%)
[hack] Checking - checking edits (X%)
[flow] Flow server us busy (/usrs/ljw/code/proj1)
```

The middle two look wrong. Users will immediately file bugs that say "Why is the word 'indexing duplicated? Why is the word 'checking' duplicated?" And so the language server developer will immediately be forced to omit the word from the title so as to look good in your reference VSCode implementation -- with the knock-on effect that they now look bad on clients which chose to display only the title. Or vice versa, they'll be forced to omit the word from the message -- and so will now look bad on clients which chose to display only the message.

Xanewok:
I'm very much open to changing the reference implementation! Frankly I wanted to have something simple working and reiterate from there. I agree that currently it does not look really good and needs to change.

Xanewok:
Now percentage is not included as text in any messages, also not using title feature in the API to prevent weird/duplicated `title - message` tooltip.

For context: (incl. couple of issues, along with reasoning on the current `withProgress` API design)
* [progress.ts (what indicator internally accepts)](https://github.com/Microsoft/vscode/blob/master/src/vs/platform/progress/common/progress.ts)
* [Tooltip being presented as `title - message`](https://github.com/Microsoft/vscode/issues/30672)
* [percentage field being removed from API](https://github.com/Microsoft/vscode/issues/25534)

dbaeumer:
@Xanewok @ljw1004 @felixfbecker thanks a lot for the very lively discussion. Highly appreciated! I read through all comments. May be it would make sense to look at the problem from a user perspective as well. From my experience so far with building progress APIs users are usually interested in the following things:

**Progress to know when things are done:**
- a progress indication (either percentage or 5/35 files indexed, were the total number of files might change)
- if no progress is indicated or an infinite progress is used (an endless spinner) then users want to see a changing message indicating what the system is doing.

**Progress Details**:
- detailed progress is only necessary for things the user is waiting for. If for example the users triggers a build and the users needs to wait for the build to finish he wants to see progress. 
- background tasks like indexing don't need detailed progress. It would be enough to simply indicate that something (or many things) are active in the background. Details might only be necessary on user request (clicking on something). 
- However the indexing progress (from pure background) might change if the user starts a search which needs the indexing to be finished first. Then a detailed progress about indexing is necessary. 

I haven't thought about the necessary protocol we need and the UI implementation to best support this.

Xanewok:
I guess one idea would be to introduce a notion of tasks and tie progress reporting to that?
It's hard to do anything more complex, e.g. encode in the protocol that some request depends on some other task finishing, without pulling in streaming support and similar or making deep changes in the protocol, I believe.

dbaeumer:
I like the idea of introducing a task/operation for which progress can be reported. I would however not make them dependent since this would need some upfront knowledge which might be hard to compute. In the index example the search operation could simple show a message `Waiting for indexing to be finished' to express the dependency.

Protocol wise I think we don't need any changes. I could imaging something like this:

- request from server to client to start a progress / operation. Response would be a progress Id.
- server reports progress against that Id.
- server can control some properties via notification on the progress Id. (switch from background to foreground, ...)

Thoughts ?

felixfbecker:
I would like to not overcomplicate this (e.g. by requiring the server to use different notifications to create progresses). The proposal in its current form has an `id` and `done` property, that is all that is needed to create a progress, update it and eventually finish it.

Xanewok:
Yes, I do believe that server->client (↪️) request to start a progress, only so the server manages it completely feels somewhat clunky.

I think using only notifications here seems more straightforward and clean in this case.

However, despite not wanting server to handle more UI logic, I do like the idea of background/foreground task. For example indexing could start off as a background task and if a client requests action that effectively requires the task to be finished (let’s say listing all symbols in the workspace), the task could be promoted to foreground to inform user that certain requests may return partial/empty results until it’s finished.
This seems fairly simple and doesn’t complicate the protocol further, while seems to be the best od two worlds.

Or the background/foreground and more complex interactions could be provided extension to the bare bones progress notification that should be in the base protocol?

dbaeumer:
Letting the server decide the id and then use it to refer to a specific operation makes things for sure easier.

Xanewok:
@dbaeumer does this need more design changes to be reviewed or does it only need merge conflict resolution?

ljw1004:
I'm still puzzled by this bit:
```
+        this.title = params.title || this.message;
 +        this.message = params.message || this.message;
 +
 +        const displayedMessage = this.message || this.title;
 +        this.progress.report({message: displayedMessage});
```

If I understand right, (1) the `percentage` field send from the LSP server is completely ignored, (2) if the progress has used the `message` field then `title` will be ignored. My understanding from the thread is this is motivated by a wish to allow different clients to render these components in different ways.

It's dangerous to design for clients that don't exist! I think we should either flesh out those putative clients to validate that this design makes sense, or abandon the extra fields and aim just for a simple small proposal. Here's my attempt at fleshing out the possible clients:


## Prototype Client 1

Displays a spinner in the status bar if there are 1 or more outstanding progresses. If you hover, then it displays a concatenation of title and message, with a hyphen in between. Percentage is ignored.

image

## Prototype Client 2

As above, but percentage is displayed inside square brackets

image

## Prototype Client 3

The titles of all ongoing progress reports are shown in the statusbar. If you hover over any individual one, then its "Title - Message [Percentage]" is displayed.

image

## Prototype Client 4

As above, but the tooltip only displays the message

image

## Prototype Client 5

Another version where the tooltip only displays `message`, but the percentage is indicated with a progress bar.

------

As before, I still think the API isn't enough to support even these different clients: LSP-server-authors will end up splitting their information into Title/Message/Percentage in a way that looks good on VSCode, and it will

Xanewok:
To be completely blunt - it's a very simple API and mechanism, which aims to solve a common and practical problem that many lsp servers have, that is more important for the end-users rather than lang-server devs.
That is why I believe we need to stabilise a common solution for this problem as quickly as possible, since the quicker we do it, the easier the transition will be and also the more happy LSP-using devs we'll have :smiley: 

I thought about this and believe that the main problem stems from interpretation of the data included, rather than the actual data and its format itself.

Firstly, like @felixfbecker said earlier, the percentage sent as a number can be mostly thought as an enhancement, so the appropriate capable UI can present deterministic progress accordingly, but I'm also fine without it now. We can always make backwards-compatible change and it's convenient, as VSCode doesn't allow us to easily display progress bars, so the reference implementation now can be even easier.
So let's ditch percentage field for now.

Secondly, there's some disagreement regarding the `title`/`message` split, how it should be present together (if at all), what does it mean etc. Taking into account more restrained UIs and for clarity/UX, I now lean more towards @ljw1004's solution from https://github.com/Microsoft/vscode-languageserver-node/pull/261#issuecomment-335295897.

I propose that params should look more like this:
```typescript
export interface ProgressParams {
    /**
     * A unique identifier to associate multiple progress notifications with the same progress.
     */
    id: string;

    /**
     * Operation being done as a part of reported progress.
     * A good example might be 'Indexing' or 'Fetching assets'.
     */
    operation: string;

    /**
     * Optional message that contains details of performed work, in association with
     * currently executed operation. May be presented along with current operation.
     * For example, this might be 'http://example.com/resource.foo', to be used with
     * 'Fetching assets' or '7/35 files', to be used with 'Indexing'.
     */
    details?: string;

    /**
     * Set to true on the final progress update.
     * No more progress notifications with the same ID should be sent.
     */
    done?: boolean;
}
```
This would mean that the lang-servers always need to specify work done in a concise manner with optional details. I believe this is simply enough, covers almost all use cases and is hard to misuse.
What are your thoughts?

tsmaeder:
@dbaeumer 
>     * request from server to client to start a progress / operation. Response would be a progress Id.
> 
>     * server reports progress against that Id.

+1 on this: however, it's essential that we can still have progress reports without a request first.

tsmaeder:
I'd like to chime in on the side of those would would like a workDone/totalWork done as it's simple to compute a percentage from this pair, but not the other way around.

felixfbecker:
This proposal has been sitting for a while, seems functional and the concerns discussed here feel like bikeshedding. Given that the [official Rust language server already implements this proposal on the server side](https://github.com/rust-lang-nursery/rls/blob/0e7432a4ff58bf3584c44bdb390d557cee846968/src/lsp_data.rs#L366-L393), we intend to ship a client implementation that [conforms to the interfaces in this proposal in its current state](https://github.com/Microsoft/vscode-languageserver-node/blob/38429afa03bdf5e878ec988b4ddd6ee483953a4d/protocol/src/protocol.progress.proposed.ts) (which is also what the Rust LS sends) in the Sourcegraph web client and some language servers.

https://github.com/sourcegraph/sourcegraph/issues/947

felixfbecker:
Sourcegraph now handles `window/progress` notifications as described in this proposal. Demo:



It would help if this PR got merged so we didn't have to copy the file and could track changes made to it easier (through automated dependency updates). It's still suffixed with `.proposed` so imo it would be fine to iterate on it later if needed. But so far it has been working well

Xanewok:
Rebased (hopefully correctly, it's been a while). I agree with @felixfbecker, I think we should land it and possibly iterate from there. @dbaeumer do you think we could merge this?

I'd like to add that while RLS does implement `window/progress` like that, we do know it's only a custom extension at this point and would be willing to change the params/type layout.

What do you think about revised layout from https://github.com/Microsoft/vscode-languageserver-node/pull/261#issuecomment-357446446? One thing that came to mind now is that we could support an optional parameter of type:
```ts
// ProgressParams.percentage?: PercentageType;
export type PercentageType = Ratio | Number | Indefinite;

export type Ratio = {
    done: Number; // non-negative
    total: Number; // positive
};
```
This way we could give freedom for the clients to render it as they wish. They could even not render the info at all, it's just that this information may be useful (and we don't lose the ratio per @tsmaeder's [comment](https://github.com/Microsoft/vscode-languageserver-node/pull/261#issuecomment-435849998)).

felixfbecker:
I don't really see a use case for it. Just `percentage` has worked fine so far, and it makes it easy to compose progress (i.e. total progress of multiple progresses). Just summing up `total` values and `done` values would be a bit weird because it's not correct to sum up multiple `done` values because the units might be different (one could be bytes downloaded while the other could be number of files processed), so `done` would lose any semantic meaning. Aggregating `percentage` though is perfectly valid because it is always the same unit (a percentage). `done` would lose its meaning to the client, and only be useful when used to calculate `done/total`, at which point you could just use `percentage`. In general `done` and `total` are not very useful as individual values because the client has no information what unit they are in and how to display them. It makes more sense to have `percentage` be for the progress bar, while letting the server decide how to format the `done` value through the `message` string.

And all that said, I would prefer to have _any_ progress proposal merged than bikeshedding these marginal differences forever :)
Let's keep it simple.

Xanewok:
> In general `done` and (...) are not very useful as individual values

I think `done` only makes sense from a protocol flow point of view - we get notified that no further messages will be received/sent with this ID.

Otherwise I could imagine server sending different LSP messages with different titles/messages but with `percentage: 100`, so it'd be awkward to use.

> And all that said, I would prefer to have any progress proposal merged than bikeshedding these marginal differences forever :)
Let's keep it simple.

Yes, please! +:100: to that.

felixfbecker:
@Xanewok note with `done` I was referring to the `{ done: number, total: number }`  pair suggested above as an alternative to `{ percentage: number }` not the `{ done: boolean }` complete indicator.

Xanewok:
Ah right, sorry for the confusion! it seems I need my midday dose of coffee :sleeping:

Xanewok:
As @felixfbecker mentioned, I'd like to move forward and merge this proposed extension. We shouldn't bikeshed this to death - it's simple enough and has been hanging for more than a year now. The implementation as-is worked good enough for the [RLS](https://github.com/rust-lang/rls) and now [Sourcegraph](https://github.com/sourcegraph/sourcegraph/pull/1375) implements it, so I'd say it's good to go at least as the proposed/experimental part of the spec.

[Rendered](https://github.com/Microsoft/vscode-languageserver-node/blob/c8ac910730a1f4499333a94f9e6cc25c0155818c/protocol/src/protocol.progress.proposed.md)

Final `ProgressParams`:
```typescript
export interface ProgressParams {
    /**
     * A unique identifier to associate multiple progress notifications with
     * the same progress.
     */
    id: string;

    /**
     * Mandatory title of the progress operation. Used to briefly inform about
     * the kind of operation being performed.
     * Examples: "Indexing" or "Linking dependencies".
     */
    title: string;

    /**
     * Optional, more detailed associated progress message. Contains
     * complementary information to the `title`.
     * Examples: "3/25 files", "project/src/module2", "node_modules/some_dep".
     * If unset, the previous progress message (if any) is still valid.
     */
    message?: string;

    /**
     * Optional progress percentage to display (value 100 is considered 100%).
     * If unset, the previous progress percentage (if any) is still valid.
     */
    percentage?: number;

    /**
     * Set to true on the final progress update.
     * No more progress notifications with the same ID should be sent.
     */
    done?: boolean;
}
```

VSCode reference implementation:
progress
(Dependencies were compiled quickly so here's a message mismatch, but it shows how the info is rendered for a corresponding LSP message)

Further tweaks to the VSCode rendering (including info tooltip hover etc.) should be done separately because of the somewhat restrictive, current `withProgress` API.

@dbaeumer could we merge this, please?

protocol/src/protocol.progress.proposed.ts
```diff
@@ -0,0 +1,62 @@
...
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import { NotificationType, NotificationHandler } from 'vscode-jsonrpc';
+
+export interface WindowProgressClientCapabilities {
+	/**
+	 * Experimental client capabilities.
+	 */
+	experimental: {
```
dbaeumer: This should be a normal capability and nothing experimental. Since it is proposed we are sill free to change it. We should extens `_ClientCapabilities` with a additional optional property `windows` with an optional property `progress`.

protocol/src/protocol.progress.proposed.md
```diff
@@ -0,0 +1,63 @@
...
+     * Examples: "3/25 files", "project/src/module2", "node_modules/some_dep".
+     * If unset, the previous progress message (if any) is still valid.
+     */
+    message?: string;
+
+    /**
+     * Optional progress percentage to display (value 100 is considered 100%).
+     * If unset, the previous progress percentage (if any) is still valid.
+     */
+    percentage?: number;
```
dbaeumer: I am not sure if percentage is a good idea since users could jump form 60 -> 50 which a lot of progress API's don't like. Could we used `worked`, `step` or `increment` instead.

dbaeumer: And should we defined that `-1` as a start value us infinite progress. I am also happy with an or type or an additional property.

protocol/src/protocol.progress.proposed.md
```diff
@@ -0,0 +1,63 @@
+#### Reporting server task progress
```
dbaeumer: Before we merge this in I would like that we at least think about how cancellation would work in this model since I think that this is a common pattern with progress. Options I see:

- we have a notification from the client to a server to signal cancellation based on the progress id
- we make the progress notification a request that returns whether this is canceled or not (not my favorite)
- we allow the progress to refer to a open request so that a client can map the things together. 
- we support that requests can provide a progress id which the server then uses to report progress on (my preferred options).

dbaeumer:
I am in favor to decouple this from what VS Code provides. However I have some general questions / suggestions which we should agree on first before merging as a proposed feature.

protocol/src/protocol.progress.proposed.md
felixfbecker: > I am not sure if percentage is a good idea since users could jump form 60 -> 50 which a lot of progress API's don't like. Could we used worked, step or increment instead.

My feeling is that just makes the API more complex. Usually the progress comes from some number of work done divided by some number of total work, e.g. bytes downloaded by value of `Content-Length` header. Having to calculate that, then diff it against the last reported progress percentage just because the API expects a delta makes it more complex on the server.

The client, if rendering a progress bar, ultimately needs to set the width of some element, which is easy using the latest absolute value. A delta API means the previous percentage value always needs to be kept around and incremented.

> which a lot of progress API's don't like

I don't know about other LSP clients, but which APIs are you thinking of?

Also some progress might actually sometimes be based on an estimate that may have to be corrected, decreasing the progress slightly.

> And should we defined that -1 as a start value us infinite progress. I am also happy with an or type or an additional property.

Could you explain what you mean by that? Isn't "Infinite progress" the same as "done", for which there is the `done: true` flag?

protocol/src/protocol.progress.proposed.md
felixfbecker: Things to consider:
- The request ID is usually not known to the implementation of a method, it's hidden in the RPC layer.
- A middleware proxy might rewrite request IDs (e.g. prefix them if managing many servers) but wouldn't be aware of also having to rewrite request IDs in the params of the progress notification.
- Progresses can often not correspond to requests, but to async actions like indexing and background initialisation, dependency installation, ... These may sometimes be cancellable (although usually you'd really want to just shut down the server), but cannot be cancelled by a request ID.

Option 1 would be a possible way forward in the future that considers these (although this proposal should not be blocked on that).

protocol/src/protocol.progress.proposed.md
Xanewok: > Could we used worked, step or increment instead.

I'd say if a client does not support "time-traveling" progress they can do a quick sanity check `this.progress = max(newProgress, thisProgress);`

> And should we defined that -1 as a start value us infinite progress.

Maybe it'd make sense to omit the `percentage` altogether in a server message when reporting an indefinite progress?

protocol/src/protocol.progress.proposed.md
Xanewok: Ditto what @felixfbecker said - I wouldn't want to block this and also it's worth thinking that progress report isn't always done against a client-issued request and doesn't have to support cancellation in every case.

protocol/src/protocol.progress.proposed.ts
Xanewok: Okay, done :+1:

dbaeumer:
Saying percentage === undefined is `infinite progress` is a good solution.

I am still not convinced that a time traveling progress monitor is something users appreciate. I usually dislike them very much :-)

And I do think that not thinking about how cancellation works in this scenario is a bad idea. If we have a progress API it should be usable for client initiated requests as well where cancellation does make sense. Consider the case where find all reference is based on an index which needs to be updated when first executed. If this takes a long time and a users doesn't want to wait for the result I think he appreciate being able to cancel them (Eclipse for example supports this nicely). I agree that this is not important for server initiated progress reporting. But I don't want to end up with two stories. I will make a proposal how that should work.

felixfbecker:
Even `Microsoft/pyright` is using custom progress messages:

https://github.com/Microsoft/pyright/blob/5323d14aa878ba7eaa7d5e0685f963ba108f039d/server/src/server.ts#L81-L90

Xanewok:
Is this blocked on CI and rebase? I'll fix this today. Should I change anything here before it's considered to be merged?

dbaeumer:
@Xanewok it is on my April plan. I will do some additional work to report progress per request as well since this is requested for other clients. But that is orthogonal to the report progress from server handled here. Having a mergable PR would definitelly be appreciated.

Xanewok:
@dbaeumer done, rebased and fixed the PR.

dbaeumer:
I looked at other tools how they report progress and all possible different kinds exists:

- precentage
- increment
- total work & worked

Since they all have pros and cons it doesn't really matter which one we pick. So I decided to stay with the percentage since it is indeed easy to compute. However I will add the following things:

- a client is allowed to not uopdate the progress if the number jumps pack
- no providing a value on first all will mean infinite progress. This can not be changed later on (e.g. switch from infinite to percentage and vice versa)
- I will also add support to attach progress to a request. I will do this by adding a progress id to the request params. This id can then be used in later calls.

dbaeumer:
I will merge the PR and then make the changes on master.

Xanewok:
Awesome, thank you! :tada: 

> * a client is allowed to not uopdate the progress if the number jumps pack
> * no providing a value on first all will mean infinite progress. This can not be changed later on (e.g. switch from infinite to percentage and vice versa)

Definitely agree with these, seems reasonable =)

PR #264, opened 2017-10-15
Specify that the workspace.applyEdit feature is enabled in the client

rcjsuen:
Not sure if this is the right way of going about it but it does address#263.


dbaeumer:
Thanks. This got lost when introducing the feature structure for the client.

PR #265, opened 2017-10-15
Renaming ProtocolConnetion to ProtocolConnection

object88:
Reading through the code, I found a typo in one of the object names.  It's unclear to me how far outside of this project this might extend, however.  Please let me know if there is any more work around this.


dbaeumer:
Thanks. To make this non breaking and will add a type def for the old name.

PR #271, opened 2017-11-10
StreamMessageReader: emit errors rather than exceptions for bad messages

### LINKED ISSUE #270: Stream read errors result in uncaught exceptions
Currently, message errors from StreamMessageReader's onData method end up as uncaught exceptions, which can be very difficult to handle when used in server contexts.

https://github.com/Microsoft/vscode-languageserver-node/blob/master/jsonrpc/src/messageReader.ts#L220

Would it be reasonable to convert such errors into error events (via `this.fireError`) so that the stream owner can perform shutdown / restart steps as necessary? I can submit a PR to do so. cc @ljw1004

hansonw:
Asynchronous exceptions are very difficult for end users to catch (in the case of server applications, this typically means crashing entirely). Instead of throwing exceptions, we'll emit an error event and stop processing any further messages (by removing all listeners and entering a disposed state).

I've added a test to cover some basic error cases, verifying that `onError` is called and that data ceases to be read.

Closes #270, contributed under CC0.

### DIFF
--- a/jsonrpc/src/messageReader.ts
+++ b/jsonrpc/src/messageReader.ts
@@ -8,7 +8,7 @@ import { Socket } from 'net';
 import { ChildProcess } from 'child_process';
 
 import { Message } from './messages';
-import { Event, Emitter } from './events';
+import { Disposable, Event, Emitter } from './events';
 import * as Is from './is';
 
 let DefaultSize: number = 8192;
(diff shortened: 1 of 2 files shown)


dbaeumer:
@hansonw you need to sign the CLA. Otherwise I will not be able to merge this.

dbaeumer:
I am closing the PR since the CLA didn't get signed.

Issue #274, opened 2017-11-18
IConnection#onWillSaveTextDocumentWaitUntil is never called


rictic:
I'm implementing a "fix lint errors on save" feature in my language server. The simplest way to do this seems to be listening for `IConnection#onWillSaveTextDocumentWaitUntil` and responding with some `TextEdit`s, but my handler is never called. Same thing for `IConnection#onWillSaveTextDocument` and `TextDocuments#onWillSaveWaitUntil`.

The `IConnection#onDidSaveTextDocument` handler is called, and the rest of my language server is working fine (autocomplete, diagnostics, code actions, etc etc).

I'm guessing that there's probably somewhere that I need to opt into receiving these events, but I've been poking around all morning and haven't found anything.

More details: 
  - my `textDocumentSync` capability is `TextDocuments#syncKind` which I think is `Full`
  - transport layer is stdio
  - VSCode 1.18.1, vscode-languageserver@3.5.0, vscode-languageclient@3.5.0, vscode-jsonrpc@3.5.0, vscode-languageserver-protocol@3.5.0

rictic:
Perfect! I totally missed that. Thank you!

I sent over a PR adding some more language docs.

dbaeumer:
@rcjsuen thanks for explaining this.

PR #275, opened 2017-11-18
Document that onWillSave handler capabilities reqs

rictic:
related to https://github.com/Microsoft/vscode-languageserver-node/issues/274


dbaeumer:
@rictic thanks for the PR. However this text is not 100% correct. Besides setting the capability on server initialize the server can always register dynamically later on. Can you add a sentence about this as well.

rictic:
+1, done

dbaeumer:
Thanks!

PR #276, opened 2017-11-23
Implement MarkdownContent and use it in documentation properties.

dbaeumer:


dbaeumer:
@aeschli changed the PR. I renamed MarkdownContent to MarkupContent and added a kind property. This will allow for better evolving (e.g. if there are only clients with HTML we could add a kind HTML). Although technically not a Markup language we could add a kind PlainText to allow sending simple strings as well using MarkupContent. But might be overkill since future API should always support Markdown.

types/src/main.ts
```diff
@@ -669,6 +669,61 @@ export namespace TextDocumentItem {
...
 
+/**
+ * Describes the content type that a client supports in various
+ * result literals like `Hover`, `ParameterInfo` or `CompletionItem`
+ */
+export namespace MarkupKind {
+	/**
+	 * Markdown is supported as a content format
+	 */
+	export const Markdown: 1 = 1;
```
aeschli: Maybe better a string so this can be extended without the need of change in our code.

aeschli: I would add PlainText as well so 'MarkdownContent' is the only needed type in representing formatted strings.
Not really needed for the completion documentation (there you can pass plain text as string), but for Hover there currently no way of just passing a plain text (it has to be MarkedString, and plain text needs to be escaped).
Especially good to have for future types with a documentation field.

protocol/src/protocol.ts
```diff
@@ -269,6 +268,13 @@ export interface TextDocumentClientCapabilities {
 			 * Client supports commit characters on a completion item.
 			 */
 			commitCharactersSupport?: boolean
+
+			/**
+			 * Client supports the follow content formats for the documentation
+			 * property beside plain text. The value must be an or value of the
+			 * constants defined in `ContentFormat`.
+			 */
+			documentationFormat?: number;
```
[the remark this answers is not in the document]
aeschli: The comment should also mention what the default client capability is.
> If no `documentationFormat` is defined, `CompletionProposal.documentation` can only be of type `string` with plain text content.

dbaeumer:
Thanks for the good comments! Make all sense. Will adjust accordingly.

dbaeumer:
Merged into master using command line.

peni4142:
Is it right that I get syntax-highliting instead of styled text?

"# Wurzel-Element. Kann nur einmal und auch nur am Anfang eines Scripts definiert werden.\n\n..."

hover

rcjsuen:
@peni4142 I suggest you open a separate issue with more details about your problem.

PR #280, opened 2017-11-27
Support extensible Symbol and CompletionItem kinds

dbaeumer:


dbaeumer:
@aeschli can you please have a look and provide your feedback.

PR #284, opened 2017-12-05
Return value after executing command request

danixeee:
Hello,
We have discussed about this issue here:
https://github.com/Microsoft/language-server-protocol/issues/329

With this change, I am able to fetch values returned from server.


dbaeumer:
@danixeee this is a very good catch. Thanks for providing the PR.

PR #295, opened 2017-12-22
Changes for supporting resource level changes

gorkem:
LSP proposal that adds resourceChanges to WorkspaceEditor to allow manipulation of files and folders. This pull request only includes the proposed changes. It is not possible to implement these changes for vscode due to [lack of APIs](https://github.com/Microsoft/vscode/issues/10659). This PR addresses https://github.com/Microsoft/language-server-protocol/issues/272


dbaeumer:
Thanks a lot for the proposal. When creating a new file shouldn't we provide an option for the content. I would ensure that we can create a file with a predefined content. Since stuff is async we couldn't otherwise.

And do we need to change something for the response to provide better error information. For example a file couldn't be created?

protocol/src/protocol.resourceChanges.proposed.md
```diff
@@ -0,0 +1,82 @@
...
+
+_WorkspaceEdit_
+
+The following additions are proposed for __WorkspaceEdit__.
+
+
+_resourceChanges_
+
+A workspace edit can optionally provide resource changes. If present resource changes are applied before the `changes/documentChanges` has been applied
+allowing the `change/documentChanges` to work on affected files. The changes are applied in the order that they are supplied.
```
rcjsuen: This `changes/documentChanges` confused me as it made it seem like a notification or a request (like a `textDocument/didOpen`). Perhaps it might be better to spell it out specifically with "`changes` or `documentChanges`" instead?

protocol/src/protocol.resourceChanges.proposed.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+## ResourceChange
+Changes to a folder or file on the workspace.
+
+```ts
+export interface ResourceChange{
+
+  /**
+  * The resource change operation type
```
rcjsuen: Period at the end?

protocol/src/protocol.resourceChanges.proposed.md
```diff
@@ -0,0 +1,82 @@
...
+  */
+  changeType: ResourceChangeType
+
+  /**
+  * The resource.
+  *
+  * If change type is CREATE a new file is created on the uri.
+  * This operation will not overwrite any existing resources and will fail.
```
rcjsuen: I think this needs to be cleared up to state that the operation will fail if the path conflicts with an existing resource.

protocol/src/protocol.resourceChanges.proposed.md
```diff
@@ -0,0 +1,82 @@
...
+  *
+  * If change type is DELETE the resource is deleted if it exists. If the
+  * uri is a folder this operation deletes all its children.
```
rcjsuen: Is it considered a success or a failure if you try to delete a file or folder that does not exist?

protocol/src/protocol.resourceChanges.proposed.md
```diff
@@ -0,0 +1,82 @@
...
+  * If change type is MOVE this uri is considered the source resource for the
+  * move and must point to an existing resource
+  */
+  path: string
+
+  /**
+  * The new uri for the resource pointed by path.
+  * Required only if change type is MOVE. Must be compatible with the path
+  * uri ie. must be a file uri if path is a file uri.
+  *
```
rcjsuen: Unnecessary extra space.

protocol/src/protocol.resourceChanges.proposed.ts
```diff
@@ -0,0 +1,46 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Red Hat Inc and others. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict'
+
+export interface ResourceChange{
+
+  /**
+  * The resource change operation type
```
rcjsuen: See above.

protocol/src/protocol.resourceChanges.proposed.ts
```diff
@@ -0,0 +1,46 @@
...
+  */
+  changeType: ResourceChangeType
+
+  /**
+  * The resource.
+  *
+  * If change type is CREATE a new file is created on the uri.
+  * This operation will not overwrite any existing resources on this uri.
```
rcjsuen: See above.

protocol/src/protocol.resourceChanges.proposed.ts
```diff
@@ -0,0 +1,46 @@
...
+  *
+  * If change type is DELETE the resource is deleted if it exists. If the
+  * uri is a folder this operation deletes all its children.
```
rcjsuen: See above.

protocol/src/protocol.resourceChanges.proposed.ts
```diff
@@ -0,0 +1,46 @@
...
+  * If change type is MOVE this uri is considered the source resource for the
+  * move and must point to an existing resource
+  */
+  path: string
+
+  /**
+  * The new uri for the resource pointed by path.
+  * Required only if change type is MOVE. Must be compatible with the path
+  * uri ie. must be a file uri if path is a file uri.
+  *
```
rcjsuen: See above.

protocol/src/protocol.resourceChanges.proposed.md
```diff
@@ -0,0 +1,82 @@
...
+The client sets the following capability if it is supporting resource Changes.
+
+```ts
+/**
+ * The workspace client capabilities
+ */
+workspace: {
+  workspaceEdit?: {
+		  /**
+		  * The client supports resource changes in `WorkspaceEdit`s
```
rcjsuen: Need a period at the end.

protocol/src/protocol.resourceChanges.proposed.md
```diff
@@ -0,0 +1,82 @@
+# Resource Changes
+
+_Client Capabilities_:
+
+The client sets the following capability if it is supporting resource Changes.
```
rcjsuen: Is there a particular reason as to why "Changes" is capitalized?

protocol/src/protocol.resourceChanges.proposed.ts
```diff
@@ -0,0 +1,46 @@
...
+  */
+  newPath?: string
+
+}
+
+/**
+ * A resource change type
+ */
+export declare namespace ResourceChangeType{
```
rcjsuen: Same as above.

protocol/src/protocol.resourceChanges.proposed.ts
```diff
@@ -0,0 +1,46 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Red Hat Inc and others. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict'
+
+export interface ResourceChange{
```
rcjsuen: Same as above.

protocol/src/protocol.resourceChanges.proposed.md
```diff
@@ -0,0 +1,82 @@
...
+  *
+  */
+  newPath?: string
+
+}
+
+/**
+ * A resource change type
+ */
+export declare namespace ResourceChangeType{
```
rcjsuen: Should add a space after ResourceChangeType to be consistent with the other files.

protocol/src/protocol.resourceChanges.proposed.md
```diff
@@ -0,0 +1,82 @@
...
+  resourceChanges?: ResourceChange[]
+
+}
+```
+
+## ResourceChange
+Changes to a folder or file on the workspace.
+
+```ts
+export interface ResourceChange{
```
rcjsuen: Should add a space after ResourceChange to be consistent with the other files.

protocol/src/protocol.resourceChanges.proposed.md
```diff
@@ -0,0 +1,82 @@
...
+
+_resourceChanges_
+
+A workspace edit can optionally provide resource changes. If present resource changes are applied before the `changes/documentChanges` has been applied
+allowing the `change/documentChanges` to work on affected files. The changes are applied in the order that they are supplied.
+
+```ts
+interface  WorkspaceEdit{
+
+  resourceChanges?: ResourceChange[]
```
rcjsuen: Since we talk about `changes` and `documentChanges` in a `WorkspaceEdit`, would it be better to also list those properties again here? I'm not sure what would be best since we're extending the interface here (by adding optional properties) without actually creating a new interface. Thoughts, @dbaeumer?

protocol/src/protocol.resourceChanges.proposed.md
```diff
@@ -0,0 +1,82 @@
...
+The following additions are proposed for __WorkspaceEdit__.
+
+
+_resourceChanges_
+
+A workspace edit can optionally provide resource changes. If present resource changes are applied before the `changes/documentChanges` has been applied
+allowing the `change/documentChanges` to work on affected files. The changes are applied in the order that they are supplied.
+
+```ts
+interface  WorkspaceEdit{
```
rcjsuen: Should add a space after WorkspaceEdit to be consistent with the other files.

gorkem:
@dbaeumer I thought the initial content could be provided via `changes` or `documentChanges` filed on `WorkspaceEdit`. `resourceChanges` are defined to run before TextEdits are applied. Do you thing this would not be enough?

gorkem:
Thanks @rcjsuen very helpful review. I think I have addressed most of them.

gorkem:
@dbaeumer On error information with the response: I think we need to address that on `workspace/applyEdit` and `textDocument/rename` which are the users of `WorkspaceEdit`.

gorkem:
Updated the proposed `ResourceChanges` so that it no longer requires additional enumeration. vscode's corresponding implementation gives a chance to define an (almost) ordered execution of the text and resource changes but I could not see a way to define it on `WorkspaceEdit` without breaking or completely defining a new field.

dbaeumer:
@gorkem @rcjsuen   thanks again for providing this. Please note that VS Code 1.20 will not ship with a resource change API due to problems we discovered with the API during the test pass (see the corresponding vscode issue).

We discussed in detail the order issue and it is our belief that we need one array describing all changes to make the behavior correct.

For the LSP I would do the following: if resource changes are supported the workspace edit result has a property resourceChanges which are either a rename, move, delete or content change.

gorkem:
@dbaeumer did a small change to allow ordered `resourceChanges`, please take a look if it looks OK, I want to proceed with the implementation.

spoenemann:
The type
```
interface  WorkspaceEdit {
    resourceChanges?: (ResourceChange | TextDocumentEdit)[];
}
```
is hard to realize in Java because both ResourceChange and TextDocumentEdit are objects, so the only way to create the correct class instance is to look ahead in the JSON token stream, which means we cannot use the default stream parsing approach of the Gson library. See eclipse/lsp4j#173.

I'd propose to split this:
```
interface  WorkspaceEdit {
    resourceChanges?: ResourceChange[];
    textEdits?: TextDocumentEdit[];
}
```

dbaeumer:
@gorkem regarding: https://github.com/Microsoft/vscode-languageserver-node/pull/295#issuecomment-353602155. I think that is not a good idea (e.g. assuming that resource changes are applied before text edits). This would mean it is harder to express to modify the content of a file and then move. The statement should be that at the edits should be one array (but we need to address @spoenemann concern) and that they are executed in order in which they are provided.

gorkem:
@dbaeumer I am not sure if there is a great benefit of having a single ordered list of resource changes and text edits. A WorkspaceEdit is essentially a single unit and it should be fully applied or failed.  I think the proposed two arrays is equivalent to having a single array.  
What is the use case that you have in mind that requires a single ordered list?

dbaeumer:
The case is you create a file and then afterwards want to edit its content.

gorkem:
The original proposal was to execute the resource changes before `TextEdits` which covers the create file case. The only case I see is modify contents and delete case which would become delete and modify in which case the implementations should drop the `TextEdits`

spoenemann:
An alternative solution for the Java problem would be to let ResourceChange extend `Partial< TextDocumentEdit>` so the WorkspaceEdit can be written as
```
interface  WorkspaceEdit {
    resourceChanges?: ResourceChange[];
}
```

spoenemann:
...or simply redeclare the properties of TextDocumentEdit as optional in ResourceChange.

dbaeumer:
I see the problem with mapping JSON to Jave however the current proposal requires IMO a lot of documentation since it is not obvious what it does. Problems I see:

- the textDocument property for a resource level operation (e.g. create, move, delete) doesn't need to be a `VersionedDocumentIdentifier`. It might make sense for move or delete but i might be hard to servers to construct this we we create a file, then change it and then move it.
- it gives the impression that only text documents could be move, created or deleted

Would a type attribute on the resource chnage help the Java library to implement the right switch early instead if having to check properties ?

spoenemann:
> Would a type attribute on the resource chnage help the Java library to implement the right switch early instead if having to check properties ?

But then we would still need to parse the type attribute before we can decide which class instance to create?

To be clear: parsing based on attributes / properties is possible also in Java / [LSP4J](https://github.com/eclipse/lsp4j), it just increases the effort for implementing such a parser and is slower at runtime. If the best solution in terms of LSP data structure design requires this, we'll manage to implement it, but it would be good to avoid it if there are acceptable alternatives for the data structure design.

dbaeumer:
@spoenemann thanks for the clarification.

gorkem:
Since this has stalled a bit because I got new responsibilities, @tsmaeder will be taking over this PR, so stay tuned

spoenemann:
@dbaeumer @gorkem @tsmaeder I found a solution for parsing object type alternatives in eclipse/lsp4j#230 that is acceptable in terms of additional required code for each such case. From the perspective of LSP4J, we can leave the original declaration

```
interface  WorkspaceEdit {
    resourceChanges?: (ResourceChange | TextDocumentEdit)[];
}
```

as it is, since we can now parse it correctly.

dbaeumer:
@spoenemann thanks

PR #298, opened 2017-12-25
Change TextDocument's getText to take an optional range

### LINKED ISSUE #255: TextDocument.getText() should accept Range as an optional parameter
[TextDocument.getText](https://github.com/Microsoft/vscode-languageserver-node/blob/master/types/src/main.ts#L1355) method should accept Range as an optional parameter. This refers to https://code.visualstudio.com/docs/extensionAPI/vscode-api#_a-nametextdocumentaspan-classcodeitem-id38textdocumentspan

rcjsuen:
I've changed `TextDocument` so that its `getText()` function is now a `getText(range?: Range)`. This fixes #255 and allows a client to extract a substring from the document instead of the entire range. If any positions are detected to be outside the bounds of the document (an offset less than zero or an offset greater than the length of the document), they will be adjusted to accordingly to fit within the bounds of the document.

### DIFF
--- a/types/src/main.ts
+++ b/types/src/main.ts
@@ -1460,11 +1460,14 @@ export interface TextDocument {
 	readonly version: number;
 
 	/**
-	 * Get the text of this document.
+	 * Get the text of this document. A substring can be retrieved by
+	 * providing a range.
 	 *
-	 * @return The text of this document.
+	 * @param range The interested range within the document to return.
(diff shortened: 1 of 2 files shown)


dbaeumer:
@aeschli can you please review and follow up?

rcjsuen:
Thank you for taking the time to look at this, Martin!

PR #308, opened 2018-02-11
New Completion Trigger Kind

ramya-rao-a:


rcjsuen:
If a new constant is going to be added, please update the [specification](https://github.com/Microsoft/language-server-protocol/) also.

aeschli:
The changes look. The contextSupport is new for 4.0, so no need to add an additional client capability.

rcjsuen:
@ramya-rao-a The current specification of the LSP is v3 but it is still receiving updates on a regular basis. In fact, `CompletionTriggerKind` did not originally exist in v3 but was only recently added in [November](https://github.com/Microsoft/language-server-protocol/commit/3ee4463067be6a357d077b45f421e4985db47f0b).

In any case, I have opened Microsoft/language-server-protocol#403 to update the protocol to include the new `TriggerForIncompleteCompletions` you added.

dbaeumer:
@ramya-rao-a in general the spec needs to be updated as well when we do something in code :-).

@rcjsuen thanks for updating the spec

ramya-rao-a:
@rcjsuen @dbaeumer Thanks, I'll keep that in mind for the next time.

PR #320, opened 2018-03-16
Fix registration parameters for workspace/didChangeWatchedFiles

rcjsuen:
It's currently not possible to register for `workspace/didChangeWatchedFiles` notifications because it has `void` set as a parameter. We should be using `DidChangeWatchedFilesRegistrationOptions` instead so that servers can define the interested files when sending its `client/registerCapability` request to the client.


dbaeumer:
Thanks for the PR

PR #322, opened 2018-03-18
Add middleware to intercept textDocument/publishDiagnostics

rcjsuen:
It is currently not possible for a client to inspect the contents of the `textDocument/publishDiagnostics` notifications from the server before letting VS Code handle them. This pull request introduces a middleware function to allow clients to intercept the contents of the notification before passing them off to VS Code.

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -7,7 +7,7 @@
 import {
 	workspace as Workspace, window as Window, languages as Languages, commands as Commands,
 	TextDocumentChangeEvent, TextDocument, Disposable, OutputChannel,
-	FileSystemWatcher as VFileSystemWatcher, DiagnosticCollection, Uri, ProviderResult,
+	FileSystemWatcher as VFileSystemWatcher, DiagnosticCollection, Diagnostic as VDiagnostic, Uri, ProviderResult,
 	CancellationToken, Position as VPosition, Location as VLocation, Range as VRange,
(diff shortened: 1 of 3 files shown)


dbaeumer:
@rcjsuen thanks for the PR. There is one thing I am not sure of how we should do it: the middleware API currently uses VS Code API types only. With this PR it will surface LSP types. Although this allows to intercept the notification as early as possible I would nevertheless change the types to VS Code API types. In the implementation that means we first convert the LSP data into VS Code data and then call the middleware.

rcjsuen:
@dbaeumer Hi Dirk, thanks for the review.

Should I still use `NextSignature` then? VS Code's `Diagnostic` does not contain information URI information so I am forced to change...
```TypeScript
publishDiagnostics?: NextSignature<PublishDiagnosticsParams, void>;
```
...to...
```TypeScript
publishDiagnostics?: NextSignature<Uri, VDiagnostic[], void>;
```
But this won't compile because `NextSignature` only takes two parameters.

Or should I instead write a new function like the following and just introduce a new `HandleDiagnosticsSignature` interface type?
```TypeScript
handleDiagnostics?: (this: void, uri: Uri, diagnostic: VDiagnostics[], next: HandleDiagnosticsSignature) => void;
```

dbaeumer:
@rcjsuen I would go with the `HandleDiagnosticsSignature` approach.

rcjsuen:
@dbaeumer Thank you. I've force pushed a new change to my branch. Feel free to take a look whenever you have the time. I'm sure you're a busy man.

dbaeumer:
Thanks for the PR.

PR #323, opened 2018-03-20
client: Fixed issue where executeCommand result from language server was always undefined

joshtynjala:
If the command is implemented inside the VSCode extension, it works correctly:

``` typescript
vscode.commands.registerCommand("myCommand", () =>
{
	return true;
});
```

``` typescript
vscode.commands.executeCommand("myCommand").then((result) =>
{
	console.log("myCommand result:", result);
});
```

However, if the command is implemented in the language server instead (in my case, I'm using eclipse/lsp4j with Java), the result is always `undefined`. With this change, it should use the correct result instead of `undefined`

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -2162,7 +2162,7 @@ class ExecuteCommandFeature implements DynamicFeature<ExecuteCommandRegistration
 						arguments: args
 					};
 					return client.sendRequest(ExecuteCommandRequest.type, params).then(
-						undefined,
+						(result) => result,
 						(error) => {
 							client.logFailedRequest(ExecuteCommandRequest.type, error);
 						}


dbaeumer:
@joshtynjala this is very strange since a not provided onFulfilled function results in a promise resolved to its previous value. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise.

So `p.then(undefined, error => { .... })` is equivalent to `p.then(result => result, error => { ... })`. Have you tested that this will solve the problem for you?

dbaeumer:
```
new Promise((resolve, reject) => {
	resolve(10);
}).then(undefined, (error) => {
	return -10;
}).then(value => {
	console.log(value);
});
```

Prints correctly 10 to the console.

dbaeumer:
I tested the command even with a look through `executeCommand` and results are returned correctly. Can you provide me with a test case very this fails for you?

joshtynjala:
> Have you tested that this will solve the problem for you?

@dbaeumer When you say "this", are you referring to the change that I made in my pull request? Yes, when I test my extension using this change, the value returned by the command implemented on my language server is no longer `undefined`.

Could it be that the Promise is not behaving correctly because the language server is implemented in Java and communicating over the language server protocol? Maybe the JavaScript VM is not treating it like a regular Promise.

joshtynjala:
In case it wasn't clear from my original post, I'll point out again that that I'm using https://github.com/eclipse/lsp4j to implement a language server in Java, and I'm registering some commands that are implemented in the language server. When I use vscode.commands.registerCommand to implement the command inside my VSCode extension, everything works fine. When I implement the command in Java with LSP4J, that's when the result seems to get lost somewhere.

dbaeumer:
The LSP protocol is JSON-RPC based. So promises are implemented on top of this using the responses returned from a request. So the only reason I can think of is that the Java server is not sending the right response.

As outlined in comment https://github.com/Microsoft/vscode-languageserver-node/pull/323#issuecomment-378618574 I tested the whole loop and results returned from the server correctly flow back to the client. My server is implemented in Node though. But it proves that the client library works as expected. 

@joshtynjala can you enable logging on the client side to see what messages flow in this case. Alternatively you can debug it. Responses are handled here: https://github.com/Microsoft/vscode-languageserver-node/blob/master/jsonrpc/src/main.ts#L613

joshtynjala:
@dbaeumer Hmmm... I just discovered that I'm now seeing the correct result passed back from my command instead of `undefined`.

Last week, I happened to delete *node_modules* for my extension and I reinstalled everything from scratch. I can only assume that an older version of one of the dependencies got updated, and somehow, that has fixed my issue.

Sorry for wasting your time! Closing.

dbaeumer:
No problem.

PR #324, opened 2018-03-21
Support for multi-location diagnostics

kieferrm:
This PR implements support for multi-location diagnostics.

I used https://github.com/kieferrm/lsp-dev-setup to setup the experimentation environment.

### DIFF
--- a/protocol/src/protocol.ts
+++ b/protocol/src/protocol.ts
@@ -263,7 +263,7 @@ export interface TextDocumentClientCapabilities {
 		 * The client supports did save notifications.
 		 */
 		didSave?: boolean;
-	}
+	};
 
 	/**
 	 * Capabilities specific to the `textDocument/completion`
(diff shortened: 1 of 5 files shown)


protocol/src/protocol.ts
```diff
@@ -502,6 +502,11 @@ export interface _ClientCapabilities {
 	 */
 	textDocument?: TextDocumentClientCapabilities;
 
+	/**
+	 * Whether the client supports diagnostics with related information.
+	 */
+	diagnosticRelatedInformation?: boolean;
```
rcjsuen: I don't think that this property belongs here. Since diagnostics are in the form of a `textDocument/publishDiagnostics` request, perhaps it should go there instead?

kieferrm: Makes sense. I moved the capability. Thanks!

dbaeumer:
@kieferrm merge the PR, but didn't publish yet. Will wait for the 1.22 release. I also updated the specification :-)

PR #332, opened 2018-04-05
Add deprecated boolean property to CompletionItem and SymbolInformation

rcjsuen:
I've added a new boolean `deprecated` property both `CompletionItem` and `SymbolInformation`.

See Microsoft/language-server-protocol#139.

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -1303,7 +1303,7 @@ class CompletionItemFeature extends TextDocumentFeature<CompletionRegistrationOp
 		let completion = ensure(ensure(capabilites, 'textDocument')!, 'completion')!;
 		completion.dynamicRegistration = true;
 		completion.contextSupport = true;
-		completion.completionItem = { snippetSupport: true, commitCharactersSupport: true, documentationFormat: [MarkupKind.Markdown, MarkupKind.PlainText] };
+		completion.completionItem = { snippetSupport: true, commitCharactersSupport: true, documentationFormat: [MarkupKind.Markdown, MarkupKind.PlainText], deprecatedSupport: false };
 		completion.completionItemKind = { valueSet: SupportedCompletionItemKinds };
 	}
 
--- a/protocol/src/protocol.ts
+++ b/protocol/src/protocol.ts
@@ -299,6 +299,11 @@ export interface TextDocumentClientCapabilities {
 			 * property. The order describes the preferred format of the client.
 			 */
 			documentationFormat?: MarkupKind[];
+
+			/**
+			 * Client supports the deprecated property on a completion item.
+			 */
+			deprecatedSupport?: boolean;
 		},
 
 		completionItemKind?: {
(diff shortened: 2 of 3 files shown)


rcjsuen:
@dbaeumer Thank you for the review, Dirk. I understand what you mean. Does this look okay?

```TypeScript
export interface TextDocumentClientCapabilities {
  completion?: {
    completionItem?: {
      /**
       * Client supports the deprecated property on a completion item.
       */
      deprecatedSupport?: boolean
    }
  }
}
```

rcjsuen:
I tried adding a capability but the Travis CI build didn't go through. Will investigate some more...
```
client/src/client.ts(1306,152): error TS2322: Type '{ snippetSupport: true; commitCharactersSupport: true; documentationFormat: MarkupKind[]; depreca...' is not assignable to type '{ snippetSupport?: boolean | undefined; commitCharactersSupport?: boolean | undefined; documentat...'.
  Object literal may only specify known properties, and 'deprecatedSupport' does not exist in type '{ snippetSupport?: boolean | undefined; commitCharactersSupport?: boolean | undefined; documentat...'.
```

rcjsuen:
I think the symlinking part of the build isn't quite working out for me.

https://github.com/Microsoft/vscode-languageserver-node/blob/ce3483119b72a751c28da4c5d578a0149011e130/.travis.yml#L22-L29

rcjsuen:
As VS Code doesn't read the `deprecated` value, the `deprecatedSupport` capability should be set to `false`, correct?

dbaeumer:
@rcjsuen correct.

rcjsuen:
@dbaeumer Thanks, Dirk. I have updated this pull request.

Since VS Code doesn't support this feature I didn't bother updating `ProtocolCompletionItem` based on your [earlier comment](https://github.com/Microsoft/vscode-languageserver-node/pull/332#issuecomment-379243203) of letting the property become "lost" between calls. If that was a misinterpretation on my part, I can of course update this pull request and change `ProtocolCompletionItem` so that it will preserve the property between `textDocument/completion` and `completionItem/resolve` requests. Just let me know, thanks!

dbaeumer:
@rcjsuen thanks for the PR. I merged it in. Think about it we should try to preserve the value also VS Code doesn't support deprecation. But you can do this with a separate PR.

dbaeumer:
@rcjsuen can you please update the specification as well. This is necessary so that we can release a new version.

rcjsuen:
Hi, @dbaeumer, thanks for merging this in. I've opened Microsoft/language-server-protocol#465 for updating the spec.

Regarding your earlier comment, please confirm the following.

1. Change VS Code side's `deprecatedSupport` to return `true`.
2. Add a new `deprecated` property to `ProtocolCompletionItem` and tweak the code accordingly so that it will be preserved across `textDocument/completion` and `completionItem/resolve` requests.

If the above is correct I will open a new PR. If I misunderstood anything, please let me know. Thanks!

dbaeumer:
@rcjsuen regarding the questions:

1. no we should keep false here since VS Code will not do anything with the value
2. correct.

rcjsuen:
Thank you, @dbaeumer. I've opened #347 based on your response.

PR #336, opened 2018-04-12
Fix documentLink/resolve by introducing a data field

rcjsuen:
The current definition of a `DocumentLink` makes resolving a link with `documentLink/resolve` essentially impossible. With only a `range` field available, it is not possible for the server to deduce what the originating content is as it does not have any information as to which text document the link is for.

To correct this, a new `data` field has been introduced to `DocumentLink` that servers can fill in with some whatever it needs when initially fulfilling the `textDocument/documentLink` request. This way, a subsequent `documentLink/resolve` request will have the necessary contextual information available to resolve the link and return back a URI to use in the `target` field.

### DIFF
--- a/client/src/codeConverter.ts
+++ b/client/src/codeConverter.ts
@@ -9,6 +9,7 @@ import * as proto from 'vscode-languageserver-protocol';
 import * as Is from './utils/is';
 import ProtocolCompletionItem from './protocolCompletionItem';
 import ProtocolCodeLens from './protocolCodeLens';
+import ProtocolDocumentLink from './protocolDocumentLink';
 import { MarkdownString } from 'vscode';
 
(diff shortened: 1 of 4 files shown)


dbaeumer:
@rcjsuen can you please update the spec. E.g. have 

```ts
	/**
	 * A data entry field that is preserved on a document link between a
	 * DocumentLinkRequest and a DocumentLinkResolveRequest.
	 */
	data?: any
```

in there as well.

rcjsuen:
Thank you for merging in the PR, Dirk. I've opened Microsoft/language-server-protocol#464 to add the `data` property to the spec.

dbaeumer:
@rcjsuen thanks

PR #338, opened 2018-04-13
Replace symlinking in Travis build with npm scripts

rcjsuen:
In npm 5 (which is included in Node.js 8), symbolic links in the `node_modules` folder are now replaced by what's in the npm registry. This means that symbolic links that were originally created in `.travis.yml` prior to invoking `npm install` were getting deleted and replaced by packages from the npm registry. As a result, we now need to create the symbolic links **after** `npm install` has completed.

Pushing the linking magic down to an npm script instead of only performing this as part of the CI build also allows it to be reused for local development. This has the advantage of making local development more closely match what is happening in the CI builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,15 +20,9 @@ before_install:
     fi
 
 install:
-  - cd protocol && mkdir node_modules && cd ..
-  - cd protocol/node_modules && ln -s ../../jsonrpc vscode-jsonrpc && cd ../..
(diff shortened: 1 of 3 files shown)


dbaeumer:
@rcjsuen thanks a lot for the PR. Very welcome.

dbaeumer:
I change the script to pure JS since there are cases where sub modules are hoisted and the links are missed.

rcjsuen:
@dbaeumer Ah, okay. Sorry I didn't catch that. Thanks for fixing the build problem!

PR #341, opened 2018-04-17
Resolve #340

### LINKED ISSUE #340: ResponseError.data may be undefined
In [`jsonrpc/src/messages.ts`](https://github.com/Microsoft/vscode-languageserver-node/blob/master/jsonrpc/src/messages.ts#L82-L94):

```TypeScript
export class ResponseError<D> extends Error {

	public readonly code: number;
	public readonly data: D;

	constructor(code: number, message: string, data?: D) {
		super(message);
		this.code = is.number(code) ? code : ErrorCodes.UnknownErrorCode;
		if (data !== void 0) {
			this.data = data;
		}
		Object.setPrototypeOf(this, ResponseError.prototype);
	}
	// ...
```

In the ctor, `data` is of type `D | undefined` and is only assigned to `this.data` if it is not `undefined`. If I read this correctly - does this mean that if `data?: D` is not passed, `this.data` also remains undefined? The spec says about [`interface ResponseError<D>`](https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#response-message) that the `data` can also be undefined.

When compiling with `strictPropertyInitialization`, this is also marked as an error.

There are also several compiler errors in `server/src/main.ts`, e.g. in [_RemoteWorkspaceImpl](https://github.com/Microsoft/vscode-languageserver-node/blob/master/server/src/main.ts#L812-L842) which actually does not check `this._connection` before `this._connection.sendRequest`, but does check in `get connection()`.

I'd file a small PR of these things, if this is considered a problem.

nikeee:
This PR resolves #340

### DIFF
--- a/jsonrpc/src/messages.ts
+++ b/jsonrpc/src/messages.ts
@@ -82,26 +82,21 @@ export interface ResponseErrorLiteral<D> {
 export class ResponseError<D> extends Error {
 
 	public readonly code: number;
-	public readonly data: D;
+	public readonly data?: D;
 
 	constructor(code: number, message: string, data?: D) {
 		super(message);
 		this.code = is.number(code) ? code : ErrorCodes.UnknownErrorCode;
-		if (data !== void 0) {
-			this.data = data;
-		}
+		this.data = data;
 		Object.setPrototypeOf(this, ResponseError.prototype);
 	}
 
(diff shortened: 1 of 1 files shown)


dbaeumer:
@nikeee same here. Please use `public readonly data: D | undefined`

nikeee:
Done.

PR #343, opened 2018-04-26
Jump empty header lines

adelarsq:
This change provide a way to jump empty header lines.


dbaeumer:
@adelarsq The change is fine for me minus some formatting. But why it is necessary. Do you receive headers with empty lines ?

adelarsq:
@dbaeumer I notice that was a error on the setup for the Java language server. I will close since is no longer necessary.

dbaeumer:
@adelarsq thanks for reaching back.

PR #346, opened 2018-05-05
[WIP] Proposal for type hierarchy

gorkem:
Add new textDocument/subTypes and textDocument/superTypes
requests. Introduces SymbolNode.

Fixes https://github.com/Microsoft/language-server-protocol/issues/136

### DIFF
--- a/protocol/protocol.typeHierarchy.proposed.md
+++ b/protocol/protocol.typeHierarchy.proposed.md
@@ -1,0 +1,77 @@
+
+#### Type Hierarchy
+
+Many language support inheritance and LSP provides retrieving the hierarchy information with the following requests.
+
+
+The `textDocument/subTypes` request is sent from the client to the server to resolve resolve the subtypes of a symbol at a given text document position. Returns the direct subtypes in no particular order.
+
+_Client Capabilities_:
+
+```ts
+  TextDocumentClientCapabilities {
+    /**
+     * Capabilities specific to the `textDocument/subTypes`
+     *  and `textDocument/superTypes
+     */
+    typeHierarchy?: {
+    /**
+     * Whether implementation supports dynamic registration.
+     */
+    dynamicRegistration?: boolean;
+  };
+```
+
+_Server Capabilities_:
+```ts
+ServerCapabilities{
+  /**
+   * The server provides type hierarchy information
+   */
+  typeHierarchyProvider?: boolean
+}
+```
+
+```ts
+/**
+ * Represents hierarchical information about programming constructs like variables, classes,
+ * interfaces etc.
+ */
+export interface SymbolNode extends SymbolInformation {
+  /**
+   * Immediate descendants of this node
+   */
+  descendants? : SymbolNode[]
+  /**
+   * true if this node has children. hasChildren can be
+   * true even when the descendants are empty. In that case
+   * the clients should do a call to retrieve the descendants.
+   */
+  hasChildren: boolean
+}
+```
+
+_Request_:
+
+method: ‘textDocument/subTypes’
+params: TextDocumentPositionParams
+
+_Response_
+
+* result: `SymbolNode[] | null`
+* error: code and message set in case an exception happens during the 'textDocument/subTypes' request
+
+
+
+The `textDocument/superTypes` request is sent from the client to the server to resolve resolve the supertypes of a symbol at a given text document position. Returns all the supertypes in bottom-up order.
+
+_Request_:
+
+method: ‘textDocument/superTypes’
+params: TextDocumentPositionParams
+
+_Response_
+
+* result: `SymbolNode[] | null`
+* error: code and message set in case an exception happens during the 'textDocument/superTypes' request
+
--- a/protocol/src/protocol.typeHierarchy.proposed.ts
+++ b/protocol/src/protocol.typeHierarchy.proposed.ts
@@ -1,0 +1,65 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Red Hat Inc and others . All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import { RequestType } from 'vscode-jsonrpc';
+import {SymbolInformation} from 'vscode-languageserver-types';
+import {TextDocumentPositionParams, TextDocumentRegistrationOptions} from './protocol'
+
+export interface TypeHierarchyClientCapabilities {
+	/**
+	 * The text document client capabilities
+	 */
+	textDocument?: {
+		/**
+		 * Capabilities specific to the `textDocument/subTypes` and `textDocument/superTypes`
+		 */
+		typeHierarchy?: {
+			/**
+			 * Whether implementation supports dynamic registration.
+			 */
+			dynamicRegistration?: boolean;
+		};
+	}
+}
+
+export interface TypeHierarchyServerCapabilities {
+	/**
+	 * The server provides type hierarchy information
+	 */
+	typeHierarchyProvider?: boolean
+}
+
+/**
+ *
+ */
+export namespace SubTypesRequest {
+  export const type = new RequestType<TextDocumentPositionParams, SymbolNode[], void, TextDocumentRegistrationOptions>('textDocument/subTypes');
+}
+
+/**
+ *
+ */
+export namespace SuperTypesRequest {
+  export const type = new RequestType<TextDocumentPositionParams, SymbolNode[], void, TextDocumentRegistrationOptions>('textDocument/superTypes');
+}
+
+
+/**
+ * Represents hierarchical information about programming constructs like variables, classes,
+ * interfaces etc.
+ */
+export interface SymbolNode extends SymbolInformation {
+  /**
+   * Immediate descendants of this node
+   */
+  descendants? : SymbolNode[]
+  /**
+   * true if this node has children. hasChildren can be
+   * true even when the descendants are empty. In that case
+   * the clients should do a call to retrieve the descendants.
+   */
+  hasChildren: boolean
+}


protocol/protocol.typeHierarchy.proposed.md
```diff
@@ -0,0 +1,77 @@
...
+```ts
+/**
+ * Represents hierarchical information about programming constructs like variables, classes,
+ * interfaces etc.
+ */
+export interface SymbolNode extends SymbolInformation {
+  /**
+   * Immediate descendants of this node
+   */
+  descendants? : SymbolNode[]
```
felixfbecker: If it's the immediate descendants, it should be called `children`

protocol/protocol.typeHierarchy.proposed.md
```diff
@@ -0,0 +1,77 @@
+
+#### Type Hierarchy
```
felixfbecker: Is there a reason this proposal is using the term "type" everywhere and not "symbol"? For example in TypeScript, those are two very different concepts

protocol/protocol.typeHierarchy.proposed.md
```diff
@@ -0,0 +1,77 @@
...
+The `textDocument/superTypes` request is sent from the client to the server to resolve resolve the supertypes of a symbol at a given text document position. Returns all the supertypes in bottom-up order.
+
+_Request_:
+
+method: ‘textDocument/superTypes’
+params: TextDocumentPositionParams
+
+_Response_
+
+* result: `SymbolNode[] | null`
```
felixfbecker: So a type can have multiple parent types? If that is true, how would it be displayed in a tree UI (vs a graph)?

felixfbecker:
I don't think this is fixes https://github.com/Microsoft/language-server-protocol/issues/136 - that issue is about a symbol hierachy tree in a workspace, but this seems to be about getting the inheritance hierachy for definition at a given location.

protocol/protocol.typeHierarchy.proposed.md
gorkem: The doc is incorrect needs to be updated. 

I am not sure about children either. This can actually hold hierarchies with depth larger than one. Probably needs a better name.

protocol/protocol.typeHierarchy.proposed.md
gorkem: I think those are very different concepts on most languages.

protocol/protocol.typeHierarchy.proposed.md
gorkem: multiple-inheritance has been around and there have been tree based representations of it implemented before. The most generic approach I have seen is to replicate the same Type hierarchy tree under all parent nodes. We can add additional information to this data structure if it is going to help with its representation.

protocol/protocol.typeHierarchy.proposed.md
felixfbecker: Even if it's depth larger than one, wouldn't the elements of this particular array still be the immediate children?

protocol/protocol.typeHierarchy.proposed.md
felixfbecker: I think you should reclarify the goal of this proposal, I misunderstood it. The linked issue describes a symbol hierarchy (e.g. namespaces -> classes -> members), not inheritance

gorkem:
Microsoft/language-server-protocol#136  refers to inheritance on object-oriented languages, I know because I have written it. I think you are thinking about a way to solve the problem created by the  very insufficient `SymbolInfomation#containerName` field, which is not this one.

felixfbecker:
Okay, I must have remembered it wrong. Sorry for the confusion. I always had a workspace symbol request in my mind.

protocol/protocol.typeHierarchy.proposed.md
gorkem: We use this same data structure for both `subTypes` and `superTypes` calls on the superTypes call they are actually the ancestors/parents. We probably need a new name to reflect that.

protocol/protocol.typeHierarchy.proposed.md
gorkem: Also take a look at https://github.com/Microsoft/language-server-protocol/issues/472 I think that should include the symbol hirearchy

dbaeumer:
Some first comments:

- shouldn't the API be more level based. What do you expect such a request returns if ask on a huge Java system for the type hierarchy of Object.
- this is somehow foreseen with the hasChildren but from the proposal it is unclear to me how a client would `resolve` the next level of the children.

Some ideas:

- we have `textDocument/subTypes` & `textDocument/superTypes` requests
- we have `symbolNode/resolve` requests for the next level.


We should also see how this aligns with the document symbol hierarchy request to generate the outline of a document.

gorkem:
The idea was to make another call to `textDocument/subTypes` with the information from the `Location` information from `SymbolNode`.  There isn't any generic enough information that can be passed to `symbolNode/resolve` that is helpful and different from`textDocument/subTypes`. 

Another idea is to pass `SymbolInformation` to `textDocument/subTypes` and `textDocument/superTypes` calls instead of current `TextDocumentPositionParams`.  
A new param type like below.  We can also amend `SymbolInformation` with an optional `data` field similar to `CompletionItem` that servers would retrieve back to skip the symbol search. The downside is we need to always retrieve the `SymbolInformation` prior to hierarchy calls.

```ts
export interface HierarchyParams {
  /**
   * The symbol for the call.
   */
	symbol: SymbolInformation
}
```

dbaeumer:
We should use the same data property mechanism that we use for document links and completion items to allow the server to identify the object in the resolve call.

I am wondering whether it makes sense to have a Tree<T> and Graph<T> instead of a SymbolNode in the protocol. Then the request would return a Graph<SymbolInformation>. We could further classify the edges in the tree / graph. This would then work for hierarchical document symbols, type hierarchies, namespace hierarchies or even reference search.

tsmaeder:
@dbaeumer I don't quite get what you mean with your last comment. Both about the "data property mechanism" and what you would like to do with trees and graphs.

protocol/protocol.typeHierarchy.proposed.md
```diff
@@ -0,0 +1,77 @@
...
+
+_Request_:
+
+method: ‘textDocument/subTypes’
+params: TextDocumentPositionParams
+
+_Response_
+
+* result: `SymbolNode[] | null`
+* error: code and message set in case an exception happens during the 'textDocument/subTypes' request
```
svenefftinge: When I open a type hierarchy on a certain position in the text, I would like to see the `SymbolNode` for that token together with its subtypes. How do I obtain the SymbolNode I am opening the hierarchy on?

tsmaeder:
@dbaeumer So the GraphNode thing is more about not linking it to the SymbolInformation than any functional difference? Making it more general, but structurally it would be similar?
Got it about the "userData".

dbaeumer:
@tsmaeder yes that is the idea.

protocol/protocol.typeHierarchy.proposed.md
MaskRay: It may sometimes be useful to expand multiple levels in one request, This is how [ccls](https://github.com/MaskRay/ccls/blob/master/src/messages/ccls_call_hierarchy.cc#L44) and cquery do.

protocol/protocol.typeHierarchy.proposed.md
MaskRay: There can be multiple dimensions of sub-super relations: template parameters related to a template, derived types related to a base, specializations. Is there some way to differentiate them in this proposal?

gorkem:
FYI since this has stalled a bit because I got new responsibilities @tsmaeder will be taking over this PR, so stay tuned

dbaeumer:
@tsmaeder in VS Code we experimented with a GraphNode and we gave up on it :-(. Mainly because it was hard to construct a graph without adding a builder or some other kind of help methods. If you have a nice idea I am still open to it. Otherwise we found that it is better to have specific return types for every request. So in the type hierarchy case something like a `TypeHierarchyNode`.

simark:
A use case I would like to see supported is calling the type hierarchy request on a method (declaration, definition or call) and seeing which nodes implement that method.  In Eclipse JDT and CDT, for example (probably others too), you can get something like this:

screenshot_2018-08-27_10-47-37

This gives the hierarchy from the point of view of the Child2 class, but also tells you that the method I called "type hierarchy" has an definition in the Parent class, but not in the Child class.

Protocol-wise, I think that calling "type hierarchy" on a method could be equivalent to doing the request on the type of the object in which the method is declared/defined or on which the method is called.  But the hierarchy nodes could also include a flag (only useful if the target of the request is a method) indicating if the node implements that method.  In addition, we would need to get information about what the user has called "type hierarchy" on (name, type, etc).

protocol/protocol.typeHierarchy.proposed.md
```diff
@@ -0,0 +1,77 @@
...
+  typeHierarchyProvider?: boolean
+}
+```
+
+```ts
+/**
+ * Represents hierarchical information about programming constructs like variables, classes,
+ * interfaces etc.
+ */
+export interface SymbolNode extends SymbolInformation {
```
kittaakos: As of LSP [`3.10.0`](https://microsoft.github.io/language-server-protocol/specification#version_3_10_0), we can use the `DocumentSymbol` as is, right? It has the `children` property.

kittaakos: We need one structural change, and we could simplify this PR:

The `DocumentSymbol` should be 
```js
export class DocumentSymbol {
	name: string;
	detail?: string;
	kind: SymbolKind;
	deprecated?: boolean;
	range: Range | Location;
	selectionRange: Range | Location;
	children?: DocumentSymbol[];
}
```

instead of the current state:
```js
export class DocumentSymbol {
	name: string;
	detail?: string;
	kind: SymbolKind;
	deprecated?: boolean;
	range: Range;
	selectionRange: Range;
	children?: DocumentSymbol[];
}
```

When the `range` and the `selectionRange` has the type `Range` then the `DocumentUri` is inferred. For example, when calling `textDocument/documentSymbol`, the `DocumentUri` can be inferred as `DocumentSymbolParams.textDocument.uri` from the param. If the `range` and the `selectionRange` are `Location` types, then the `DocumentUri` is explicitly defined (`Location.uri`).

If we accept this structural change in the `DocumentSymbol`, we can defined our two new methods as `textDocument/subTypes ` and `textDocument/superTypes` with `DocumentSymbol[]` return type.

We can even stick to the `children` name for the `textDocument/superTypes`; the supertypes will be real children of the `DocumentSymbol` node, only the way of the traversal differs.

CC: @tsmaeder (since you have [taken over the PR](https://github.com/Microsoft/vscode-languageserver-node/pull/346#issuecomment-400379761))

svenefftinge: +1 See related https://github.com/Microsoft/vscode-languageserver-node/pull/420#discussion_r223275639. 
Btw. the documentation for `DocumentSymbol#selectionRange` says 
```
The range that should be selected and revealed when this symbol is being picked, e.g the name of a function.
	 * Must be contained by the `range`.
```
So with this contract it is enough if the `range` can alternatively be a  `Location`.

svenefftinge: Actually I would vote for adding an option `uri: URI` property to DocumentSymbol, because the property name `range` highly suggests that the datatype is a `Range`, too. Also it will not break existing clients.

kittaakos:
I am done with a reference [server implementation for Java](https://github.com/eclipse/eclipse.jdt.ls/pull/823), also with the [client for Eclipse Theia](https://github.com/theia-ide/theia/pull/3148). (The Java implementation of the extended LSP is [here](https://github.com/eclipse/lsp4j/pull/273)).

Type Hierarchy in action:

screencast 2018-10-11 15-18-32

----

Besides declaring the `textDocument/subTypes` and `textDocument/superTypes` methods, I had to add an [optional URI](https://github.com/Microsoft/language-server-protocol/issues/582) (`uri?: DocumentUri`) property to the `DocumentSymbol`. No other changes were invovled.

----

Since I am not allowed to make any changes in this PR, how should we proceed to speed things up a bit? Is it OK if I create a separate PR with the proposed changes? (CC: @dbaeumer, @tsmaeder)

dbaeumer:
Sorry for the long delay on this. From the experiences I gained with the LSP so far I would opt to not overload any existing types. We are better off introducing new once even if they look similar or at least introduce a type alias. Otherwise it will be hard to evolve them separately. In this regard I would opt for the following:

- have a type `TypeHierarchyItem`
- it should be possible to lazy resolve a type hierarchy.

Given the above to I would end up with something like this:

```TS
interface TypeHierarchyItem {
	name: string;
	detail?: string;
	kind: SymbolKind;
	deprecated?: boolean;
	uri: string;
	range: Range;
	selectionRange: Range;

	parents?: TypeHierarchyItem[];
	children?: TypeHierarchyItem[];

	data?: any
}
```

A first request would return a TypeHierarchyItem for a given cursor location. The request would also allow to specify if the item should be resolved and whether sub or super types are to be resolved. Something like this:

```TS
interface TypeHierarchyParams extends TextDocumentPositionParams {
    resolve?: number; // the levels to resolve. 0 indicates no level.
    direction?: 'parents' | 'children' | 'both';
}
```

We also need another request to resolve an unresolved `TypeHierarchyItem` (which is indicated if parents or children is undefined. If resolved and no parents or children are available then an empty array is returned). The request would look the same as resolving a compeletion item with the same `resolve` and `direction` params as above. So something like this

interface ResolveTypeHierarchyItemParams {
    item: TypeHierarchyItem;
    resolve: number; // the levels to resolve. 0 indicates no level.
    direction: 'parents' | 'children' | 'both';
}

The optional data field can be used to identify a hierarchy item in a resolve request.

Does this make sense?

@jrieken any comments / objections on this design from a VS Code API perspective.

dbaeumer:
Closing in favour of https://github.com/Microsoft/vscode-languageserver-node/pull/426

PR #347, opened 2018-05-08
CompletionItem's deprecated property should be preserved across requests

rcjsuen:
Although VS Code does not currently support appending a visual indicator that a completion item is "deprecated", the `deprecated` property itself of a `CompletionItem` should still be preserved across requests. This will allow language servers to read and write to that field across a `textDocument/completion` request and a `completionItem/resolve` request.

I used `if (x.val === true || x.val === false) { y.val = x.val; }` to ensure that only `boolean` values would be converted as I didn't want to encourage an "anything goes" attitude.


dbaeumer:
As always nice work.

rcjsuen:
Thanks for merging it in so quickly, Dirk!

PR #350, opened 2018-05-18
Add support for code action literal as a return value of the textDocument/codeAction request

dbaeumer:


types/src/main.ts
```diff
@@ -1350,6 +1436,14 @@ export interface CodeActionContext {
...
 	 */
 	diagnostics: Diagnostic[];
+
+	/**
+	 * Requested kind of actions to return.
+	 *
+	 * Actions not of this kind are filtered out by the client before being shown. So servers
+	 * can omit computing them.
+	 */
+	only?: CodeActionKind
```
rcjsuen: Why isn't this a `CodeActionKind[]` so that the client can request two separate kinds of code actions? What if I don't want `quickfix` but I want both `refactor` and `source`?

dbaeumer:
Good point. Agree to change this.

client/src/client.ts
```diff
@@ -1671,7 +1671,22 @@ class CodeActionFeature extends TextDocumentFeature<TextDocumentRegistrationOpti
...
+			codeActionKind: {
+				valueSet: [
+					VCodeActionKind.Empty.value!,
+					VCodeActionKind.QuickFix.value!,
+					VCodeActionKind.Refactor.value!,
+					VCodeActionKind.RefactorExtract.value!,
+					VCodeActionKind.RefactorInline.value!,
+					VCodeActionKind.RefactorRewrite.value!,
+					VCodeActionKind.Source.value!,
+					VCodeActionKind.SourceOrganizeImports.value!
```
rcjsuen: Wouldn't it make more sense to use `CodeActionKind` from the language server protocol instead of using the VS Code one? That sounds a lot safer and is more agnostic since then we're only using stuff that the protocol explicitly declares.

types/src/main.ts
```diff
@@ -1341,6 +1350,83 @@ export interface WorkspaceSymbolParams {
...
+ *
+ * The set of kinds is open and client needs to announce the kinds it supports to the server during
+ * initialization.
+ */
+export type CodeActionKind = string;
+
+/**
+ * A set of predefined code action kinds
+ */
+export namespace CodeActionKind {
```
rcjsuen: There is no `Empty` type but VS Code does have one. Is this intentional?

client/src/client.ts
```diff
@@ -1671,7 +1671,22 @@ class CodeActionFeature extends TextDocumentFeature<TextDocumentRegistrationOpti
 	}
 
 	public fillClientCapabilities(capabilites: ClientCapabilities): void {
-		ensure(ensure(capabilites, 'textDocument')!, 'codeAction')!.dynamicRegistration = true;
+		const cap = ensure(ensure(capabilites, 'textDocument')!, 'codeAction')!;
+		cap.dynamicRegistration = true;
+		cap.codeActionLiteralSupport = {
+			codeActionKind: {
+				valueSet: [
+					VCodeActionKind.Empty.value!,
```
rcjsuen: The protocol doesn't have a `CodeActionKind.Empty` declared. Is this an issue? See below.

types/src/main.ts
dbaeumer: I though that this is not necessary in the LSP since we can simply use the empty string. In VSCode it is a class with a private constructor hence you can only make new kinds by starting with Empty and then append. Since we have literals I felt this is overkill and left it out.

client/src/client.ts
dbaeumer: See above.

client/src/client.ts
dbaeumer: Agree. Will change it.

Juni66623 (approved):

PR #357, opened 2018-06-06
Typeguards

felixfbecker:
- Adds tests for some typeguards.
- While writing tests, I noticed bugs - `Position.is()` and `Range.is()` would return `true` for `null` (but not for `undefined`). That made `Position.is` and `Range.is` ineffective as validating for a valid Position or Range, as there are no guarantees that the properties of a Position/Range can actually be accessed (and the LSP spec only allows `null` in specific places). As an example of why this is surprising, `Array.isArray()` returns `false` for `null`.
- Add typeguard and tests for `Hover` and contained types. I found myself needing these in multiple projects and given most interfaces have typeguards I thought it would be good to add them to the official package.
- package-lock.json changes came automatically from running `npm install`, seems like they weren't up to date with their package.json.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "vscode-languageclient",
-	"version": "4.2.0",
+	"version": "4.2.1",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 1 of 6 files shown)


dbaeumer:
Thanks. Valuable PR. Merged in the changes by hand. Mainly because:

- no semicolon at the end of a statement
- no reuse of what is available in `Is`
- add `Is.objectLiteral` to make the code consistent again

felixfbecker:
> no semicolon at the end of a statement

It would be nice if this repo had tslint and/or prettier to avoid this. Would you be open to add this?

dbaeumer:
Yes, I thought the same and I am open for a PR that adds this.

felixfbecker:
Would you mind doing a release that includes this PR?

dbaeumer:
Published Types 3.8.2

PR #360, opened 2018-06-08
Add prettier

felixfbecker:
The .prettierrc is just a suggestion, can be easily changed and `npm run prettier` will reformat.
I tried following the existing code style (semicolons etc).
The diff is quite big because line endings were inconsistent I think.


dbaeumer:
@felixfbecker thanks for the PR. I think we misunderstood each other. I would accept a PR to add tslint to check for the semicolon and formatter settings for the built in JS/TS formatter. I think there is no need to add prettier to it. And most important I would like to stay with tabs for indentation :-)

octref:
@dbaeumer Do you mind removing `tabSize` here though? https://github.com/Microsoft/vscode-languageserver-node/blob/master/.vscode/settings.json

Users should be allowed to configure their `tabSize` since it's purely presentational preferences.

dbaeumer:
@octref good point. Removed it.

dbaeumer:
I will close the PR. The repository now uses ESLint for linting and auto correction on save.

PR #366, opened 2018-06-22
Introduce Traceformat to allow JSON format log

octref:
Things I plan to do but haven't done yet:

- [ ] When `verbosity` is set to `message`, ignore `params` on Request / Notification messages
- [ ] Expose `traceFormat` to client's initialization params

Questions:

- [ ] Can we rename this to `[langId].lsp.trace`? `server` is misleading as the logs are actually logged from the client perspective. I'll write documentation for this too so I hope we can get this right.

@dbaeumer Would like to get some high-level insight - let me know if this is the right direction, what more things you would want to me to add, etc.

More details are at https://github.com/Microsoft/language-server-protocol-inspector/issues/8

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -22,7 +22,7 @@ import {
 	RequestType, RequestType0, RequestHandler, RequestHandler0, GenericRequestHandler,
 	NotificationType, NotificationType0,
 	NotificationHandler, NotificationHandler0, GenericNotificationHandler,
-	MessageReader, MessageWriter, Trace, Tracer, Event, Emitter,
+	MessageReader, MessageWriter, Trace, TraceFormat, Tracer, Event, Emitter,
 	createProtocolConnection,
 	ClientCapabilities, WorkspaceEdit,
 	RegistrationRequest, RegistrationParams, UnregistrationRequest, UnregistrationParams, TextDocumentRegistrationOptions,
@@ -100,7 +100,7 @@ interface IConnection {
 	onNotification(method: string, handler: GenericNotificationHandler): void;
 	onNotification(method: string | RPCMessageType, handler: GenericNotificationHandler): void;
 
-	trace(value: Trace, tracer: Tracer, sendNotification?: boolean): void;
+	trace(value: Trace, format: TraceFormat, tracer: Tracer, sendNotification?: boolean): void;
 
 	initialize(params: InitializeParams): Thenable<InitializeResult>;
 	shutdown(): Thenable<void>;
@@ -161,7 +161,7 @@ function createConnection(input: any, output: any, errorHandler: ConnectionError
 		sendNotification: (type: string | RPCMessageType, params?: any): void => connection.sendNotification(Is.string(type) ? type : type.method, params),
 		onNotification: (type: string | RPCMessageType, handler: GenericNotificationHandler): void => connection.onNotification(Is.string(type) ? type : type.method, handler),
 
-		trace: (value: Trace, tracer: Tracer, sendNotification: boolean = false): void => connection.trace(value, tracer, sendNotification),
+		trace: (value: Trace, format: TraceFormat, tracer: Tracer, sendNotification: boolean = false): void => connection.trace(value, format, tracer, sendNotification),
 
 		initialize: (params: InitializeParams) => connection.sendRequest(InitializeRequest.type, params),
 		shutdown: () => connection.sendRequest(ShutdownRequest.type, undefined),
@@ -2249,6 +2249,7 @@ export abstract class BaseLanguageClient {
 	private _stateChangeEmitter: Emitter<StateChangeEvent>;
 
 	private _trace: Trace;
+	private _traceFormat: TraceFormat;
 	private _tracer: Tracer;
 
 	private _c2p: c2p.Converter;
@@ -2302,6 +2303,9 @@ export abstract class BaseLanguageClient {
 		this._tracer = {
 			log: (message: string, data?: string) => {
 				this.logTrace(message, data);
+			},
+			logLSP: (message: string) => {
+				this.logLSPTrace(message);
 			}
 		};
 		this._c2p = c2p.createConverter(clientOptions.uriConverters ? clientOptions.uriConverters.code2Protocol : undefined);
@@ -2438,7 +2442,17 @@ export abstract class BaseLanguageClient {
 		this._trace = value;
 		this.onReady().then(() => {
 			this.resolveConnection().then((connection) => {
-				connection.trace(value, this._tracer);
+				connection.trace(value, this._traceFormat, this._tracer);
+			})
+		}, () => {
+		});
+	}
+
+	public set traceFormat(value: TraceFormat) {
+		this._traceFormat = value;
+		this.onReady().then(() => {
+			this.resolveConnection().then((connection) => {
+				connection.trace(this._trace, value, this._tracer);
 			})
 		}, () => {
 		});
@@ -2496,6 +2510,10 @@ export abstract class BaseLanguageClient {
 		if (data) {
 			this.outputChannel.appendLine(this.data2String(data));
 		}
+	}
+
+	private logLSPTrace(message: string): void {
+		this.outputChannel.appendLine(`[LSP   - ${(new Date().toLocaleTimeString())}] ${message}`);
 	}
 
 	public needsStart(): boolean {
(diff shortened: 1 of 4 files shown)


dbaeumer:
@octref goes absolutely into the right direction. Only one little comment so far.

octref:
Hey Dirk, there are a bunch of updates that happened, so I'm wondering if you can take another look at this PR. Also /cc @aeschli in case you are interested.

### High Level Changes

This PR remain as it is, so:
  - `trace.server` continues to work
  - `trace.server` can be an object now, for example:
    ```
    "mls.trace.server": {
      "verbosity": "verbose", // values same as old `trace.server`
      "format": "json" // or `text` for old behavior
    }
    ```

For [streaming support](https://github.com/Microsoft/language-server-protocol-inspector/issues/5), it can be implemented trivially by Language Server authors using the existing API `LanguageClientOptions.outputChannel`. For example: https://github.com/octref/vscode-language-server-template/commit/4181af806233f98f2dd654968e4171ee1dabcc33.

This goes **against** what we initially agreed on, e.g:

```
{
  "format": "json" | "text",
  "verbosity": "off" | "messages" | "verbose",
  "output": "channel" | "log" | "websocket"
}
```

the benefit of leaving out logging / websocket-streaming is that we don't add dependencies (`ws` for websocket) and we give ext authors more freedom on handling logs.

### How Things Work Now



There are 3 repos:

- [octref/vscode-lsp-inspector](https://github.com/octref/vscode-lsp-inspector): The Inspector in webview with streaming support. It:
  - launches the web app in webview
  - opens a Websocket server to listen to incoming connections that would send LSP log data
  - sends LSP log data into the web app and update UI.
- [Microsoft/language-server-protocol-inspector](https://github.com/Microsoft/language-server-protocol-inspector): I'll merge [octref/vscode-lsp-inspector](https://github.com/octref/vscode-lsp-inspector) into this. So this repo will have 2 build targets. One is on LSP Website. The other one is the webview extension.
- [Microsoft/vscode-languageserver-node](https://github.com/Microsoft/vscode-languageserver-node): People using the new version of the lib can specify output format of the LSP log. They can then save the log, stream it into the Inspector or do whatever they want.

### Play With It

- Download the VSIX: https://github.com/octref/vscode-lsp-inspector/releases/tag/v0.0.1
- Build and Run the `stream` branch of: https://github.com/octref/vscode-language-server-template/tree/stream
- Open a MD file. The LS should connect to the LSP Inspector Webview extension's WebSocket and start streaming data into it.

jsonrpc/src/messages.ts
```diff
@@ -121,6 +121,23 @@ export interface ResponseMessage extends Message {
...
+  | 'send-response'
+  | 'receive-response'
+  | 'send-notification'
+  | 'receive-notification';
+
+export interface LSPLogMessage {
+    type: LSPMessageType;
+    message: RequestMessage | ResponseMessage | NotificationMessage;
+    timestamp: number;
+}
```
dbaeumer: LSP uses tabs not spaces

jsonrpc/src/main.ts
```diff
@@ -205,6 +206,36 @@ export namespace Trace {
 	}
 }
 
+export enum TraceFormat {
```
dbaeumer: Instead of having a number enum and a TraceFormatValues would it make sense to use a string Enum which is now supported by TS?

jsonrpc/src/main.ts
```diff
@@ -224,6 +255,7 @@ export namespace LogTraceNotification {
 
 export interface Tracer {
 	log(message: string, data?: string): void;
+	logLSP(message: string): void;
```
dbaeumer: This is a breaking API change. We should either make it optional or reuse the `log` signature. I looked at the client code and I was under the impression that the only difference is that the `log` prints [Trace - ...] and `logLSP` prints [LSP - ...]. Since the client knows the log format could this be decided on the client side.

protocol/src/main.ts
```diff
@@ -157,7 +157,7 @@ export interface ProtocolConnection {
 	/**
 	 * Enables tracing mode for the connection.
 	 */
-	trace(value: Trace, tracer: Tracer, sendNotification?: boolean): void;
+	trace(value: Trace, traceFormat: TraceFormat, tracer: Tracer, sendNotification?: boolean): void;
```
dbaeumer: This is breaking as well. We should support this in a non breaking way by having two signatures. If the old one is used we do the old text tracing.

client/src/client.ts
```diff
@@ -2438,7 +2442,17 @@ export abstract class BaseLanguageClient {
...
 		this.onReady().then(() => {
 			this.resolveConnection().then((connection) => {
-				connection.trace(value, this._tracer);
+				connection.trace(value, this._traceFormat, this._tracer);
+			})
+		}, () => {
+		});
+	}
+
+	public set traceFormat(value: TraceFormat) {
```
dbaeumer: Would it make sense to fold that into trace as an optional param. If not provided it is text. I doubt that users set this independently.

protocol/src/main.ts
dbaeumer: Same as on the implementation side. This is breaking...

dbaeumer:
@octref provided by comments.

jsonrpc/src/main.ts
dbaeumer: Actually we should have a log method that takes an JSON object literal. So the JSON RPC code shouldn't even convert it to a strings. Then someone could plug in a tracer that writes the JSON objects to a file.

jsonrpc/src/main.ts
octref: Was aligning to `TraceValues`, string enum makes sense.

jsonrpc/src/main.ts
octref: > Actually we should have a log method that takes an JSON object literal.

Wouldn't that be breaking as well?

client/src/client.ts
```diff
@@ -2498,6 +2505,17 @@ export abstract class BaseLanguageClient {
 		}
 	}
 
+	private logObjectTrace(data: any): void {
+		if (data.isLSPMessage && data.type) {
```
octref: @dbaeumer Not sure if this is what you meant by "if the data is an object then log it by JSON stringify here". Also not sure if that `isLSPMessage` is the best way to go. If you have better suggestions, let me know!

client/src/client.ts
octref: Sure, since `trace` function already covers the format as an optional param. However I'm wondering if you mind adding `traceFormat` as one of the `LanguageClientOptions`? I feel this is something people want to configure on LS Client init but not setting it each time.

octref:
Addressed comments, but left some questions too.

client/src/client.ts
```diff
@@ -161,7 +161,7 @@ function createConnection(input: any, output: any, errorHandler: ConnectionError
 		sendNotification: (type: string | RPCMessageType, params?: any): void => connection.sendNotification(Is.string(type) ? type : type.method, params),
 		onNotification: (type: string | RPCMessageType, handler: GenericNotificationHandler): void => connection.onNotification(Is.string(type) ? type : type.method, handler),
 
-		trace: (value: Trace, tracer: Tracer, sendNotification: boolean = false): void => connection.trace(value, tracer, sendNotification),
+		trace: (value: Trace, tracer: Tracer, sendNotification: boolean = false, traceFormat: TraceFormat = TraceFormat.Text): void => connection.trace(value, tracer, sendNotification, traceFormat),
```
dbaeumer: I know it is easier to implement if we simply add an optional arg at the endd but I would prefer to have to signatures like this:

```
trace: (value: Trace, tracer: Tracer, sendNotification: boolean = false)
trace: (value: Trace, traceFormat: TraceFormat, tracer: Tracer, sendNotification: boolean = false)
```

to keep the args together and may be deprecate the old signature.

client/src/client.ts
```diff
@@ -2300,9 +2301,15 @@ export abstract class BaseLanguageClient {
 		this._telemetryEmitter = new Emitter<any>();
 		this._stateChangeEmitter = new Emitter<StateChangeEvent>();
 		this._tracer = {
-			log: (message: string, data?: string) => {
-				this.logTrace(message, data);
-			}
+			log: (...params: any[]) => {
```
dbaeumer: Can this not still be message & data and message can be string | any. ...params makes it harder to understand what data can actually come.

client/src/client.ts
dbaeumer: No, in this case I would simply do 

```
this.outputChannel.append(`[LSP   - ${(new Date().toLocaleTimeString())}] `);
this.outputChannel.appendLine(`${JSON.stringify(data)}`);
```

client/src/client.ts
octref: @dbaeumer Decided to have this:

```ts
trace(value: Trace, tracer: Tracer, sendNotification?: boolean): void;
trace(value: Trace, tracer: Tracer, traceOptions?: TraceOptions): void;

// In json-rpc

export interface TraceOptions {
	sendNotification?: boolean;
	traceFormat?: TraceFormat;
}
```

I initially had what you suggested, but later abandoned it in favor of `traceFormat?` at the end, as
- I'd prefer to keep `traceFormat` optional
- A call with three args can be of either signature.

With this new one it's easier to detect the params of the overloaded functions, and in the future it's easier to add more options without breaking backward compatibility.

dbaeumer: No, this is not what I meant. I was suggesting to have this:

```ts
trace: (value: Trace, tracer: Tracer, sendNotification: boolean = false)
trace: (value: Trace, traceFormat: TraceFormat, tracer: Tracer, sendNotification: boolean = false)
```

with one implementation overload like this:

```ts
trace(arg0: Trace, arg1: Tracer | TraceFormat, arg2?: Tracer | boolean, arg3?: boolean)
```

There is enough information in the signatures to decide which on is meant.

PR #367, opened 2018-06-27
Proposal of the semantic highlighting protocol extension

kittaakos:
Task: #368.

### DIFF
--- a/client/src/semanticHighlighting.proposed.ts
+++ b/client/src/semanticHighlighting.proposed.ts
@@ -1,0 +1,109 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) TypeFox. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import { Disposable, Uri, Range, window, DecorationRenderOptions, TextEditorDecorationType, workspace, TextEditor } from 'vscode';
+import {
+	TextDocumentRegistrationOptions, ClientCapabilities, ServerCapabilities, DocumentSelector, NotificationHandler,
+	SemanticHighlightingNotification, SemanticHighlightingParams, SemanticHighlightingInformation
+} from 'vscode-languageserver-protocol';
+
+import * as UUID from './utils/uuid';
+import { TextDocumentFeature, BaseLanguageClient } from './client';
+
+export class SemanticHighlightingFeature extends TextDocumentFeature<TextDocumentRegistrationOptions> {
+
+	protected readonly toDispose: Disposable[];
+	protected readonly decorations: Map<string, any>;
+	protected readonly handlers: NotificationHandler<SemanticHighlightingParams>[];
+
+	constructor(client: BaseLanguageClient) {
+		super(client, SemanticHighlightingNotification.type);
+		this.toDispose = [];
+		this.decorations = new Map();
+		this.handlers = [];
+		this.toDispose.push({ dispose: () => this.decorations.clear() });
+		this.toDispose.push(workspace.onDidCloseTextDocument(e => {
+			const uri = e.uri.toString();
+			if (this.decorations.has(uri)) {
+				// TODO: do the proper disposal of the decorations.
+				this.decorations.delete(uri);
+			}
+		}));
+	}
+
+	dispose(): void {
+		this.toDispose.forEach(disposable => disposable.dispose());
+		super.dispose();
+	}
+
+	fillClientCapabilities(capabilities: ClientCapabilities): void {
+		if (!!capabilities.textDocument) {
+			capabilities.textDocument = {};
+		}
+		capabilities.textDocument!.semanticHighlightingCapabilities = {
+			semanticHighlighting: true
+		};
+	}
+
+
+	initialize(capabilities: ServerCapabilities, documentSelector: DocumentSelector): void {
+		if (!documentSelector) {
+			return;
+		}
+		const capabilitiesExt: ServerCapabilities & { semanticHighlighting?: { scopes: string[][] | undefined } } = capabilities;
+		if (capabilitiesExt.semanticHighlighting) {
+			const { scopes } = capabilitiesExt.semanticHighlighting;
+			if (scopes && scopes.length > 0) {
+				// this.toDispose.push(this.semanticHighlightingService.register(this.languageId, scopes));
+				const id = UUID.generateUuid();
+				this.register(this.messages, {
+					id,
+					registerOptions: Object.assign({}, { documentSelector: documentSelector }, capabilitiesExt.semanticHighlighting)
+				});
+			}
+		}
+	}
+
+	protected registerLanguageProvider(options: TextDocumentRegistrationOptions): Disposable {
+		if (options.documentSelector === null) {
+			return new Disposable(() => { });
+		}
+		const handler = this.newNotificationHandler.bind(this)();
+		this._client.onNotification(SemanticHighlightingNotification.type, handler);
+		return new Disposable(() => {
+			const indexOf = this.handlers.indexOf(handler);
+			if (indexOf !== -1) {
+				this.handlers.splice(indexOf, 1);
+			}
+		})
+	}
+
+	protected newNotificationHandler(): NotificationHandler<SemanticHighlightingParams> {
+		return (params: SemanticHighlightingParams) => {
+			const editorPredicate = this.editorPredicate(params.textDocument.uri);
+			window.visibleTextEditors.filter(editorPredicate).forEach(editor => this.applyDecorations(editor, params));
+		};
+	}
+
+	protected editorPredicate(uri: string): (editor: TextEditor) => boolean {
+		const predicateUri = Uri.parse(uri);
+		return (editor: TextEditor) => editor.document.uri.toString() === predicateUri.toString();
+	}
+
+	protected applyDecorations(editor: TextEditor, params: SemanticHighlightingParams): void {
+		console.log('TODO: Apply the decorations on the editor.', editor, params);
+	}
+
+	protected decorationType(options: DecorationRenderOptions = {}) {
+		return window.createTextEditorDecorationType(options);
+	}
+
+	protected map2Decoration(lines: SemanticHighlightingInformation[]): [TextEditorDecorationType, Range[]] {
+		console.log('TODO: Map the lines (and the tokens) to the desired decoration type.', lines);
+		return [this.decorationType(), []];
+	}
+
+}
--- a/protocol/src/protocol.semanticHighlighting.proposed.md
+++ b/protocol/src/protocol.semanticHighlighting.proposed.md
@@ -1,0 +1,134 @@
+#### Semantic Highlighting
+
+While the syntax highlighting is done on the client-side and can handle keywords, strings, and other low-level tokens from the grammar, it cannot adequately support complex coloring. Semantic highlighting information is calculated on the language server and pushed to the client as a notification. This notification carries information about the ranges that have to be colored. The desired coloring details are given as [TextMate scopes](https://manual.macromates.com/en/language_grammars) for each affected range. For the semantic highlighting information the following additions are proposed:
+
+_Client Capabilities_:
+
+Capability that has to be set by the language client if it can accept and process the semantic highlighting information received from the server.
+
+```ts
+/**
+ * The text document client capabilities.
+ */
+textDocument?: {
+
+	/**
+	 * The client's semantic highlighting capability.
+	 */
+	semanticHighlightingCapabilities?: {
+
+		/**
+		 * `true` if the client supports semantic highlighting support text documents. Otherwise, `false`. It is `false` by default.
+		 */
+		semanticHighlighting: boolean;
+
+	}
+
+}
+```
+
+_Server Capabilities_:
+
+If the client declares its capabilities with respect to the semantic highlighting feature, and if the server supports this feature too, the server should set all the available TextMate scopes as a "lookup table" during the `initialize` request.
+
+```ts
+/**
+ * Semantic highlighting server capabilities.
+ */
+semanticHighlighting?: {
+
+	/**
+	 * A "lookup table" of semantic highlighting [TextMate scopes](https://manual.macromates.com/en/language_grammars)
+	 * supported by the language server. If not defined or empty, then the server does not support the semantic highlighting
+	 * feature. Otherwise, clients should reuse this "lookup table" when receiving semantic highlighting notifications from
+	 * the server.
+	 */
+	scopes?: string[][];
+}
+```
+
+##### SemanticHighlighting Notification
+
+The `textDocument/semanticHighlighting` notification is pushed from the server to the client to inform the client about additional semantic highlighting information that has to be applied on the text document. It is the server's responsibility to decide which lines are included in the highlighting information. In other words, the server is capable of sending only a delta information. For instance, after opening the text document (`DidOpenTextDocumentNotification`) the server sends the semantic highlighting information for the entire document, but if the server receives a `DidChangeTextDocumentNotification`, it pushes the information only about the affected lines in the document.
+
+The server never sends delta notifications, if no new semantic highlighting ranges were introduced but the existing onces have been shifted. For instance, when inserting a new line to the very beginning of the text document. The server receives the `DidOpenTextDocumentNotification`, updates its internal state, so that the client and the server shares the same understanding about the highlighted positions but the server does not send any notifications to the client. In such cases, it is the client's responsibility to track this event and shift all existing markers.
+
+The server can send a `SemanticHighlightingInformation` to the client without defining the `tokens` string. This means, the client must discard all semantic highlighting information in the line. For instance when commenting out a line.
+
+_Notification_:
+
+* method: 'workspace/semanticHighlighting'
+* params: `SemanticHighlightingParams` defined as follows:
+
+```ts
+/**
+ * Parameters for the semantic highlighting (server-side) push notification.
+ */
+export interface SemanticHighlightingParams {
+
+	/**
+	 * The text document that has to be decorated with the semantic highlighting information.
+	 */
+	textDocument: VersionedTextDocumentIdentifier;
+
+	/**
+	 * An array of semantic highlighting information.
+	 */
+	lines: SemanticHighlightingInformation[];
+
+}
+
+/**
+ * Represents a semantic highlighting information that has to be applied on a specific line of the text document.
+ */
+export interface SemanticHighlightingInformation {
+
+	/**
+	 * The zero-based line position in the text document.
+	 */
+	line: number;
+
+	/**
+	 * A base64 encoded string representing every single highlighted characters with its start position, length and the "lookup table" index of
+	 * of the semantic highlighting [TextMate scopes](https://manual.macromates.com/en/language_grammars).
+	 * If the `tokens` is empty or not defined, then no highlighted positions are available for the line.
+	 */
+	tokens?: string;
+
+}
+```
+
+_Tokens_:
+
+Tokens are encoded in a memory friendly way straight from the wire. The `tokens` string encapsulates multiple tokens as a `base64` encoded string. A single semantic highlighting token can be interpreted as a range with additional TextMate scopes information. The following properties can be inferred from a single token: `character` is the zero-based offset where the range starts. It is represented as a 32-bit unsigned integer. The `length` property is the length of the range a semantic highlighting token. And finally, it also carries the TextMate `scope` information as an integer between zero and 2<sup>16</sup>-1 (inclusive) values. Clients must reuse the `scopes` "lookup table" from the `initialize` request if they want to map the `scope` index value to the actual TextMate scopes represented as a string.
+
+_Encoding the Tokens_:
+
+Following example shows how three individual tokens are encoded into its `base64` form.
+Let assume, there is a series of token information (`[12, 15, 1, 2, 5, 0, 7, 1000, 1]`) that can be interpreted as the following.
+```json
+[
+	{
+		"character": 12,
+		"length": 15,
+		"scope": 1
+	},
+	{
+		"character": 2,
+		"length": 5,
+		"scope": 0
+	},
+	{
+		"character": 7,
+		"length": 1000,
+		"scope": 1
+	}
+]
+```
+The `character` (`12` )property will be stored as is but the `length` (`15`) and the `scope` (`1`) will be stored as a single 32-bit unsigned integer. The initial value of this 32-bit unsigned integer is zero. First, we set the value of the `length`, then we make some room (2<sup>16</sup>) for the `scope` by shifting the `length` 16 times to the left and applying a bitwise OR with the value of the `scope`.
+```
+00000000000000000000000000000000 // initial
+00000000000000000000000000001111 // set the `length` value (15)
+00000000000011110000000000000000 // shift [<< 0x0000010] the `length` and make some space for the scope
+00000000000011110000000000000001 // bitwise OR the `scope` value (1)
+```
--- a/protocol/src/protocol.semanticHighlighting.proposed.ts
+++ b/protocol/src/protocol.semanticHighlighting.proposed.ts
@@ -1,0 +1,91 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) TypeFox. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import { NotificationType } from 'vscode-jsonrpc';
+import { VersionedTextDocumentIdentifier } from 'vscode-languageserver-types';
+
+/**
+ * Parameters for the semantic highlighting (server-side) push notification.
+ */
+export interface SemanticHighlightingParams {
+
+	/**
+	 * The text document that has to be decorated with the semantic highlighting information.
+	 */
+	textDocument: VersionedTextDocumentIdentifier;
+
+	/**
+	 * An array of semantic highlighting information.
+	 */
+	lines: SemanticHighlightingInformation[];
+
+}
+
+/**
+ * Represents a semantic highlighting information that has to be applied on a specific line of the text document.
+ */
+export interface SemanticHighlightingInformation {
+
+	/**
+	 * The zero-based line position in the text document.
+	 */
+	line: number;
+
+	/**
+	 * A base64 encoded string representing every single highlighted characters with its start position, length and the "lookup table" index of
+	 * of the semantic highlighting [TextMate scopes](https://manual.macromates.com/en/language_grammars).
+	 * If the `tokens` is empty or not defined, then no highlighted positions are available for the line.
+	 */
+	tokens?: string;
+
+}
+
+/**
+ * Language server push notification providing the semantic highlighting information for a text document.
+ */
+export namespace SemanticHighlightingNotification {
+	export const type = new NotificationType<SemanticHighlightingParams, void>('textDocument/semanticHighlighting');
+}
+
+/**
+ * Capability that has to be set by the language client if that supports the semantic highlighting feature for the text documents.
+ */
+export interface SemanticHighlightingClientCapabilities {
+
+	/**
+	 * The text document client capabilities.
+	 */
+	textDocument?: {
+
+		/**
+		 * The client's semantic highlighting capability.
+		 */
+		semanticHighlightingCapabilities?: {
+
+			/**
+			 * `true` if the client supports semantic highlighting support text documents. Otherwise, `false`. It is `false` by default.
+			 */
+			semanticHighlighting: boolean;
+
+		}
+
+	}
+}
+
+/**
+ * Semantic highlighting server capabilities.
+ */
+export interface SemanticHighlightingServerCapabilities {
+
+	/**
+	 * A "lookup table" of semantic highlighting [TextMate scopes](https://manual.macromates.com/en/language_grammars)
+	 * supported by the language server. If not defined or empty, then the server does not support the semantic highlighting
+	 * feature. Otherwise, clients should reuse this "lookup table" when receiving semantic highlighting notifications from
+	 * the server.
+	 */
+	scopes?: string[][];
+
+}
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "vscode-languageclient",
-	"version": "4.3.0",
+	"version": "4.4.0",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
@@ -10,10 +10,10 @@
 			"integrity": "sha1-c7Xuyj+rZT49P5Qis0GtQiBdyWU=",
 			"dev": true,
 			"requires": {
-				"co": "^4.6.0",
-				"fast-deep-equal": "^1.0.0",
-				"fast-json-stable-stringify": "^2.0.0",
-				"json-schema-traverse": "^0.3.0"
+				"co": "4.6.0",
+				"fast-deep-equal": "1.1.0",
+				"fast-json-stable-stringify": "2.0.0",
+				"json-schema-traverse": "0.3.1"
 			}
 		},
 		"ansi-cyan": {
@@ -46,8 +46,8 @@
 			"integrity": "sha1-aHwydYFjWI/vfeezb6vklesaOZo=",
 			"dev": true,
 			"requires": {
-				"arr-flatten": "^1.0.1",
-				"array-slice": "^0.2.3"
+				"arr-flatten": "1.1.0",
+				"array-slice": "0.2.3"
 			}
 		},
 		"arr-flatten": {
@@ -80,7 +80,7 @@
 			"integrity": "sha1-mjRBDk9OPaI96jdb5b5w8kd47Dk=",
 			"dev": true,
 			"requires": {
-				"array-uniq": "^1.0.1"
+				"array-uniq": "1.0.3"
 			}
 		},
 		"array-uniq": {
@@ -144,7 +144,7 @@
 			"dev": true,
 			"optional": true,
 			"requires": {
-				"tweetnacl": "^0.14.3"
+				"tweetnacl": "0.14.5"
 			}
 		},
 		"block-stream": {
@@ -153,7 +153,7 @@
 			"integrity": "sha1-E+v+d4oDIFz+A3UUgeu0szAMEmo=",
 			"dev": true,
 			"requires": {
-				"inherits": "~2.0.0"
+				"inherits": "2.0.3"
 			}
 		},
 		"brace-expansion": {
@@ -162,7 +162,7 @@
 			"integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
 			"dev": true,
 			"requires": {
-				"balanced-match": "^1.0.0",
+				"balanced-match": "1.0.0",
 				"concat-map": "0.0.1"
 			}
 		},
@@ -172,9 +172,9 @@
 			"integrity": "sha1-uneWLhLf+WnWt2cR6RS3N4V79qc=",
 			"dev": true,
 			"requires": {
-				"expand-range": "^1.8.1",
-				"preserve": "^0.2.0",
-				"repeat-element": "^1.1.2"
+				"expand-range": "1.8.2",
+				"preserve": "0.2.0",
+				"repeat-element": "1.1.2"
 			}
 		},
 		"browser-stdout": {
@@ -225,9 +225,9 @@
 			"integrity": "sha512-Bq6+4t+lbM8vhTs/Bef5c5AdEMtapp/iFb6+s4/Hh9MVTt8OLKH7ZOOZSCT+Ys7hsHvqv0GuMPJ1lnQJVHvxpg==",
 			"dev": true,
 			"requires": {
-				"inherits": "^2.0.1",
-				"process-nextick-args": "^2.0.0",
-				"readable-stream": "^2.3.5"
+				"inherits": "2.0.3",
+				"process-nextick-args": "2.0.0",
+				"readable-stream": "2.3.6"
 			}
 		},
 		"co": {
@@ -242,7 +242,7 @@
 			"integrity": "sha1-cj599ugBrFYTETp+RFqbactjKBg=",
 			"dev": true,
 			"requires": {
-				"delayed-stream": "~1.0.0"
+				"delayed-stream": "1.0.0"
 			}
 		},
 		"commander": {
@@ -275,7 +275,7 @@
 			"integrity": "sha1-hTz6D3y+L+1d4gMmuN1YEDX24vA=",
 			"dev": true,
 			"requires": {
-				"assert-plus": "^1.0.0"
+				"assert-plus": "1.0.0"
 			}
 		},
 		"debug": {
@@ -293,7 +293,7 @@
 			"integrity": "sha1-sJJ0O+hCfcYh6gBnzex+cN0Z83s=",
 			"dev": true,
 			"requires": {
-				"is-obj": "^1.0.0"
+				"is-obj": "1.0.1"
 			}
 		},
 		"delayed-stream": {
@@ -320,10 +320,10 @@
 			"integrity": "sha512-fO3Di4tBKJpYTFHAxTU00BcfWMY9w24r/x21a6rZRbsD/ToUgGxsMbiGRmB7uVAXeGKXD9MwiLZa5E97EVgIRQ==",
 			"dev": true,
 			"requires": {
-				"end-of-stream": "^1.0.0",
-				"inherits": "^2.0.1",
-				"readable-stream": "^2.0.0",
-				"stream-shift": "^1.0.0"
+				"end-of-stream": "1.4.1",
+				"inherits": "2.0.3",
+				"readable-stream": "2.3.6",
+				"stream-shift": "1.0.0"
 			}
 		},
 		"ecc-jsbn": {
@@ -333,7 +333,7 @@
 			"dev": true,
 			"optional": true,
 			"requires": {
-				"jsbn": "~0.1.0"
+				"jsbn": "0.1.1"
 			}
 		},
 		"end-of-stream": {
@@ -342,7 +342,7 @@
 			"integrity": "sha512-1MkrZNvWTKCaigbn+W15elq2BB/L22nqrSY5DKlo3X6+vclJm8Bb5djXJBmEX6fS3+zCh/F4VBK5Z2KxJt4s2Q==",
 			"dev": true,
 			"requires": {
-				"once": "^1.4.0"
+				"once": "1.4.0"
 			}
 		},
 		"es6-object-assign": {
@@ -359,7 +359,7 @@
 		},
 		"event-stream": {
 			"version": "3.3.4",
-			"resolved": "http://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
+			"resolved": "https://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
 			"integrity": "sha1-SrTJoPWlTbkzi0w02Gv86PSzVXE=",
 			"dev": true,
 			"requires": {
@@ -378,7 +378,7 @@
 			"integrity": "sha1-3wcoTjQqgHzXM6xa9yQR5YHRF3s=",
 			"dev": true,
 			"requires": {
-				"is-posix-bracket": "^0.1.0"
+				"is-posix-bracket": "0.1.1"
 			}
 		},
 		"expand-range": {
@@ -387,7 +387,7 @@
 			"integrity": "sha1-opnv/TNf4nIeuujiV+x5ZE/IUzc=",
 			"dev": true,
 			"requires": {
-				"fill-range": "^2.1.0"
+				"fill-range": "2.2.4"
 			}
 		},
 		"extend": {
@@ -402,7 +402,7 @@
 			"integrity": "sha1-Gda/lN/AnXa6cR85uHLSH/TdkHE=",
 			"dev": true,
 			"requires": {
-				"kind-of": "^1.1.0"
+				"kind-of": "1.1.0"
 			}
 		},
 		"extglob": {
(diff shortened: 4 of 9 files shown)


protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,88 @@
...
+	/**
+	 * The length of the token. If negative, defaults to `0`. If the sum of the `character` and the `length` exceeds the length of the line,
+	 * it defaults back to the end of the line.
+	 */
+	length: number;
+
+	/**
+	 * An array of semantic highlighting scopes for the current token.
+	 */
+	scopes: string[];
```
svenefftinge: Documentation should point out that this corresponds to TextMate scopes. https://manual.macromates.com/en/language_grammars

protocol/src/protocol.semanticHighlighting.proposed.ts
```diff
@@ -0,0 +1,90 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
```
svenefftinge: I don't think it's MS copyright

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,88 @@
...
+	 * `true` if the client has the semantic highlighting support for the text document. Otherwise, `false`. It is `false` by default.
+	 */
+	semanticHighlighting?: boolean;
+}
+```
+
+##### SemanticHighlighting Notification
+
+The `textDocument/semanticHighlighting` notification is pushed from the server to the client to inform the client about additional semantic highlighting information that has to be applied on the text document.
+
```
svenefftinge: It should explain that incremental updates are done on a line-by-line basis.

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,89 @@
+#### Semantic Highlighting
+
+While the syntax highlighting are done on the client-side and can handle keywords, strings, and other low-level tokens from the grammar, it cannot adequately support complex coloring. Semantic highlighting information is calculated on the language server and pushed to the client as a notification. This notification carries information about the ranges that have to be colored. The desired coloring details are given as [TextMate scopes](https://manual.macromates.com/en/language_grammars) for each affected range. For the semantic highlighting information the following additions are proposed:
+
+_Client Capabilities_:
+
+Capability that has to be set by the language client if that can accept and process the semantic highlighting information received from the server.
```
rcjsuen: `if that can` should be `if it can`.

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,89 @@
...
+
+```ts
+/**
+ * The text document client capabilities.
+ */
+workspace: {
```
rcjsuen: I think you meant to name this variable `textDocument`?

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,89 @@
...
+
+	/**
+	 * `true` if the client has the semantic highlighting support for the text document. Otherwise, `false`. It is `false` by default.
```
rcjsuen: `true` if the client has the semantic highlighting support for the text document.
...to...
`true` if the client supports semantic highlighting support text documents.

jacobdufault:
cquery implements semantic highlighting as a notification the client sends to the server and supports rainbow highlighting (ie, each variable is a different shade of the same color, which makes flow analysis easier). Here is what the protocol looks like:

```cpp
struct Symbol {
  // a stable ID that can be used for rainbow highlighting; ie, this symbol will have
  // the same stableId across multiple notifications (w.r.t. each file)
  int stableId = 0;
  // what of type of symbol declares this symbol? ie, is this a free-standing variable
  // or an instance variable of a class
  SymbolKind parentKind; 
  // what type of symbol is this, ie, variable, func, etc
  SymbolKind kind;
  // what's the memory storage for this variable? ie, is it a global variable?
  StorageClass storage;
  // locations in the document where the symbol is used
  Range[] ranges;
};

// Standard notification
struct CqueryPublishSemanticHighlightingNotification {
  struct Params {
    // The file this applies to
    DocumentUri uri;
    // A list of symbols in the file that should be highlighted
    Symbol[] symbols;
  };
  Params params;
};
```

svenefftinge:
> cquery implements semantic highlighting as a notification the client sends to the server

You meant the inverse, i.e. notifications from the server to the client, right?

I think the main difference between this proposal and the protocol you have in cquery is, that we rely on textmate scopes rather than passing semantic information to the client. I don't see why your use cases shouldn't be covered. The server just needs to send the right scopes and the client needs to support them.

Also, this proposal doesn't support multi-line tokens as the server publishes symbols on a per line basis. This is so it can update incrementally.

jacobdufault:
> You meant the inverse, i.e. notifications from the server to the client, right?

Oops, yep. cquery binary sends a notification to the client, ie, vscode.

> I think the main difference between this proposal and the protocol you have in cquery is, that we rely on textmate scopes rather than passing semantic information to the client. I don't see why your use cases shouldn't be covered. The server just needs to send the right scopes and the client needs to support them.

Does the current proposal support rainbow style semantic highlighting, where each unique variable has a similar but slightly different shade of the same color? I'm not familiar enough with textmate scopes.

As for multi-line tokens, this seems like a rare use-case and is not one that cquery needs.

svenefftinge:
> Does the current proposal support rainbow style semantic highlighting, where each unique variable has a similar but slightly different shade of the same color? I'm not familiar enough with textmate scopes.

See https://www.sublimetext.com/docs/3/scope_naming.html for more info on textmate scopes.

It does not directly support rainbow style, but I think it could be done by sending corresponding scopes. For instance, the LS could send `variable.other.c1` to `variable.other.c9`. If the color theme supports it you will get rainbow colors, if not the more general `variable` coloring would apply.

Levertion:
Why is this in [Microsoft/vscode-languageserver-node](https://github.com/Microsoft/vscode-languageserver-node), rather than [Microsoft/language-server-protocol](https://github.com/Microsoft/language-server-protocol)? I presume that it's just to allow the development of the reference implementation and the proposal to be linked?

svenefftinge:
> Why is this in Microsoft/vscode-languageserver-node, rather than Microsoft/language-server-protocol? 

The process is described here https://github.com/Microsoft/language-server-protocol/blob/master/contributing.md
So this is the mandatory reference implementation for the VSCode language client library.
The problem is, that the VSCode API doesn't expose enough API to implement it. E.g. there is no way to access the TextMate scheme registry.
@dbaeumer what would be the way to go in such cases?

Levertion:
Fair enough. I just didn't know why Microsoft/language-server-protocol#124 was closed, but that explains it.

client/src/semanticHighlighting.proposed.ts
```diff
@@ -0,0 +1,99 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
```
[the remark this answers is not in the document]
kittaakos: Thanks for the remark, @rcjsuen. I already had the same feedback from @svenefftinge. See here: https://github.com/Microsoft/vscode-languageserver-node/pull/367#discussion_r198430700

I do not know why this PR is not picking up my change:
https://github.com/kittaakos/vscode-languageserver-node/blob/e773d7c2569414a8ffe9dccfed8f79e068713a82/protocol/src/protocol.semanticHighlighting.proposed.ts#L1-L4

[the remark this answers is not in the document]
kittaakos: I have compared another module. Sorry for the noise.

svenefftinge:
We should open a corresponding PR on language-server-protocol, with the relevant changes for the markdown file. @kittaakos WDYT?

dbaeumer:
@svenefftinge @rcjsuen actually the document here https://github.com/Microsoft/language-server-protocol/blob/master/contributing.md says that

> Proposed extensions need to be announced via an issue in this repository.

where this repository is https://github.com/Microsoft/language-server-protocol

Agree that this can be made more clear by moving it to the top.

dbaeumer:
Moved to the top. I am open to accepting a reference implementation for a different well used client as well.

kittaakos:
Thanks for the clarification, @dbaeumer. I open one for the proposal in the https://github.com/Microsoft/language-server-protocol repository too.

> Agree that this can be made more clear by moving it to the top.

Just naming the repository instead of having a link in the README would help too.

svenefftinge:
> I am open to accepting a reference implementation for a different well used client as well.

Ok, good to know. It would be very cool, to have support for it in VSCode, though.
Is there a chance you are exposing a bit more API, so it could work? 
Also, what do you think about the incremental updated based on lines?

alexandrudima:
I think it is a good idea to push the semantic highlights from the server. The only thing to be aware here would be the case that there are 100 opened files in the client and that only 1 file is visible, so the semantic highlights would only be useful for a single file. @dbaeumer I don't know if the protocol has visibility APIs yet, such that the server could limit its pushing of semantic highlights to the currently visible files or the currently visible regions of files?

Having worked before on the problem of how to represent in a memory-friendly way tokens (see https://code.visualstudio.com/blogs/2017/02/08/syntax-highlighting-optimizations), and having dealt with a lot of problems stemming from large files, may I suggest that the protocol in this area be designed with the following goals in mind:

1. tokens should be encoded in a memory friendly way straight from the wire (JSON).

2. there should be a way for the server to incrementally update the semantic highlights.

---

I think goal 1 is somewhat straight-forward to achieve. I suggest that each token does not point to an array of scopes (i.e. `SemanticHighlightingToken.scopes`), but rather the entire set of possible scopes are registered statically up front and this field should be replaced with an `index` into the legend. My assumption is that the possible number and types of semantic tokens is finite. I also believe using scopes (which can be targeted by TM themes) it is a good abstraction level. i.e. we should not talk about colors at this layer, as that is something for themes to decide.

So, assuming the entire possible set of scopes is registered/sent over up front, the token interface can be reduced to:
```ts
interface SemanticHighlightingInformation {
  line: number;
  tokens: SemanticHighlightingToken[];
}
interface SemanticHighlightingToken {
  character: number;
  length: number;
  scopes: number;
}
```

The next immediate memory/size optimisation here is to drop the `SemanticHighlightingToken` object entirely and inline everything in `SemanticHighlightingInformation`. i.e.

```ts
interface SemanticHighlightingInformation {
  line: number;
  tokens: number[]; // 3*i is character, 3*i+1 is length, 3*i+2 is scopes
}
```

Other optimisations could be pursued, such as noticing that the length is typically a small integer, and if we assume that the possible number of scopes is smaller than 2^16, we can get away with two 32 bit unsigned integers per token:
```ts
interface SemanticHighlightingInformation {
  line: number;
  tokens: number[]; // 2*i is character, (2*i+1)&0xffff0000>>16 is length, (2*i+1)&0x0000ffff is scopes
}
```

In terms of TypeScript, this would be best represented using typed arrays, `UInt32Array`:
```ts
interface SemanticHighlightingInformation {
  line: number;
  tokens: UInt32Array[];
}
```

The source for an `Uint32Array` is always an `ArrayBuffer`, so we could just store that:
```ts
interface SemanticHighlightingInformation {
  line: number;
  tokens: ArrayBuffer;
}
```

And, finally, the best and easiest way I'm aware of for efficiently encoding a byte array in JSON is to use base64 encoding, so at the wire level:
```ts
interface SemanticHighlightingInformation {
  line: number;
  tokens: string; // base64 encoded ArrayBuffer that should be viewed as an UInt32Array
}
```

---

Now, the second point is a bit more difficult as we would need to clearly specify what is expected of the client when edits occur. For example, we need to specify how the tokens should be backed by "live-markers" such that they *stick* with the text. i.e. the server and the client need some shared understanding on how the semantic tokens are adjusted. Second, we need for the notification event to be able to replace a certain region in the file with new tokens.

For example, when opening a large file in an editor, let's suppose the server sends a notification (which must include the document version id) which defines all the semantic tokens in the entire file. Now, when an edit occurs, for example, a new line is pressed on the first position in a file, both client and server should share the understanding that all the previously set tokens have been moved by one line (they stick with the text they're sitting on). So, when pressing a new line on the first position in a file, the server needs not to send any notification to update the semantic tokens.

Another example would be typing a character. This gets sent to the server as a text delta which mentions the position and the new inserted character. Similarly, the server should be able to send a notification of a size in direct proportion with the amount of semantic tokens changes. If the user typed a character in an identifier, then a notification updating only that specific semantic token should suffice.

I'm sorry I'm just describing the general mechanism here and not providing a sketch like above, but again, this is a much more complicated goal to achieve. Anyways, the requirements on the client can be reduced if there is some sort of negotiation where the client indicates that it cannot track semantic tokens, in which case the server would always send the tokens for the entire file. (You can think of this as something similar to formatting; there are some formatting providers which are very good and send N edits that target precisely the whitespace that needs to be changed... and then there are formatting providers that send a single huge edit encompassing the entire file and replacing it with new text). In other words, this should be done in a way where a smart client and a smart server can manage to achieve the correct maintenance of a high volume of semantic tokens as efficiently as possible.

svenefftinge:
Thanks, @alexandrudima for the feedback! Optimizing the protocol like you suggested makes sense to me. I saw you are doing something similar in the tokenizer2 interfaces in vs code, too?

> The only thing to be aware here would be the case that there are 100 opened files in the client and that only 1 file is visible, so the semantic highlights would only be useful for a single file.

Given that token updates would be sent after a change only, non-visible files might not be too heavily affected during typical editing.

Regarding incremental updates, this PR is proposing updated on a line by line basis. I.e. if the server sends tokens for a certain line the full line gets updated. I believe it is a good compromise between simplicity and efficiency, as it provides a similar life-cycle to the diagnostics, just with a smaller granularity (line rather than full document).
It also seems quite practical as lines are reasonably short. With e.g. minified code it could be not so great :), but I would say that is a corner case for which a server should optimize (e.g. disable).

Of course, this means that tokens need to stick with text regions on the client. I.e. if I insert or delete text all tokens are moved along.
AFAIK it is the default behavior for decorations in VS Code and annotations in Eclipse.
But still we would need to write that algorithm down.

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,89 @@
...
+}
+```
+
+##### SemanticHighlighting Notification
+
+The `textDocument/semanticHighlighting` notification is pushed from the server to the client to inform the client about additional semantic highlighting information that has to be applied on the text document. It is the server's responsibility to decide which lines are included in the highlighting information. In other words, the server is capable of sending only a delta information. For instance, after opening the text document (`DidOpenTextDocumentNotification`) the server sends the semantic highlighting information for the entire document, but if the server receives a `DidChangeTextDocumentNotification`, it pushes the information only about the affected lines in the document.
+
+_Notification_:
+
+* method: 'workspace/semanticHighlighting'
```
mostafaeweda: should this be `textDocument/semanticHighlighting` ?

matklad:
>Given that token updates would be sent after a change only, non-visible files might not be too heavily affected during typical editing.

Not saying that this is typical, but other files might be affect. A common rule for highlighting is that you color definitions based on kind of definition (class/function/module) and then you color references the same color as the definition it resolves to. In this scheme, changing a popular definition affects highlights in many file.

For general subscription mechanisms, I suggest taking a look at the Dart protocol: https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#request_analysis.setSubscriptions.

matklad:
A drive-by observation: if we ask only for highlights in a visible region, the amount of works per highlight becomes more-or-less constant and independent of size of the document. 

That is, I think that we don't necessary need *incremental* highlighting if we can filter by viewport? And such filtering should be much simpler to implement than incremental highlighting updates, because it does not require client and server to share state.

kittaakos:
@alexandrudima, @dbaeumer 

I have prepared the first prototype for the LSP extension.
 - A reference Java LS implementation is available here: https://github.com/eclipse/eclipse.jdt.ls/pull/746
 - A reference client implementation with Eclipse Theia (using the monaco API) is available here: https://github.com/theia-ide/theia/pull/2332
 - Here is the Java implementation of the proposed LSP change: https://github.com/eclipse/lsp4j/pull/198

The protocol uses the [proposed binary format](
https://github.com/Microsoft/vscode-languageserver-node/pull/367#issuecomment-403085124) for the communication. Delta updates are supported.

Supporting it in VS Code would be great, unfortunately, the `TokenTheme` and the `TokenMetadata` is not exposed via the current VS Code API.

 - The `TokenTheme` would be used to get the metadata for the TM scope: https://github.com/theia-ide/theia/pull/2332/files#diff-004ad6cff101057395eb8fc53f88c913R130
 - The`TokenMetadata` would be used to get the CSS inline class name for the decoration: https://github.com/theia-ide/theia/pull/2332/files#diff-004ad6cff101057395eb8fc53f88c913R131

Static:
static_member

Member vs. Parameter:
member-vs-parameter

Deprecated:
deprecated

Auto (un)boxing:
autoboxing

dbaeumer:
@kittaakos nice work. Is there any PR that defined the changes to the protocol in terms of TS interface?

kittaakos:
> Is there any PR that defined the changes to the protocol in terms of TS interface?

@dbaeumer, I am a bit puzzled. This PR contains the [proposed protocol changes in TS](https://github.com/Microsoft/vscode-languageserver-node/pull/367/files#diff-ac2186ae4b8ba5cc9f17deebc42c9e28). The [`SemanticHighlightingFeature`](https://github.com/Microsoft/vscode-languageserver-node/pull/367/files#diff-b88987665ddcee661243b054585df99e) for the client is just a mock, though. What TS interfaces do you mean? Thanks!

dbaeumer:
@kittaakos sorry I miss that.

We will look at the PR. We are also currently defining VS Code's roadmap for the next 6 months and support for semantic highlighting will be on that roadmap. I would also like us to validate the implementation in a second editor (e.g. VS Code) since this is for sure more complicated then defining a Goto definition request :-).

sanboxrunner:
@dbaeumer any update on VSC exposing semantic highlighting?

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,134 @@
...
+
+_Server Capabilities_:
+
+If the client declares its capabilities with respect to the semantic highlighting feature, and if the server supports this feature too, the server should set all the available TextMate scopes as a "lookup table" during the `initialize` request.
+
+```ts
+/**
+ * Semantic highlighting server capabilities.
+ */
+semanticHighlighting?: {
```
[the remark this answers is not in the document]
rcjsuen: Also note that instead of directly referencing a `{ }` instance, you should probably create and refer to a separate, new `SemanticHighlightingOptions` interface kind of like `DocumentLinkOptions` or `DocumentOnTypeFormattingOptions`.

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,134 @@
...
+
+The server never sends delta notifications, if no new semantic highlighting ranges were introduced but the existing onces have been shifted. For instance, when inserting a new line to the very beginning of the text document. The server receives the `DidOpenTextDocumentNotification`, updates its internal state, so that the client and the server shares the same understanding about the highlighted positions but the server does not send any notifications to the client. In such cases, it is the client's responsibility to track this event and shift all existing markers.
+
+The server can send a `SemanticHighlightingInformation` to the client without defining the `tokens` string. This means, the client must discard all semantic highlighting information in the line. For instance when commenting out a line.
+
+_Notification_:
+
+* method: 'workspace/semanticHighlighting'
```
rcjsuen:
```suggestion
* method: 'textDocument/semanticHighlighting'
```
I assume this should be `textDocument/semanticHighlighting` instead based on what's on line 52?

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,134 @@
...
+
+The server never sends delta notifications, if no new semantic highlighting ranges were introduced but the existing onces have been shifted. For instance, when inserting a new line to the very beginning of the text document. The server receives the `DidOpenTextDocumentNotification`, updates its internal state, so that the client and the server shares the same understanding about the highlighted positions but the server does not send any notifications to the client. In such cases, it is the client's responsibility to track this event and shift all existing markers.
```
rcjsuen:
```suggestion
The server will not send delta notifications if no new semantic highlighting ranges were introduced but the existing ones have simply been shifted. For instance, when inserting a new line to the very beginning of the text document. The server receives the `DidOpenTextDocumentNotification`, updates its internal state, so that the client and the server shares the same understanding about the highlighted positions but the server does not send any notifications to the client. In such cases, it is the client's responsibility to track this event and shift all existing markers.
```

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,134 @@
...
+```ts
+/**
+ * Parameters for the semantic highlighting (server-side) push notification.
+ */
+export interface SemanticHighlightingParams {
+
+	/**
+	 * The text document that has to be decorated with the semantic highlighting information.
+	 */
+	textDocument: VersionedTextDocumentIdentifier;
```
rcjsuen: All existing requests use `TextDocumentIdentifier` instead of a `VersionedTextDocumentIdentifier`. What is the reasoning behind making this a `VersionedTextDocumentIdentifier` instead?

Note that I'm not necessarily challenging this because I too have made a similar suggestion in the past regarding the pull model in a [different issue](https://github.com/Microsoft/language-server-protocol/issues/358#issuecomment-376845468).

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,134 @@
...
+	 * feature. Otherwise, clients should reuse this "lookup table" when receiving semantic highlighting notifications from
+	 * the server.
+	 */
+	scopes?: string[][];
+}
+```
+
+##### SemanticHighlighting Notification
+
+The `textDocument/semanticHighlighting` notification is pushed from the server to the client to inform the client about additional semantic highlighting information that has to be applied on the text document. It is the server's responsibility to decide which lines are included in the highlighting information. In other words, the server is capable of sending only a delta information. For instance, after opening the text document (`DidOpenTextDocumentNotification`) the server sends the semantic highlighting information for the entire document, but if the server receives a `DidChangeTextDocumentNotification`, it pushes the information only about the affected lines in the document.
```
HighCommander4: Since this is a server-to-client notification, should the protocol be called `textDocument/publishSemanticHighlighting`, to be consistent with `textDocument/publishDiagnostics`?

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,134 @@
...
+ */
+semanticHighlighting?: {
+
+	/**
+	 * A "lookup table" of semantic highlighting [TextMate scopes](https://manual.macromates.com/en/language_grammars)
+	 * supported by the language server. If not defined or empty, then the server does not support the semantic highlighting
+	 * feature. Otherwise, clients should reuse this "lookup table" when receiving semantic highlighting notifications from
+	 * the server.
+	 */
+	scopes?: string[][];
```
HighCommander4: I had a look at the link in the comment, but I still don't understand what this 2D array is expected to contain.

I assume every top-level element of the array represents one semantic highlighting scope, and the indexes encoded in the `tokens` string index into that top-level array.

If each of those top-level elements is itself an array of strings, what goes into those arrays? The name of the scope, which is one string... and what else?

Perhaps an example would help clarify the intended usage.

takanuva:
Any news on this one?

dbaeumer:
@takanuva since semantic coloring is a challanging feature to get right we would like to have at least an implementation for another editor (preferable VS Code :-)) as well to ensure this works as expected especially around performance. To make this happen we added this to the VS Code roadmap: https://github.com/Microsoft/vscode/wiki/Roadmap. So we hope to make progress in the next 6 months.

protocol/src/protocol.semanticHighlighting.proposed.md
MaskRay: Can the name be shortened to `textDocument/publishSemanticHighlight`? `highlight` can also be used as a noun and the naming would be consistent with `documentHighlight`.

protocol/src/protocol.semanticHighlighting.proposed.md
HighCommander4: Ok, I think I shed some light on this by looking at the [Java language server's implementation](https://github.com/eclipse/eclipse.jdt.ls/blob/6819b791a7f5d83e583dd830fd2a7f6948e5e80c/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/highlighting/SemanticHighlightings.java#L69).

IIUC, every semantic highlighting token kind has multiple scopes associated with it, ordered from most specific to least specific. For example, "static method" in Java has:

```
"storage.modifier.static.java", "entity.name.function.java", "meta.function-call.java", "meta.method.body.java", "meta.method.java", "meta.class.body.java", "meta.class.java",
"source.java"
```

That seems like an awful lot of scopes to me, but I guess the general idea is that it allows clients to use both coarse-grained token kinds (e.g. "static method" will match "function", so if a client is configured to highlight functions with a particular color, that will be applied to static methods) and fine-grained token kinds (if a client has a more specific color configured for "static methods", then that will take precedence over the color configured for functions).

Let me know if I'm understanding this correctly.

kjeremy:
What can we do to push this along?

krzyzanowskim:
Thanks for your work. This extension is highly demanded by community, and I see it being implemented by more and more LSP imolementations despite the fact not being officially approved.
It becomes de-facto standard, and this is not the most desirable situation for the community.

To whoever is decisive: what can we do to push it further?

dbaeumer:
Here is what has happened so far:

- in VS Code we started to separate the token store from the text model to allow for two things:
    - generating tokens async
    - contributing custom tokens from the extension host

- initial thinking how to define semantic tokens (we can't use colors :-)

As soon as this makes it into a proposed API state in VS Code we can definitely start to implement a PT to get a feeling how we best define LSP requests based on the two implementation we then have (Eclipse and VS Code)

egamma:
Please see also the discussion in https://github.com/microsoft/vscode/issues/77140

takanuva:
@dbaeumer, is there anything the community can do as of now to contribute to this? I'm really interested in seeing this feature in the LSP (mainly to keep the lexing and highlighting logic in the compiler/provider rather than in the text editor itself).

dbaeumer:
@takanuva I would like to wait a little until microsoft/vscode#77140 makes some progress.

yyoncho:
@kittaakos 
How the client is supposed to handle the case when the version does not match the current version? The client has to revert to the corresponding version, apply the changed and move back to the latest version?

HighCommander4:
Having contributed to an implementation of this proposed protocol in the C++ language server clangd, and used the result, my main feedback is this:

A flat list of highlighting types seems like an inefficient representation for all the things you may want to highlight differently in a given language.

To take C++, it's useful to highlight the following differently:
  * free functions vs. member functions
  * static functions vs. nonstatic functions
  * declarations of functions vs. references to functions

If expressed as a flat list, that's 8 different highlighting kinds:
   * declaration of static free function
   * reference to static free function
   * declaration to nonstatic free function
   * ... etc.

Instead, I would suggest representing a highlighting kind as a tuple of several different pieces of information, each classifying the entity along a given axis. For example:

  * **kind**: The fundamental language entity kind, e.g. function, variable, class, namespace, etc.
 * **parentKind**: The kind of the entity's parent. This covers distinctions like "free" (parentKind = namespace) vs. "member" (parentKind = class).
 * **storageClass**: Allows for distinguishing between additional properties of an entity, such as static vs. nonstatic. (The term "storage class" is somewhat C++-specific, I would be happy with a more generic term in the protocol.)
 * **role**: This classifies not the entity itself, but the specific occurrence being colored, allowing for distinctions between declaration, definition, reference, etc.

This is how the cquery language server, another C++ language server that implements semantic highlighting but using its own protocol, [does it](https://github.com/cquery-project/vscode-cquery/blob/4aac325dcf0735aa5d5f98960f28d6af6ca83d50/src/extension.ts#L84). I find this to be a much richer and more expressive representation than a flat list.

torokati44:
Since you mentioned how it's done with cquery, let me also refer to ccls (the solution in it seems rather similar) which also has semantic highlighting - in fact it assigns a unique shade to every single name/identifier:
https://github.com/MaskRay/vscode-ccls/blob/5d90ca19a3f51ebdeb68e3f3bd7dc82408941e6c/src/semantic.ts#L32

dbaeumer:
@HighCommander4 thanks for your feedback. Can I ask you to participate in the discussion [here](https://github.com/microsoft/vscode/issues/77133) which contains a more concrete proposal on how to specify these tokens.

@aeschli  please see https://github.com/microsoft/vscode-languageserver-node/pull/367#issuecomment-536363837

HighCommander4:
> @HighCommander4 thanks for your feedback. Can I ask you to participate in the discussion [here](https://github.com/microsoft/vscode/issues/77133) which contains a more concrete proposal on how to specify these tokens.

@dbaeumer Thanks, I've commented there. One question: that discussion seems to be about a VSCode extension API. Will its outcome inform the design of the LSP protocol for semantic highlighting as well?

HighCommander4:
> Since you mentioned how it's done with cquery, let me also refer to ccls (the solution in it seems rather similar) which also has semantic highlighting

The similarity is not accidental - ccls is a fork of cquery :)

> - in fact it assigns a unique shade to every single name/identifier:
> https://github.com/MaskRay/vscode-ccls/blob/5d90ca19a3f51ebdeb68e3f3bd7dc82408941e6c/src/semantic.ts#L32

Yeah, this is a feature called "rainbow semantic highlighting". cquery/ccls is the first I've encountered this. It's kind of neat, though it does involve more bookkeeping work for the server.

dbaeumer:
@HighCommander4 Regarding 

> Will its outcome inform the design of the LSP protocol for semantic highlighting as well?

Yes. @aeschli, @alexandrudima and me work closely on the API for VS Code to ensure it will make sense in LSP as well.

protocol/src/protocol.semanticHighlighting.proposed.md
milianw: As another implementer on the client side, I also have questions as to how this should be handled. @HighCommander4 when you say 

> every semantic highlighting token kind has multiple scopes associated with it

how does this relate to the definition of a token below? In the Encoding of Tokens section it says that a token has only a single scope index. How does this then map into the 2D space defined here?

When looking at the clangd server side output e.g. I see that the scopes array is 1D in practice anyways:

```
[["variable.other.cpp"],["entity.name.function.cpp"],["entity.name.function.method.cpp"],["variable.other.field.cpp"],["entity.name.type.class.cpp"],["entity.name.type.enum.cpp"],["variable.other.enummember.cpp"],["entity.name.namespace.cpp"],["entity.name.type.template.cpp"]]
```

Does someone know a server side implementation that produces a 2D array?

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,134 @@
...
+	 * A base64 encoded string representing every single highlighted characters with its start position, length and the "lookup table" index of
+	 * of the semantic highlighting [TextMate scopes](https://manual.macromates.com/en/language_grammars).
+	 * If the `tokens` is empty or not defined, then no highlighted positions are available for the line.
+	 */
+	tokens?: string;
+
+}
+```
+
+_Tokens_:
```
milianw: could this section please make the binary encoding of the base64 encoded data more explicit?

Basically the following things are missing in my opinion:
- the encoded string represents an array of 64-bit sized tokens
- size of the token string is thus size of the decoded string divided by 8
- what endianness is used for the unsigned integers in the token? system default? can we make this explicit and enforce little endian, assuming one can in principle run a lsp server on a big endian machine and connect to it from a little endian client or vice versa?

milianw: regarding endianness, it seems to be defined as big endian, cf. clangd source:

```
     // Writes the token to LineByteTokens in the byte format specified by the
      // LSP proposal. Described below.
      // |<---- 4 bytes ---->|<-- 2 bytes -->|<--- 2 bytes -->|
      // |    character      |  length       |    index       |

      write32be(Token.R.start.character, OS);
      write16be(Token.R.end.character - Token.R.start.character, OS);
      write16be(static_cast<int>(Token.Kind), OS);
```

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,134 @@
...
+
+Tokens are encoded in a memory friendly way straight from the wire. The `tokens` string encapsulates multiple tokens as a `base64` encoded string. A single semantic highlighting token can be interpreted as a range with additional TextMate scopes information. The following properties can be inferred from a single token: `character` is the zero-based offset where the range starts. It is represented as a 32-bit unsigned integer. The `length` property is the length of the range a semantic highlighting token. And finally, it also carries the TextMate `scope` information as an integer between zero and 2<sup>16</sup>-1 (inclusive) values. Clients must reuse the `scopes` "lookup table" from the `initialize` request if they want to map the `scope` index value to the actual TextMate scopes represented as a string.
```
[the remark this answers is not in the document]
HighCommander4: At the start of the line, since we already know what line this token string pertains to. I agree that the spec could be more explicit about this.

protocol/src/protocol.semanticHighlighting.proposed.md
HighCommander4: > Does someone know a server side implementation that produces a 2D array?

This [Java server implementation](https://github.com/eclipse/eclipse.jdt.ls/blob/6819b791a7f5d83e583dd830fd2a7f6948e5e80c/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/highlighting/SemanticHighlightings.java#L69) (which I also linked to in the comment that you're replying to) does. The line I link to contains a list of multiple scope strings associated with a single token kind, `staticFinalField`.

> As another implementer on the client side, I also have questions as to how this should be handled. @HighCommander4 when you say
> 
> > every semantic highlighting token kind has multiple scopes associated with it
> 
> how does this relate to the definition of a token below? In the Encoding of Tokens section it says that a token has only a single scope index. How does this then map into the 2D space defined here?

My understanding is that, after indexing into the 2D array with the scope index, the intention is for the client to loop through the resulting 1D array, which is expected to be ordered from "most specific" to "least specific" scope, and use the first one that has a matching style in the user's theme.

So, taking the `staticFinalField` example above, the client would first try the scope `"storage.modifier.static.java"`, and if that matches a style in the user's theme, then apply that. Otherwise, keep looping and try the next scope. This way, even if the user's theme didn't contain a more specific rule (e.g. one for static variables), they'll still get a coloring for a less specific one (e.g. one for variables).

protocol/src/protocol.semanticHighlighting.proposed.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+export interface SemanticHighlightingInformation {
+
+	/**
+	 * The zero-based line position in the text document.
+	 */
+	line: number;
+
+	/**
+	 * A base64 encoded string representing every single highlighted characters with its start position, length and the "lookup table" index of
```
Marwes: Which type of base64 encoding is used here? (standard, url_safe? padding?) (https://docs.rs/base64/0.11.0/base64/index.html#constants) cc https://github.com/gluon-lang/lsp-types/pull/127

protocol/src/protocol.semanticHighlighting.proposed.md
```diff
@@ -0,0 +1,134 @@
+#### Semantic Highlighting
+
+While the syntax highlighting is done on the client-side and can handle keywords, strings, and other low-level tokens from the grammar, it cannot adequately support complex coloring. Semantic highlighting information is calculated on the language server and pushed to the client as a notification. This notification carries information about the ranges that have to be colored. The desired coloring details are given as [TextMate scopes](https://manual.macromates.com/en/language_grammars) for each affected range. For the semantic highlighting information the following additions are proposed:
```
rmunn: It seems desirable to provide a way for the client to say "Send this range of text to a different language server if one is available", e.g. to provide a way to implement https://github.com/microsoft/vscode/issues/1751 (see https://github.com/microsoft/vscode/issues/77140#issuecomment-533953100 as well). This would allow, for example, an HTML language server to parse `onclick="event.preventDefault()"` and say "If a Javascript language server is available, hand it the `event.preventDefault()` text and let it parse that text instead". Other use cases where reparsing by a different language server would be useful include (but are not limited to):

* HTML in a string literal in a PHP file
* CSS in a `<style>` element in an HTML page
* SQL in a string literal in many different languages
* Markdown inside comments in many different languages
* Code embedded in Markdown in literate-programming documentation, Python `doctest` tests, etc

The common factor in all the above examples is that the "root" language server can identify the range of text in which a different language is embedded, and (usually) which language is embedded in that range. Therefore, one possible scope the language server may want to return is one with the meaning "Please make another semantic-highlighting request to a different language server that serves language XYZ, for this range of text".

To enable this scenario without making any change to the current protocol proposal, I propose a new root-level TextMate scope (in addition to `comment`, `keyword`, `string`, et al): `reparse`. A client that does not know about the `reparse` scope would ignore it; a client that understands it would understand a `reparse.language.markdown` scope as meaning "take the text in this range, and send it to a Markdown language server for parsing".

Since scopes (as seen below) are a 2D array, the scope that includes `reparse.language.markdown` could also include a second scope name that would be a fallback if no Markdown language server is available (e.g., if no editor extension is installed that provides Markdown parsing). This is most likely to be either a string or a comment. E.g., if entry 4 in the `scopes` array is `['reparse.language.markdown', 'string.quoted']` and entry 5 is `['reparse.language.markdown', 'comment.line.number-sign']`, then `"*bold*"` would return scope 4 while `# *bold*` would return scope 5.

There are other ways that https://github.com/microsoft/vscode/issues/1751 could be implemented in this API, such as adding a new `reparse` capability so that clients can signal whether or not they understand reparse messages, but that would involve a change in the API and might therefore hinder implementation by extensions that have already started to implement this. So I think a new top-level TextMate scope is probably the simplest way to achieve multi-language parsing.

matklad: Highlighting and injection seems orthogonal to me. You'd want JS code completion in `"event.preventDefault()"`, and it would be strange if completion would have to block on syntax highlighting of the whole file. 

Rather, we need a special "document/languageInjections" request/notification, which returns something like this:

```TypeScript
interface InjectionInfo {
    language: string,
    range: Range,
    // We often want to strip leading indent or `//` from the injected fragment, 
    // `rangeMapping` should do that. An example would be like this
    // ```javascript
    // console.log({
    //   key: "<- that '//' is not a part of javascript snippet!"
    // })
    // ```
    rangeMapping?: Map<Range, Range>, 
    // Sometimes, just removing some ranges is not enough, 
    // a full string escaping might be required
    //
    //    onclick="console.log(&quotHello World!&quot)"
    textContent?: string
}
```

Levertion: >   // Sometimes, just removing some ranges is not enough, 
    // a full string escaping might be required
    //
    //    onclick="console.log(&quot;Hello World!&quot;)"
    textContent?: string

I don't think this is enough either - how does the `secondary` language server know how `textContent` was reached.

<details>
<summary>Demonstration</summary>
For the purposes of example take a pathological case where `&quot;` actually expands to `Hello World! ` in the `onclick` handler. A more realistic scenario would probably involve some backslashes

In this case, `console.log(&quot;Hello World!&quot;)` in the document would lead to a `textContent` of `console.log(Hello World! Hello World! Hello World! )`.

When hovering or typing at `Hello <|>World` in the document, how does the client know that the hover position know that the cursor position should be `31` [`console.log(Hello World! Hello <|>World! Hello World! )`] - from its perspective the one of the `&quot;`s may have expanded to each of ` ` or `Hello World! Hello World!`.
</details>

I think we need to make `rangeMapping` more general, i.e. replace it with a `TextEdit[]`, remove `textContent` and add an (optional?) `InjectionEncodeId`.

This would be passed in the `InjectionEncode` request, which is `InjectionEncodeParams -> TextEdit[]`, which turns 'in Injection' `TextEdit`s to 'out of Injection' `TextEdit`s.
```ts
interface InjectionEncodeParams {
	id: InjectionEncodeId, // Probably `number`
	edits: TextEdit[],
}
```

<details>
<summary>Worked Example</summary>

In the case of `onclick="console.log(&quot;Hello World!&quot;)"`, assuming that this is the entire document, the request would be:
`[server->client notification] "document/languageInjections"`

```jsonc
{
    "injections": [
        {
            "language": "javascript",
            "range": {
                "start": {
                    // Just after the open quote
                },
                "end": {
                    //Just after the close quote
                }
            },
            "edits": [
                {
                    "range": {
                        // Around the first `&quot;`
                    },
                    "newText": "\""
                },
                {
                    "range": {
                        // Around the second `&quot;`

                    },
                    "newText": "\""
                }
            ],
            "id": 0 // Some ID
        }
    ]
}
```

If the user then typed an edit (or accepted a completion) at "Hello <|>World", the client would make `InjectionEncode` request. This would (html) escape any quotes in the edit to be made for example.

N.B. I ran out of time to type out the request and responses in detail for this case.
</details>

Also, I think we would need some way of passing some sort of argument into this Injection, for example for Json Schemas for a sub file.

rcjsuen:
With 4c0963d6fdfd657044092c5d4db61a87b54471ef in `master`, not sure what is next for this pull request... 🤔

dbaeumer:
We spend quite some time having a end to end Implementation of this over the last couple of months to ensure that semantic tokens are (a) easy to provide (b) can be provided in a performant way (c) can be easily updated.

The biggest changes to the proposal @alexdima did a while back ago are:

- semantic token are a pull model not a push model. In general we realized that a pull model is better for clients and it doesn't require to sync UI state to the server (which editors are open / visible which ranges are visible, ...).
- servers can encode a delta to minimize the data to be transferred. We are also planning to add the concept of eTag to the protocol to have this feature for other areas as well. This reduces the data transfer cost in cases the client pulls for information which has not been changed.
- split of token into types and modifiers to easier describe the token.
- although there are predefined token types and modifiers in the protocol there will not be a standard. It is negotiated between the client and the server (see client capabilities and the SemanticTokenLegend). This has the advantage that every server can use its own set of token types and modifiers.

I would be great to get feedback on the proposed API. There is although an example in the testbed here (https://github.com/microsoft/vscode-languageserver-node/blob/master/testbed/server/src/server.ts#L475)

dbaeumer:
And I forgot: we will also add the concept of busy to the pull model where a server can indicate that it is currently a bad time to ask for that information and the client should try again later.

rcjsuen:
> I would be great to get feedback on the proposed API.

Got caught off-guard because of the deltas but after fixing it up it seems to be working okay in VS Code Insiders. Thanks you and @alexdima for your effort, @dbaeumer.

image

I've still got a long ways to go before I'm done and it will take me some time to figure out what tokens to use given that a Dockerfile is not quite the same as a regular programming language source file... 🤔

bsaviano-intersystems:
Are there any updates for when this pull request will make it into the master branch? Semantic tokens are a key part of our IDE strategy and we're very excited to add them to our language server!

rcjsuen:
> Are there any updates for when this pull request will make it into the master branch? Semantic tokens are a key part of our IDE strategy and we're very excited to add them to our language server!

@bsaviano-intersystems This proposal has been superseded by a new proposal. I suggest you look at the [draft 3.16 specification](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/) and share your thoughts in https://github.com/microsoft/language-server-protocol/issues/18.

bsaviano-intersystems:
@rcjsuen Thanks for the response. I've already seen the proposed spec in 3.16. I was wondering if there's a version of the vscode-languageserver/languageclient packages that contains the proposal so I can repurpose my existing DocumentSemanticTokenProvider and test it out. Thanks again.

rcjsuen:
> @rcjsuen I was wondering if there's a version of the vscode-languageserver/languageclient packages that contains the proposal so I can repurpose my existing DocumentSemanticTokenProvider and test it out. Thanks again.

@bsaviano-intersystems I know 6.1.1 and up for `vscode-languageserver` and 6.1.3 and up for `vscode-languageclient` should be good. I don't know off hand when it was first introduced though.

dbaeumer:
@rcjsuen thanks for answering. The latest server version is 7.0.0-next.6 and client 7.0.0-next.8

bsaviano-intersystems:
Thank you both for the responses. I will install the “next” packages and try it out. Looking forward to the official 3.16 release! Has a timeframe for that been established?

kjeremy:
@dbaeumer `npm outdated` shows me that `7.0.0-next.1` is the latest vscode-languageclient

dbaeumer:
@kjeremy that is incorrect. Let me check what happened.

dbaeumer:
That is strange since npm lists everything correctly

capture

What version do you have installed right now. Not sure if outdated works correctly with next versions.

kjeremy:
@dbaeumer 

> npm --version
6.13.7

> npm outdated

image

dbaeumer:
@kjeremy I mean which version of the `vscode-languageclient`

kjeremy:
@dbaeumer  7.0.0-next.1

dbaeumer:
@kjeremy what do you expect to happen. To my knowledge `next` versions are not considered to be semver. So npm will never install `7.0.0-next.2` if you prereq `7.0.0-next.1`

kjeremy:
@dbaeumer I expect that `7.0.0-next.8` shows up as the latest version available and it does not.

dbaeumer:
@kjeremy but this is then something that npm needs to deal with. IMO the releases are tag correct. Have you raised the issue with them?

Hebe970217:
excuse me？I see  KamasamaK Removing Semantic Highlighting,  no support on Oct 20, 2020 at after？

SPGoding:
That was the title of an issue on the eclipse/lsp4j repo to deprecate their Semantic Highlighting API in favor of LSP's Semantic Tokens API, unrelated to microsoft/vscode-languageserver-node.

Hebe970217:
> That was the title of an issue on the eclipse/lsp4j repo to deprecate their Semantic Highlighting API in favor of LSP's Semantic Tokens API, unrelated to microsoft/vscode-languageserver-node.

Okay.
lsp server init result is "documentHighlightProvider": true. 
I tried to use textdocument / documenthighlight to request the backend Clangd, and I didn't see the independent return result. I'm writing a CPP client

Hebe970217:
Maybe I should create a new issues

PR #372, opened 2018-07-03
Add preselect field to CompletionItem per VS Code 1.25

### LINKED ISSUE #371: Add 'preselect'  field to the CompletionItem
VS Code 1.25 adds new `preselect` field to the `CompletionItem`. The field allows completion providers to specify which item should be initially selected in the completion list. 

By default clients usually select first or last used items, which is limiting. A completion provider may choose to select different item based on other conditions, such as current code editor context, probability of use of certain items, and so on. 

This request is to add equivalent field to the CompletionItem structure therefore allowing use of preselection in VS Code from language servers as it is possible from the regular completion provider.

https://github.com/Microsoft/vscode/blob/9ba606df0ee5c2a386eac532771ac343a62f1985/src/vs/workbench/api/node/extHostTypes.ts

Per 
https://github.com/Microsoft/vscode/issues/35551
https://github.com/Microsoft/vscode/commit/62e0f34a3569f2cab4d16b2ef36dde285d39a9cf

MikhailArkhipov:
Fixes #371 

Do I need to rev the version up? Not sure.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -1,11 +1,11 @@
 {
 	"name": "vscode-languageclient",
 	"description": "VSCode Language client implementation",
-	"version": "4.2.1",
+	"version": "4.2.2",
 	"author": "Microsoft Corporation",
 	"license": "MIT",
 	"engines": {
-		"vscode": "^1.23"
+		"vscode": "^1.25"
 	},
 	"repository": {
 		"type": "git",
@@ -21,7 +21,7 @@
 		"vscode": "^1.1.18"
 	},
 	"dependencies": {
-		"vscode-languageserver-protocol": "^3.8.1"
+		"vscode-languageserver-protocol": "^3.8.2"
 	},
 	"scripts": {
 		"prepare": "npm run update-vscode",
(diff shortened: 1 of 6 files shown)


dbaeumer:
@MikhailArkhipov thanks for the PR. But changing the code is only half the work. This needs to be updated in the spec as well which is maintained here: https://github.com/Microsoft/language-server-protocol

I would suggest you open an item at https://github.com/Microsoft/language-server-protocol to get feedback on how we would best add the property. The problem is that this needs a capability as well since completion items travel from the client to the server as well (in code action calls) and servers might be surprised if the property got lost since clients other than VS Code will not support it.

MikhailArkhipov:
The property is already there and works in TypeScript. The problem is that it does not work from the language server. This issue is pretty urgent. Thanks!

MikhailArkhipov:
LSP PR https://github.com/Microsoft/language-server-protocol/pull/515

dbaeumer:
Thanks.

dbaeumer:
Closing. New version are available.

Issue #362, opened 2018-06-19
Color Provider and vscode-languageserver-types


nikeee:
I'm trying to implement a color provider referring only `vscode-languageserver-types`. However, I cannot find the definitions for `ColorInformation` inside that package. It seems that they are only in `vscode-languageserver-protocol`, which requires Node.js.

Should it also be available in `vscode-languageserver-types` or am I missing something?

dbaeumer:
Just to clarify: the current protocol implementation we provide depends on node. We have don now effort yet to provide a protocol implementation independent from node.

nikeee:
I'm only missing the types used in this API like `Color`, `ColorInformation` and `ColorPresentation`. I thought they would be in  `vscode-languageserver-types` lust like `Location`, `TextEdit` or `Command`. My current solution is to have a have file with the declarations of the three interfaces mentioned above in it.

*Should* they be in `types` of am I missing some point that makes the color provider different from the other ones?

aeschli:
Yes, these types can now go into vscode-languageserver-types. I'll add them.

PR #375, opened 2018-07-13
Address #362

nikeee:
Also, there was no `TextEdit.is`, which I also added.

Currently, this build fails. If the types package is published, we can remove the definitions in `protocol/src/protocol.colorProvider.ts` and use the ones shipped in `vscode-languageserver-types`.


dbaeumer:
@aeschli can you please review and merge.

PR #379, opened 2018-07-26
[client] expose builtin features and connection

akosyakov:
I am working on a new version of `monaco-languageclient` to catch it up with `vscode-languageclient`: https://github.com/TypeFox/monaco-languageclient/pull/89
And for it, I have to customize how a connection is created and which features are enabled by default.

cc @dbaeumer

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -23,7 +23,7 @@ import {
 	NotificationType, NotificationType0,
 	NotificationHandler, NotificationHandler0, GenericNotificationHandler,
 	MessageReader, MessageWriter, Trace, Tracer, Event, Emitter,
-	createProtocolConnection,
+	createProtocolConnection, ProtocolConnection,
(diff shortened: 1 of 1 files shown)


akosyakov:
Another case for it: a client wants to provide a custom implementation of some features. For example to narrow command registration to a specific editor.

dbaeumer:
I have to say I am actually against this. Making them API means that we need to keep them stable and maintain them. `main.ts` has currently a statement `export * from './client';` which would make them official API. 

If people want to remove feature and replace them with different once we need to offer API to remove features based on a `RPCMessageType`

dbaeumer:
The protocol npm module does contain methods to create connections. So I am wondering if it wouldn't be better to base the monaco client on the protocol and see if we should move things there instead of reusing the VS Code client which is based on the assumption to have full access to the VS Code API.

PR #385, opened 2018-07-28
Add CodeActionOptions with providedCodeActionKinds

Gama11:
See here for the motivation: https://github.com/Microsoft/language-server-protocol/issues/389#issuecomment-407717530

Corresponding language-server-protocol PR: https://github.com/Microsoft/language-server-protocol/pull/534


dbaeumer:
@Gama11 you forgot the dynamic registration and that a feature needs to keep the state per registration and can't almost never have a single global state. I will do the changes. You might want to look at them for the next PR :-)

PR #390, opened 2018-08-12
Add textDocument/prepareRename for preparing a rename request

rcjsuen:
Implemented a new `textDocument/resolveRenameLocation` request for preparing and validation before a `textDocument/rename` happens.

This request allows a language server to resolve and validate the current cursor location before a rename is actually attempted. This allows clients to interrupt a rename request if for example the user's text cursor is currently within a language keyword and as such cannot be renamed.

This addresses Microsoft/language-server-protocol#455.

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -49,11 +49,11 @@ import {
 	CodeLensRequest, CodeLensResolveRequest, CodeLensRegistrationOptions,
 	DocumentFormattingRequest, DocumentFormattingParams, DocumentRangeFormattingRequest, DocumentRangeFormattingParams,
 	DocumentOnTypeFormattingRequest, DocumentOnTypeFormattingParams, DocumentOnTypeFormattingRegistrationOptions,
-	RenameRequest, RenameParams,
+	RenameRequest, RenameParams, RenameRegistrationOptions, PrepareRenameRequest, TextDocumentPositionParams,
 	DocumentLinkRequest, DocumentLinkResolveRequest, DocumentLinkRegistrationOptions,
 	ExecuteCommandRequest, ExecuteCommandParams, ExecuteCommandRegistrationOptions,
 	ApplyWorkspaceEditRequest, ApplyWorkspaceEditParams, ApplyWorkspaceEditResponse,
-	MarkupKind, SymbolKind, CompletionItemKind, Command, CodeActionKind, DocumentSymbol, SymbolInformation,
+	MarkupKind, SymbolKind, CompletionItemKind, Command, CodeActionKind, DocumentSymbol, SymbolInformation, Range,
 	CodeActionRegistrationOptions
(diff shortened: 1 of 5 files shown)


dbaeumer:
@rcjsuen thanks a lot for the PR. As usually the implementation looks very good.

One thing  I would like to discuss though is the naming. The protocol uses the term resolve right now to resolve additional properties for an already existing item and a resolve request actually returns the same object type. 

To not confuse things I would not name that request resolve. Ideas are: `textDocument/prepareRename` or `textDocument/computeRenameInput`. I would not put location into it since the request computes more than the location. I am open for other ideas as well.

dbaeumer:
I like the names.

client/src/client.ts
```diff
@@ -1972,7 +1979,7 @@ class RenameFeature extends TextDocumentFeature<TextDocumentRegistrationOptions>
 		});
 	}
 
-	protected registerLanguageProvider(options: TextDocumentRegistrationOptions): Disposable {
+	protected registerLanguageProvider(options: RenameRegistrationOptions): Disposable {
```
dbaeumer: RenameRegistrationOptions should also be the type of the TextDocumentFeature (e.g. TextDocumentFeature<RenameRegistrationOptions>)

protocol/src/protocol.ts
```diff
@@ -808,7 +823,7 @@ export interface _ServerCapabilities {
 	/**
 	 * The server provides rename support.
 	 */
-	renameProvider?: boolean;
+	renameProvider?: boolean | RenameOptions;
```
dbaeumer: We need to document here that the RenameOptions type is only valid if the client signaled prepareSupport in its capabilities. Otherwise we would break old clients.

dbaeumer:
Two minor things. I will merge the branch and fix them.

@rcjsuen can you update the specification as well.

rcjsuen:
@dbaeumer I have fixed them already.

dbaeumer:
@rcjsuen just saw you already fixed them. Thanks. Nice work

Razzeee:
How would I go about returning an error? It seems to be in the spec, but I can't seem to find how it's implemented. What am I missing?

dbaeumer:
What do you mean by

> but I can't seem to find how it's implemented.

Errors are either returned by rejecting the promise or by throwing an Error.

Razzeee:
Thank you, that seems to go in the right direction. I was looking for a return type or something along the lines of `ResponseError` from the spec.

Rejecting the problem seems to not give back the message I set.

`throw Error("Test123");` seems to print this:
image

Can I get rid of the text that's not provided by me?

dbaeumer:
You need to throw a `ResponseError` to get full control over the message. This is done since I need to handle random exceptions as well.

PR #396, opened 2018-08-15
fix #394: make sure that document listeners are installed on restart

### LINKED ISSUE #394: DidOpenTextDocumentFeature should clean up a listener on dispose
It disposes `didOpen` listener but does not set it to `undefined`. So, on `register` a new listener is not installed, for example during restart.

akosyakov:


dbaeumer:
Good catch. Thanks!

PR #400, opened 2018-08-19
Change textDocument/foldingRange to textDocument/foldingRanges

rcjsuen:
According to the 3.10.0 specification, the request should be `textDocument/foldingRanges` and not `textDocument/foldingRange`. If you take a look at the [original specification in the Microsoft/vscode-languageserver-protocol-foldingprovider](https://github.com/Microsoft/vscode-languageserver-protocol-foldingprovider/blob/f0db4c8a93468888b3d7b4694acf8345c989a6cf/protocol.foldingProvider.md), you can see that the confusion was already there as both `textDocument/foldingRange` and `textDocument/foldingRanges` is mentioned.

### DIFF
--- a/protocol/src/protocol.foldingRange.ts
+++ b/protocol/src/protocol.foldingRange.ts
@@ -15,7 +15,7 @@ export interface FoldingRangeClientCapabilities {
 	 */
 	textDocument?: {
 		/**
-		 * Capabilities specific to `textDocument/foldingRange` requests
+		 * Capabilities specific to `textDocument/foldingRanges` requests
 		 */
 		foldingRange?: {
 			/**
@@ -116,5 +116,5 @@ export interface FoldingRangeRequestParam {
  * that resolves to such.
  */
 export namespace FoldingRangeRequest {
-	export const type: RequestType<FoldingRangeRequestParam, FoldingRange[] | null, any, any> = new RequestType('textDocument/foldingRange');
+	export const type: RequestType<FoldingRangeRequestParam, FoldingRange[] | null, any, any> = new RequestType('textDocument/foldingRanges');
 }


dbaeumer:
@aeschli this might be something we want to change sooner than later.

aeschli:
@rcjsuen Any objections if I correct the spec? There are already multiple implementors for VSCode that use the `textDocument/foldingRange`.
If you know of many users of `textDocument/foldingRanges` I can also look into compatibility code.

rcjsuen:
@aeschli It doesn't really matter to me personally either way as I am only implementing this for one little language server I write and develop myself.

Perhaps you can [raise an issue](https://github.com/Microsoft/language-server-protocol/issues) in the Microsoft/language-server-protocol repository first to see what other LSP clients have done thus far (as they may have implemented `textDocument/foldingRanges`?).

The reality is that even the [online specification](https://github.com/Microsoft/language-server-protocol/blob/6b9ef042b75678224e9fc605cfd3699f2ae7c74c/specification.md) is inconsistent. You can see both `textDocument/foldingRange` and `textDocument/foldingRanges` mentioned so depending on which you saw first you might have already written your constant (be it an LSP client or an LSP server) a certain way.

aeschli:
Yes, it's a bit embarrassing, sorry for causing all that. I think I'll take the risk and just fix the spec now, given that the implementation has been out there quite a bit.

rcjsuen:
@aeschli No need to apologize. It happens. I am partially guilty as well because I helped review the specification while it was in the proposal stage but I didn't catch it either... :P

rcjsuen:
@aeschli OK to close? I did a search in this repository and I don't actually see any references to `textDocument/foldingRanges` so I think we don't need to do anything in this particular repository?

aeschli:
Closing, yes, looks like this only affects the protocol repo.

aeschli:
Updated the spec (https://github.com/Microsoft/language-server-protocol/commit/09ecb5ecd97cc0888f591cfe2065dce4c6208df4)

PR #417, opened 2018-09-23
Support MarkupContent[] in a Hover

### LINKED ISSUE #374: Prettier Hover/MarkupContent "regions"
Hello,

I hope this is clear here. I am creating the basics of a language server right now. I am not very well experienced in this, so this point might be moot and I may be missing something.

I am trying to make a Hover request similar to the Typescript hover requests built into VSCode. Although I can make markdown *within* my Hover, I can't make separate "regions" for my Hover request. 

Let me demonstrate with pictures what I mean. Here's a nice looking Hover in VSC:

capture

Notice the way that my theme's colors define the two "regions". It's subtle, but very effective.

I'm trying to replicate that within my own Hover, but adding line breaks, header info, or any other method doesn't really pull off the same effect:

capture1

Although this isn't, obviously, bad, I think it's worse than the above image, since it's an extremely bold line, which doesn't even run the full length of its container. 

I believe, based on some experimenting, that this is really just VSCode pooling two Hover requests together. However, I cannot send two Hovers (or at least, it does not appear that I can) through my Hover handler in an LSP. 

If there is a known solution to this that I haven't been able to find, I would appreciate hearing about it. Thank you for your time!

DJMcNab:
Fixes #374 and Microsoft/language-server-protocol#518.


dbaeumer:
@DJMcNab can you elaborate why this is necessary. The reason why I am asking is that this is a breaking change for a client and we need to provide capabilities for this.

If the reason for this is the rendering in the VS Code UI (horizontal lines) I think we are better off treating `---` special and split the content on the client side into an array there.

DJMcNab:
Certainly - the reason is for that UI effect. I created this PR mostly because you labelled #374 as 'feature-request', without raising this issue, which then seemed dormant and unactioned. I also encountered this limitation in a language server I am developing, where I had to fall back on the old deprecated API, which seemed wrong. However, seeing @felixfbecker's solution as laid out in https://github.com/Microsoft/language-server-protocol/issues/518#issuecomment-437880445 makes me realise that splitting on the `---` is a good idea.

However, the usual visual style of the `---` is ugly, so perhaps it would be good to clarify that client implementations should use minimal styling for a `---` in markdown for a hover.

dbaeumer:
@DJMcNab I think the styling of `---` is very specific to VS Code and how it combines hovers from different providers. In VS Code for a language `X` there can be two hover providers which are separated using a `hline`. The API allows to separate section with a different styled hline by providing an array. IMO this shouldn't leak into the LSP API. So we should simple here https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/protocolConverter.ts#L257  convert the hover content into an array based on `---`.

DJMcNab:
So, just to clarify, I should close Microsoft/language-server-protocol#572 and instead make the change in https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/protocolConverter.ts#L257.

What would be the best approach to this - should I force push an update? Is it worth the slight jankiness of splitting based on the raw string `---`. Alternatively, the actual vscode hover markdown support replace `Horizontal Break` with the thinner line style.

DJMcNab:
There, I have added the suggested changes.

Note that Sourcegraph uses css rules to make the markdown content with horizontal seperators look the same as multiple array items: see sourcegraph/codeintellify@2ab2bf6.

@felixfbecker, do you think we should use this approach in vscode, although (I think) that would require a PR against Microsoft/vscode.

client/src/protocolConverter.ts
```diff
@@ -257,7 +263,8 @@ export function createConverter(uriConverter?: URIConverter): Converter {
 			let result: code.MarkdownString;
 			switch (value.kind) {
 				case ls.MarkupKind.Markdown:
-					return new code.MarkdownString(value.value);
+					// https://github.com/Microsoft/vscode-languageserver-node/issues/374
+					return value.value.split(/\s*---\s*/g).map(asMarkdownString);
```
felixfbecker: Is this correct? What if `---` appears in a code snippet, or in the middle of a sentence?

dbaeumer: I am currently fixing it :-) We need to consider `***` and `___` as well and ignore html comments

dbaeumer:
Actually this is getting too complicated. We would need to handle `hline` as well and ignore `---` in inline html. Doing this right would almost mean to convert the whole markdown. Will think about another solution.

DJMcNab:
OK! Well I'm glad we're thinking about a solution! Don't worry about it.

DJMcNab:
OK! Well I'm glad we're thinking about a solution! Don't worry about it.

dbaeumer:
Fixed via: https://github.com/Microsoft/vscode/issues/65094

The following code:
```
connection.onHover((textPosition): Hover => {
	// connection.tracer.log('A trace message from the server', 'Verbose content');
	return {
		contents: {
			kind: MarkupKind.Markdown,
			value: [
				'```typescript',
				'function validate(document: TextDocument): Diagnostic[]',
				'```',
				'___',
				'Some doc',
				'',
				'_@param_ `document` '
			].join('\n')
		}
	};
});
```

produces the following hover:

capture

felixfbecker:
Nice, that's the solution we chose in Sourcegraph too 👍

DJMcNab:
Awesome work! Thanks.

Should we have a note about this in the protocol document - I suppose https://github.com/Microsoft/language-server-protocol/issues/518 should track that?

Also, does tsserver currently use this or does it use the deprecated `MarkedString[]` approach.

dbaeumer:
TSServer does use the deprecated `MarkedString[]` appraoch.

I also updated https://github.com/Microsoft/language-server-protocol/issues/518

@DJMcNab what do you suggest to add to the protocol document ?

DJMcNab:
Something like:

> It is recommended that implementations display markdown seperators (such as denoted by `---` or `<hr/>`) as a minimal horizontal line, as can be seen in vscode.

Maybe in slightly better English :smiley:.

DJMcNab:
Yeah, that's fine. It might be worth noting that `---` (or `<hr />`) is the standard separator (such as between definitions and documentation or conflicting definitions), to clarify the situation for servers. Otherwise, it can just be left.

PR #420, opened 2018-10-05
Add proposed `textDocument/callHierarchy` extension

AlexTugarev:
This PR proposes a `textDocument/callHierarchy` request sent from the client to the server to request the call hierarchy for a symbol at the given text document position.

A prototype implementation for TypeScript LS is implemented in https://github.com/theia-ide/typescript-language-server/pull/98.

Related LSP issue: https://github.com/Microsoft/language-server-protocol/issues/468

### DIFF
--- a/protocol/src/protocol.callHierarchy.proposed.md
+++ b/protocol/src/protocol.callHierarchy.proposed.md
@@ -1,0 +1,127 @@
+
+#### Call Hierarchy
+
+The LSP provides retrieving the call hierachy information with the following request.
+
+_Client Capabilities_:
+
+```ts
+CallHierarchyClientCapabilities {
+    /**
+     * The text document client capabilities
+     */
+    textDocument?: {
+        /**
+         * Capabilities specific to the `textDocument/callHierarchy`
+         */
+        callHierarchy?: {
+            /**
+             * Whether implementation supports dynamic registration. If this is set to `true`
+             * the client supports the new `(TextDocumentRegistrationOptions & StaticRegistrationOptions)`
+             * return value for the corresponding server capability as well.
+             */
+            dynamicRegistration?: boolean;
+        };
+    }
+}
+```
+
+_Server Capabilities_:
+
+```ts
+CallHierarchyServerCapabilities {
+    /**
+     * The server provides Call Hierarchy support.
+     */
+    callHierarchyProvider?: boolean | (TextDocumentRegistrationOptions & StaticRegistrationOptions);
+}
+```
+
+##### Call Hierarchy Request
+
+_Request_:
+
+The `textDocument/callHierarchy` request is sent from the client to the server to request the call hierarchy for a symbol defined (or referenced) at the given text document position.
+
+Returns a collection of calls from one symbol to another.
+
+* method: ‘textDocument/callHierarchy'
+* params: `CallHierarchyParams` defined as follows:
+
+```ts
+export interface CallHierarchyParams extends TextDocumentPositionParams {
+    /**
+     * The direction of calls to provide.
+     */
+    direction: CallHierarchyDirection;
+}
+
+export namespace CallHierarchyDirection {
+    export const Incoming: 1 = 1;
+    export const Outgoing: 2 = 2;
+}
+```
+
+_Response_:
+
+The server will send a collection of `CallHierarchyCall` objects, or `null` if no callable symbol is found at the given document position.
+
+Each `CallHierarchyCall` object defines a call from one `CallHierarchySymbol` to another.
+
+* result: `CallHierarchyCall[]` | `null`
+
+```ts
+export interface CallHierarchyCall {
+
+    /**
+     * The source range of the reference. The range is a sub range of the `from` symbol range.
+     */
+    range: Range;
+
+    /**
+    * The symbol that contains the reference.
+     */
+    from: CallHierarchySymbol;
+
+    /**
+     * The symbol that is referenced.
+    */
+    to: CallHierarchySymbol;
+}
+
+export interface CallHierarchySymbol {
+
+    /**
+     * The name of the symbol targeted by the call hierarchy request.
+     */
+    name: string;
+
+    /**
+     * More detail for this symbol, e.g the signature of a function.
+     */
+    detail?: string;
+
+    /**
+     * The kind of this symbol.
+     */
+    kind: SymbolKind;
+
+    /**
+     * URI of the document containing the symbol.
+     */
+    uri: string;
+
+    /**
+     * The range enclosing this symbol not including leading/trailing whitespace but everything else
+     * like comments. This information is typically used to determine if the the clients cursor is
+     * inside the symbol to reveal in the symbol in the UI.
+     */
+    range: Range;
+
+    /**
+     * The range that should be selected and revealed when this symbol is being picked, e.g the name of a function.
+     * Must be contained by the the `range`.
+     */
+    selectionRange: Range;
+}
+```
--- a/protocol/src/protocol.callHierarchy.proposed.ts
+++ b/protocol/src/protocol.callHierarchy.proposed.ts
@@ -1,0 +1,132 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) TypeFox and others. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import { RequestType, RequestHandler } from 'vscode-jsonrpc';
+import { SymbolKind, Range } from 'vscode-languageserver-types';
+import { TextDocumentRegistrationOptions, StaticRegistrationOptions, TextDocumentPositionParams } from './protocol';
+
+export interface CallHierarchyClientCapabilities {
+    /**
+     * The text document client capabilities
+     */
+    textDocument?: {
+        /**
+         * Capabilities specific to the `textDocument/callHierarchy`
+         */
+        callHierarchy?: {
+            /**
+             * Whether implementation supports dynamic registration. If this is set to `true`
+             * the client supports the new `(TextDocumentRegistrationOptions & StaticRegistrationOptions)`
+             * return value for the corresponding server capability as well.
+             */
+            dynamicRegistration?: boolean;
+        };
+    }
+}
+
+export interface CallHierarchyServerCapabilities {
+    /**
+     * The server provides Call Hierarchy support.
+     */
+    callHierarchyProvider?: boolean | (TextDocumentRegistrationOptions & StaticRegistrationOptions);
+}
+
+/**
+ * Request to provide the call hierarchy at a given text document position.
+ *
+ * The request's parameter is of type [CallHierarchyParams](#CallHierarchyParams). The response
+ * is of type [CallHierarchyCall[]](#CallHierarchyCall) or a Thenable that resolves to such.
+ *
+ * Evaluates the symbol defined (or referenced) at the given position, and returns all incoming or outgoing calls to the symbol(s).
+ */
+export namespace CallHierarchyRequest {
+    export const type = new RequestType<CallHierarchyParams, CallHierarchyCall[], void, TextDocumentRegistrationOptions>('textDocument/callHierarchy');
+    export type HandlerSignature = RequestHandler<CallHierarchyParams, CallHierarchyCall[] | null, void>;
+}
+
+/**
+ * The parameter of a `textDocument/callHierarchy` request extends the `TextDocumentPositionParams` with the direction of calls to resolve.
+ */
+export interface CallHierarchyParams extends TextDocumentPositionParams {
+    /**
+     * The direction of calls to provide.
+     */
+    direction: CallHierarchyDirection;
+}
+
+/**
+ * The direction of a call hierarchy request.
+ */
+export namespace CallHierarchyDirection {
+    /**
+     * The callers of a symbol.
+     */
+    export const Incoming: 1 = 1;
+
+    /**
+     * The callees of a symbol.
+     */
+    export const Outgoing: 2 = 2;
+}
+
+export type CallHierarchyDirection = 1 | 2;
+
+/**
+ * The result of a `textDocument/callHierarchy` request.
+ */
+export interface CallHierarchyCall {
+
+    /**
+     * The source range of the reference. The range is a sub range of the `from` symbol range.
+     */
+    range: Range;
+
+	/**
+	 * The symbol that contains the reference.
+	 */
+	from: CallHierarchySymbol;
+
+	/**
+	 * The symbol that is referenced.
+	 */
+	to: CallHierarchySymbol;
+}
+
+export interface CallHierarchySymbol {
+
+    /**
+     * The name of the symbol targeted by the call hierarchy request.
+     */
+    name: string;
+
+    /**
+     * More detail for this symbol, e.g the signature of a function.
+     */
+    detail?: string;
+
+    /**
+     * The kind of this symbol.
+     */
+    kind: SymbolKind;
+
+    /**
+     * URI of the document containing the symbol.
+     */
+    uri: string;
+
+    /**
+     * The range enclosing this symbol not including leading/trailing whitespace but everything else
+     * like comments. This information is typically used to determine if the the clients cursor is
+     * inside the symbol to reveal in the symbol in the UI.
+     */
+    range: Range;
+
+    /**
+     * The range that should be selected and revealed when this symbol is being picked, e.g the name of a function.
+     * Must be contained by the the `range`.
+     */
+    selectionRange: Range;
+}
--- a/protocol/src/main.ts
+++ b/protocol/src/main.ts
@@ -35,7 +35,19 @@ export * from './protocol';
 
 export { FoldingRangeParams as FoldingRangeRequestParam } from './protocol'; // for backward compatibility
 
+import * as callHierarchy from './protocol.callHierarchy.proposed';
+
 export namespace Proposed {
+	export type CallHierarchyClientCapabilities = callHierarchy.CallHierarchyClientCapabilities;
+	export type CallHierarchyServerCapabilities = callHierarchy.CallHierarchyServerCapabilities;
+
+	export namespace CallHierarchyRequest {
+		export const type = callHierarchy.CallHierarchyRequest.type;
+		export type HandlerSignature = callHierarchy.CallHierarchyRequest.HandlerSignature;
+	}
+
+	export type CallHierarchyParams = callHierarchy.CallHierarchyParams;
+	export type CallHierarchyCall = callHierarchy.CallHierarchyCall;
 }
 
 export interface ProtocolConnection {


protocol/src/protocol.calls.proposed.ts
```diff
@@ -0,0 +1,111 @@
...
+    callLocation: Location;
+    /**
+     * Symbol refered to by this call. For outgoing calls this is a callee,
+     * otherwise a caller.
+     */
+    symbol: DocumentSymbol;
+    /**
+     * Location of the symbol, which can be used for a follow-up request to obtain the next level of calls.
+     */
+    symbolLocation: Location;
```
svenefftinge: Instead of adding the `symbolLocation` we should allow a `Location` instead of the `Range` within `DocumentSymbol`. As suggested by @kittaakos [here](https://github.com/Microsoft/vscode-languageserver-node/pull/346#discussion_r221659062).

protocol/src/protocol.calls.proposed.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/**
+ * Represents a directed call.
+ */
+export interface Call {
+    /**
+     * Direction of the call.
+     */
+    direction: CallDirection;
```
svenefftinge: This is redundant as it always the same as provided in the `CallsParams`.

protocol/src/protocol.calls.proposed.ts
kittaakos: Unfortunately, that is a huge breaking change in the Java implementation of the LSP. :(

types/src/main.ts
```diff
@@ -1860,6 +1860,11 @@ export class DocumentSymbol {
 	 */
 	range: Range;
 
+	/**
+	 * Uri of the document.
+	 */
+	uri?: string
```
kittaakos: See: [Added an optional `uri` property to the `DocumentSymbol`.](https://github.com/Microsoft/vscode-languageserver-node/pull/425)

dbaeumer:
A call hierarchy is IMO very similar to a type hierarchy. I can be expanded in two ways and we should be able to resolve the hierarchy lazy. So I would mimic what I have suggested for the TypeHierarchy here: https://github.com/Microsoft/vscode-languageserver-node/pull/346#issuecomment-440283344

AlexTugarev:
> A call hierarchy is IMO very similar to a type hierarchy. I can be expanded in two ways and we should be able to resolve the hierarchy lazy. So I would mimic what I have suggested for the TypeHierarchy here: #346 (comment)

@dbaeumer, thanks for the feedback! So, you suggest to have two separate request types for the lazy loading? In the beginning to send `CallHierarchyParams extends TextDocumentPositionParams` and then continue to resolve by sending `ResolveCallHierarchyItemParams`. I can update my proposal quite easily to meet this.

dbaeumer:
Yes. That is the idea.

protocol/src/protocol.calls.proposed.md
```diff
@@ -0,0 +1,110 @@
...
+     * List of calls.
+     */
+    calls: Call[];
+}
+
+export interface Call {
+    /**
+     * Direction of the call.
+     */
+    direction: CallDirection;
```
tsmaeder: I think this field is implied by the call it is a result of, no?

AlexTugarev: That's correct.

AlexTugarev:
@dbaeumer, I have updated the proposal. Please have a look!

AlexTugarev:
I have updated the PR again and added a separate `callHierarchy/resolve` request (instead of having an either type for the `textDocument/callHierarchy` request's parameter.)

protocol/src/protocol.callHierarchy.proposed.md
```diff
@@ -0,0 +1,142 @@
+
+#### Call Hierarchy
+
+The LSP provides retrieving the call hierachy information with the following request.
+
+_Client Capabilities_:
+
+```ts
+CallHierarchyClientCapabilities {
```
kittaakos: I propose changing the name to `CallHierarchyCapabilities`. Client capabilities inside the `TextDocumentClientCapabilities` do not have the `Client` segment in their name:
```
SynchronizationCapabilities
CompletionCapabilities
HoverCapabilities
SignatureHelpCapabilities
ReferencesCapabilities
DocumentHighlightCapabilities
...
```

What do you think?

AlexTugarev: Sounds good!

kittaakos: Thanks for the prompt reply! I will go with `CallHierarchyCapabilities` in LSP4J.

dbaeumer:
+1 for the separate request since it is more in line with the rest of the LSP. I have two questions so:

- why do we need `callLocation`
- can you clarify the comment in `callers` and `callees`. Why do we need two properties. Given the fact that direction can either be 'incoming' or 'outgoing' and not both.

protocol/src/protocol.callHierarchy.proposed.md
```diff
@@ -0,0 +1,142 @@
...
+
+Returns a call hierarchy item for the requested call direction.
+
+* method: ‘textDocument/callHierarchy'
+* params: `CallHierarchyParams` defined as follows:
+
+```ts
+export interface CallHierarchyParams extends TextDocumentPositionParams {
+    resolve?: number;
+    direction?: 'incoming' | 'outgoing';
```
dbaeumer: I do like strings literals over numbers but so far we used numbers for all enum like types. To keep things consistent I would like to opt to say with this unless we have a strong reason to do otherwise.

AlexTugarev: +1 for consistency. 'Going to change to enum.

AlexTugarev:
@dbaeumer, thank you for the feedback!

> why do we need `callLocation`?

given this example:

```ts
// bar.ts
import { foo } from `./foo`;
function bar() {
  foo(); // callLocation
}
function do() {
  foo(); // callLocation
}  
```

My interpretation of the `CallHierarchyItem`'s properties for the **callers** of `foo`:
* `name`, `uri`, and `range`/`symbolRange` of the callers' symbols (`function bar()` and `function do()` in **bar.ts**)
* `callLocation` in **bar.ts**

My interpretation of the `CallHierarchyItem`'s properties for the **callee** of `bar`:
* `name`, `uri`, and `range`/`symbolRange` of the callee symbol (`function foo()` in **foo.ts**)
* `callLocation` in **bar.ts**

While implementing that proposal for the `typescript-language-server` and in Theia, I realized that the locations of both the callee and the call site is necessary to reveal the expected calls in the editor.


--- 

> can you clarify the comment in callers and callees. Why do we need two properties. Given the fact that direction can either be 'incoming' or 'outgoing' and not both.

I changed it from a single `calls` property to the two properties as a reaction to the comment on the type hierarchy proposal. considering the differences of the direction kind, I'm more than happy to change to something like this:

```ts
    /**
      * List of incoming/outgoing calls to/from the callable identified by this item.
      *
      * *Note*: if undefined, this item is unresolved.
      */
     calls?: CallHierarchyItem[];
```

protocol/src/protocol.callHierarchy.proposed.md
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * List of outgoing calls.
+     *
+     * *Note*: The items is _unresolved_ if `callers` and `callees` is undefined.
+     */
+	callees?: CallHierarchyItem[];
+	/**
+     * Optional data to identify an item in a resolve request.
+     */
+    data?: any;
```
kittaakos: Maybe, we could add a `recursive?: boolean` property. `true` if the method/function represents a recursion call. Thoughts?

kittaakos: A `deprecated?: boolean` could be useful too.

AlexTugarev: on `recursive?: boolean`: 
> `true` if the method/function represents a recursion call

I recall discussing that while prototyping the call hierarchy feature in Theia, and we came to the conclusion to leave the detection of cycles to the clients. 

Detecting cycles (of random length) might be quite hard when implementing the lazy-loading strategy?

kittaakos:
> why do we need callLocation?

> given this example:

@AlexTugarev, I have another example, it's Java. Let's assume the following Java source and treat `<|>` as the cursor position:
```java
public class Foo {

  protected int <|>field = 36;
  
  public static class Bar extends Foo {
    
    public Bar() {
      field = 42;
      field = 43;
    }
    
  }

}
```

Let's call the `textDocument/callHierarchy` method with `resolve = 1` to collect the callers of the `field`. It is from the constructor of `Bar`, but the number of call locations is two (`field = 42;` and `field = 43`) instead of one. How should this fit the proposed `callLocation?: Location` API? I would have expected `callLocations?: Location[]`.

From JDT:
screen shot 2019-01-28 at 13 15 55


Thank you!

AlexTugarev:
@kittaakos, I think field can be treated same as method:

> Let's call the textDocument/callHierarchy method with resolve = 1 to collect the callers of the field. It is from the constructor of Bar, but the number of call locations is two ...

The response would contain a `CallHierarchyItem` having two caller items currently. The root item's `callLocation` is `undefined`. The contained `caller` (`calls` after the following PR update) items would contain the symbol information of the caller as well as the call location.

> I would have expected callLocations?: Location[].

Well. This could be the better way to design it! A client would not need to match identical caller symbols to show something like `(2 matches)`. Initially I had some issues with the outgoing calls, but now I don't see any. E.g. callees of `public Bar()` would be returned as one `CallHierarchyItem` having two call locations. Even if there would be calls between those lines, it shouldn't be a problem to display a proper sequence on the client side.

kittaakos:
> I think field can be treated same as method:

Yes. This was clear.

> The response would contain a `CallHierarchyItem` having two caller items currently. The root item's `callLocation` is `undefined`. The contained `caller` (`calls` after the following PR update) items would contain the symbol information of the caller as well as the call location.

OK. Thanks for the clarification.

> Well. This could be the better way to design it!

Are you planning the apply this change to your LSP proposal? Although the Java implementation is still in progress, I would like to know what will be included in the final state. (It does not imply you have to update the proposal right now :)) Thanks!

AlexTugarev:
> Are you planning the apply this change to your LSP proposal?

Yes, definitely. 'Doing it right now.

protocol/src/protocol.callHierarchy.proposed.ts
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/**
+ * The direction of a call hierarchy.
+ */
+export namespace CallHierarchyDirection {
+    /**
+     * The callers of a symbol.
+     */
+    export const Incoming: 1 = 1;
```
kittaakos: Can you please explain it to me why does it start from `1` instead of `0`? Most of the values start from `1`, but for instance, `TextDocumentSyncKind` starts with `0`.

The only pattern I have seen so far is when a value is optional on the use-site, then it can start from `0`. Otherwise, it starts with a non-zero integer, `1`. Thanks!

See: https://github.com/Microsoft/vscode-languageserver-node/pull/426/files#diff-1b6afc80d5099ec4d774052b71cc1d10R87

AlexTugarev: > Can you please explain it to me ...

I can't. I looked into the `protocol.ts` and found a couple of examples where it starts from `1`, and aligned accordingly.

kittaakos: @dbaeumer, could you please advise? Thank you!

dbaeumer: New enums should start with 1. We have some that start with 0 (old once) and I don't want to break.

dbaeumer:
Regarding `callLocation`: isn't the URI in the location redundant information. What we want to express is the following in the given example:

```
function bar() {
   foo();
   faz();
}
```

The call hierarchy item for `bar` has two outgoing calls. Instead of modeling the location inside of the items for `foo` and `faz` should we model them inside of `bar`. Something like

```
calls : {
   ranges: Range[];
   item: CallHierarchyItem
}
```

If a client needs to detect recursions how will the client do this. We would need to specify which attributes on a call hierarchy item need to be considered to check for identity. 

Assuming that call hierarchies today might not be rendered using a tree but a graph I would say it makes sense to have `caller` and `callee`. But before you change back lets think about it a little more :-). Would also make it consistent with type hierarchies which as you pointed out has both.

AlexTugarev:
@dbaeumer, thanks again for the feedback!

> Instead of modeling the location inside of the items for foo and faz should we model them inside of bar. 

👍  Should we name it `Call`?

```ts
calls: Call[];

interface Call {
   ranges: Range[];
   item: CallHierarchyItem;
}
```

> ...  I would say it makes sense to have `caller` and `callee`.

Having a `Call.direction` would be another option to support rendering as graphs.

AlexTugarev:
> Regarding callLocation: isn't the URI in the location redundant information.

Yes. The information necessary to navigate to the call location is contained in the hierarchy tree, even if it contains simple ranges.

Adding a URI to call location makes the interpretation easier when it comes to implementing the navigation from the call hierarchy to editor. Here is another very reduced example, just to point out to which URIs those ranges would be then related in a comparison of callers and callees:

```jsonc

// callees of `bar()`

{
    name: bar, uri: bar.ts
    calls: [
        {
            ranges: [{ "range on l.2" }] // refers to bar.ts (container item)
            item: {name: foo, uri: foo.ts}
        },
        {
            ranges: [{ "range on l.3" }] // refers to bar.ts (container item)
            item: {name: faz, uri: foz.ts}
        },
    ]
}

// callers of `foo()`

{
    name: foo, uri: foo.ts
    calls: [
        {
            ranges: [{ "range on l.2" }] // refers to bar.ts (contained item)
            item: {name: bar, uri: bar.ts}
        }
    ]
}
```

This way it requires to read the URI either from the container or from the containing item depending on the direction.

protocol/src/protocol.callHierarchy.proposed.md
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * The kind of this symbol.
+     */
+    kind: SymbolKind;
+
+    /**
+     * `true` if the hierarchy item is deprecated. Otherwise, `false`. It is `false` by default.
+     */
+    deprecated?: boolean;
```
jrieken: Not sure that `deprecated` fits well into the call hierarchy domain. I understand the desire to represent if a symbol is deprecated but that's also true for completions or outline.  I think deprecated-knowledge deserves it's in own, independent API

AlexTugarev4: +1 for removig it.

AlexTugarev: +1 for removing.

AlexTugarev:
@jrieken, awesome so see that you're looking into this. 

Please note that there is a remark in https://github.com/Microsoft/vscode-languageserver-node/pull/420#issuecomment-459755896 I'd like to address as described.

And in https://github.com/Microsoft/vscode-languageserver-node/pull/420#issuecomment-459765019 I'd like to convince you of attaching the URI to the call location.

jrieken:
> And in #420 (comment) I'd like to convince you of attaching the URI to the call location.

Yeah, I am all in for explicit information but my focus is also the vscode API which is less about the transport of data.

protocol/src/protocol.callHierarchy.proposed.ts
```diff
@@ -0,0 +1,172 @@
...
+     * *Note*: undefined in root item.
+     */
+    callLocations?: Location[];
+
+    /**
+     * List of incoming (or outgoing) calls to (or from) a callable identified by this item.
+     *
+     * *Note*: if undefined, this item is unresolved.
+     */
+    calls?: CallHierarchyItem[];
```
jrieken: What is the relationship between `calls` and `callLocations`? Does one location correspond to one call? E.g could this be though of as array of tuples, like `calls: [Location, CallHierarchyItem][]`?

AlexTugarev: > E.g could this be though of as array of tuples, like calls: [Location, CallHierarchyItem][]

We even [discussed](https://github.com/Microsoft/vscode-languageserver-node/pull/420#issuecomment-458109531), that this should be `[Location[], CallHierarchyItem][]` to squash multiple calls to/from the the same callable item.


> Yeah, I am all in for explicit information but my focus is also the vscode API which is less about the transport of data.

My point _is_ about the API: if you separate URI from call locations, you'd need to read the URI from different positions in the hierarchy just to implement the reveal. WDYT?

protocol/src/protocol.callHierarchy.proposed.md
dbaeumer: Actually the LSP supports deprectated on `CompletionItem` and `DocumentSymbol`. @jrieken how would a independent API look like in your opinion. When LSP introduced it it seemed to be the most logical to add it to the items. The only thing I could think of having it as a tag now that we started to introduce this for diagnostics. But it would still go onto the items. Or?

dbaeumer:
#Under construnction

@jrieken thanks a lot for looking into this for the VS Code API. This is a big help. I would like to clarify some of the suggestions I made so far. I will use an example to better describe this:

```ts
function foo() {
   bar();
   baz();
   bar();
}

function faz() {
   foo();
   raz();
}

function fuz() {
   raz();
   foo();
}
```

For the call hierarchy item for `foo` we need to model the following:

- it calls `bar` 2 times at different ranges
- it calls `baz` one time
- it is called by `faz` and `fuz`

In the model so far I tried to avoid introducing call hierarchy item identity. So a call hierarchy item representing the same function can appear more than once in a result. However I tried to avoid it for multiple calls from the same function. But I am open to change this.

For the `calls` relation ship things are releatively easy. The call hiarachy item for `foo` looks like this (CHI === calll hierachy item)

```
{
   uri: '....'
   range: '....'
   calls: [
      { item: 'CHI for bar', ranges: [{r1, r3}] },
      { item: 'CHI for baz', ranges [{r2}] }
   ]
}
```

Only returniung one call hierarchy item for bar has the advantage that the client doesn't need to dedupe them for a single call. Alternatively we can return:

```
{
   uri: '....'
   range: '....'
   calls: [
      { item: 'CHI for bar', ranges: {r1} ] },
      { item: 'CHI for baz', ranges {r2} }
      { item: 'CHI for bar', ranges: {r3} },
   ]
}
```

AlexTugarev:
> And fill for faz and fuz the calls relationship with one level. A client would then need to map the CHI item for foo to the one returned earlier using the CHI's uri and range property.

@dbaeumer, does this mean that the server has to return *all* `calls` for `faz` and `fuz` at once? or should it just return the items necessary to create the mapping mentioned above?

My concern is that is's could be inconvenient to implement for both sides, the client and the server. Can you explain which part of the story has to be as easy as possible, please? 

You mentioned above that a uri would be redundant. It would be something like `{ item: CHI, locations: Location }[]`. Could it be that the last example is quite comparable, e.g. `chi1.uri == chi1.calledBy[0].calls[0].uri`,  or did I get it wrong?

protocol/src/protocol.callHierarchy.proposed.md
jrieken: yeah, never thought much about it but I think I would have done something similar to diagnostics or decorations. However, if LSP already gone down this road for completions and outline that I think its fair to continue like that

jrieken:
Looking at this through my vscode-api-glasses I would probably approach this slightly different. I am wondering if this can be similar to reference search, for instance:

```ts
enum CallHierarchyDirection {
    CallsFrom,
    CallsTo
}

interface CallHierarchyProvider {
    provideCallHierarchy(
      document: TextDocument,
      position: Position, 
      direction: CallHierarchyDirection, 
      token: CancellationToken
   ): ProviderResult<Location[]>;
}
```

This biggest differences with this approach are

* The request location (`document` and `position`) isn't resolved but assumed to be OK. Not sure about the LSP-land but this is how it is vscode-land, e.g. with reference search, finding implementors, and others. So, instead of saying "resolve and return a hierarchy item at the position" it is "assuming there is symbol at the position, find its calls/callers".
* The other difference is that there is no support for deep resolving, e.g. for knowing the callers of a caller you would need to invoke `provideCallHierarchy` again. Not sure how language brains implement call hierarchies but I don't think that "callers of callers" or "calls of calls" are the byproduct of computing one level. 

Wrt to the request location we have been pragmatic, e.g providers return no results or an error when a location doesn't make sense (invoke go to definition inside a comment). For rename, which benefits from request location validation, we have added [`prepareRename`](https://github.com/Microsoft/vscode/blob/471b98c10208c1329a67728e656e48a7ce7bac18/src/vs/vscode.d.ts#L2989) so that provider have a chance to validate/resolve a location before the editor shows the input-UI.

dbaeumer:
For me the biggest difference with that approach is that it needs additional resolving to come from a location to something that is a document symbol or at least the text. A call hierarchy in its current proposal is more like document symbols and not so much like find references. So the resulting graph can be rendered without having any document open (minus the start position).

jrieken:
Yeah, as said this idea mostly driven by how reference search is spec'ed. I didn't spend enough time on how the UX should look and don't know if something that worked for references will or won't work for call hierarchies... E.g for references we show files, resolve them when you expand one, and show matches then. This approach might not work for call hierarchies as things aren't grouped by file

jrieken:
Ok, after looking further at the UX that's established in this area I does make sense to model this closer to document symbols. Containers are symbols and not files, so it makes sense to have them prominently in the API. 

---

In my playground I have extracted something like this but I also wonder if it make sense to have `callsFrom` and `callsTo` on the hierarchy item, e.g. having vertices and edges in one model object, or to leave edges "outside". 

```ts
class CallHierarchyItem {
    kind: vscode.SymbolKind;
    name: string;
    detail: string;
    uri: vscode.Uri // really needed?
    range: vscode.Range;
    selectionRange: vscode.Range;

    // unsure about those
    callsFrom?
    callsTo?
}
```

A vertices-only approach would have a resolve-function that represents edges, return resulting vertices. The sample below follows @dbaeumer approach of grouping locations by item - using `Location` or maybe `Range` with `Uri` on the item itself. 

```ts
resolveCallHierarchyItem(
      item: CallHierarchyItem,
      direction: CallHierarchyDirection,
      token: vscode.CancellationToken
): vscode.ProviderResult<[CallHierarchyItem, vscode.Location[]][]>; // Range or Location, also depends on CallHierarchyItem#uri
```

I think decoupling vertices and edges makes the object model simpler and prevents pointer-spaghetti, e.g when resolving a 'callsFrom'-item should its 'callsTo'-property point back? 

Somewhat different from other providers is that to get started a call to get a `CallHierarchyItem` is needed, e.g. the `provideCallHierarchyItem` would only resolve a location to a hierarchy item but that by itself isn't of much use and another resolve-call will follow. However, that's only needed to get started. 

A `CallHierarchyItem` is like a document symbol, describing a symbol. The beefy part is in `[CallHierarchyItem, vscode.Location[]][]` which is returned from the resolve-function. It's the edge and its interpretation depends on the direction. The location (or range) is always relative to the "caller" with the caller either being the provided item or the returned item. 


Staying in @dbaeumer's sample (added line numbers):

```
 1: function foo() {
 2:    bar();
 3:    baz();
 4:    bar();
 5: }
 6: 
 7: function faz() {
 8:    foo();
 9:    raz();
10: }
11: 
12: function fuz() {
13:    raz();
14:    foo();
15: }
```

* invoke `provideCallHierarchyItem` on line 1 returns everything UX-relevant for function `foo`
* invoke `resolveCallHierarchyItem` with `callsFrom` and the `foo`-item returns `[[bar-item, [2, 4]], [baz-item, [3]]`, e.g. the lines 2,4 and 3 are relative to the caller `foo`
* invoke `resolveCallHierarchyItem` with `callsTo` and the `foo`-item returns `[[faz-item, [8]], [fuz-item, [14]]]`, e.g. line 8 is relative to the caller `faz` and line 14 is relative to the caller `fuz`

AlexTugarev:
> class CallHierarchyItem { ...
> uri: vscode.Uri // really needed?
> .. }

At least the initial item has to have it in order to reveal in the editor. Also if you think of alternative representations than trees, e.g. graphs, you might also need this per vertex.

> Somewhat different from other providers is that to get started a call to get a CallHierarchyItem is needed, e.g. the provideCallHierarchyItem would only resolve a location to a hierarchy item but that by itself isn't of much use and another resolve-call will follow. However, that's only needed to get started.

Actually, it's quite handy to start with a [pre-resolved number of items](https://github.com/Microsoft/vscode-languageserver-node/pull/420/files#diff-2c8fbc195a890a3e9c35db45c512b416R69).

> A CallHierarchyItem is like a document symbol, describing a symbol. The beefy part is in [CallHierarchyItem, vscode.Location[]][] which is returned from the resolve-function. It's the edge and its interpretation depends on the direction. The location (or range) is always relative to the "caller" with the caller either being the provided item or the returned item.

Agree. Just a note to emphasize this: if the call is modeled as a `Location` (with the URI) here, there is no need to interpret it in relation to the provided/returned item and the direction.

dbaeumer:
@jrieken 👍 for the idea to separate the items from the edges. We do the same by the way for LSIF and it makes outputing a graph so much easier. We should do the same then for type hierarchies to keep things consistent.

aeschli:
I discussed with @jrieken and we came up with a new simplified proposal:
https://github.com/Microsoft/vscode/issues/70231#issuecomment-477042744

```ts
interface CallHierarchyItemProvider {
    /** 
     * Evaluates the symbols(s) defined (or referenced) at given position, and 
     * returns all incoming or outgoing calls to the symbol(s).
     */
    provideCalls(doc: TextDocument, pos: Position | Range, dir: CallHierarchyDirection) : Promise<CallHierarchyCall[]>;

}

enum CallHierarchyDirection {
    CallsFrom,
    CallsTo,
}

class CallHierarchyCall {
   /** The source range of the reference. The range is a sub range of the 'from' symbol range */
   range: Range;

   /** The symbol that contains the reference */
   from: CallHierarchySymbol;

   /** The symbol that is referenced */
   to: CallHierarchySymbol; 
}

class CallHierarchySymbol {
    kind: SymbolKind;
    name: string;
    detail?: string;
    uri: Uri;
    range: Range;
    selectionRange: Range;
}
```

The proposal avoids the [Locaction[], Symbol] tuples and accepts that some symbols will be repeated.
It also handles the case that for a document offset there can be more than one definition (as with the definition provider)

AlexTugarev:
I'm a bit puzzled about the change of direction ;-) 

@aeschli, I like that proposal, which is quite similar to what this PR was about in the beginning. 

@dbaeumer, you stated some pros for the [Locaction[], Symbol] tuples. How to proceed now? I'd really like to update this asap.

tsmaeder:
@aeschli could you explain why a "Range" is necessary in the "pos" parameter of "provideCalls"?

dbaeumer:
@AlexTugarev the big change that happend (see comment https://github.com/Microsoft/vscode-languageserver-node/pull/420#issuecomment-472805971 and https://github.com/Microsoft/vscode-languageserver-node/pull/420#issuecomment-473216560) is that we moved from a vertex based result to an edge based result. The cost is that the result contains duplicate information (e.g. the `[Locaction[], Symbol]` discussion). This is fine for me since the edge based model brings quite some advantages.

AlexTugarev:
@dbaeumer, thanks! This sounds good. I'll update this proposal shortly.

aeschli:
@tsmaeder I allowed `Range` in `provideCalls` so that you can pass in a user selection or the range of a `CallHierarchySymbol`.

AlexTugarev:
> pass in a user selection or the range of a CallHierarchySymbol

@aeschli, which part would be responsible to check if there is only a single reference (call) or a definition in that `pos` Range?
and what is the downside of using `selectionRange.start`, besides that it might break when using a user selection?

aeschli:
@AlexTugarev When the symbol at the position (or range) resolves in multiple definitions, the provideCalls result will contain calls to/from *all* of the definitions. The UI can decide to group the results under multiple root nodes (one for each definition).

Position is probably also good enough. The definition provider and the reference provider also just take a position. I added it for convenience, and future considerations: There might be cases where a range would allow the user to be more specific on what to look for. E.g. in a reference like List<String> one could say that selecting `List` would resolve to the non-parameterized List, while ` List<String>` the list of strings.
But given that the definition provider and the reference provider don't support such capabilities, we are probably good to go with just `Position`.

dbaeumer:
@AlexTugarev we plan to ship a new stable version of VS Code end of this week. So if you have something that corresponds to the new API I could create a new next version with the call hierarchy support as proposed so that people can try it out.

AlexTugarev:
@dbaeumer, that sounds good. I'll update it this PR tonight.

dbaeumer:
@AlexTugarev ping me when the changes are available

AlexTugarev:
@dbaeumer, please have a look.

I'm not sure about the `client`/`server` part regarding the proposed API. I cannot find your example how to create it.

dbaeumer:
On the server part we usually register a request handler as long as the protocol is proposed.

I will merge the PR in and will then prepare a new next version with the selection range proposal.

AlexTugarev:
Thanks @dbaeumer!

MaskRay:
I think the proposal has several deficiency that require further improvement. Comments inlined below:

```typescript
export interface CallHierarchyCall {
  // If there are multiple `CallHierarchyCall` elements with the same `from`, will the `from: CallHierarchySymbol` be duplicated multiple times?
  range: Range;
  from: CallHierarchySymbol;
  to: CallHierarchySymbol;
}
result: CallHierarchyCall[] | null

export interface CallHierarchySymbol {
  // There is no id field. `name` or `detail` may not identify a specific target that can be suitable for further requests (e.g. node expansion).
  name: string;
  detail?: string;
  kind: SymbolKind;
  uri: string;
  range: Range;
  selectionRange: Range;
}
```

In ccls (https://github.com/MaskRay/ccls/blob/master/src/messages/ccls_call.cc), we use:

```cpp
struct Out_cclsCall {
  Usr usr;  // language server specific
  std::string id;
  std::string_view name;
  Location location;
  CallType callType = CallType::Direct;
  int numChildren;
  // Empty if the |levels| limit is reached.
  std::vector<Out_cclsCall> children;
};
```

The method `$ccls/call` was designed so that the result can either be a hierarchy or a flattened list (like `textDocument/references`).

dbaeumer:
@MaskRay can you please open a GitHub issue to discuss this.

HighCommander4:
Does this being merged mean the extension is now an official part of LSP? Or are there additional steps required for that to happen?

rcjsuen:
@HighCommander4 I wouldn't consider anything final until microsoft/language-server-protocol#468 has been closed. I think people look to the [website](https://microsoft.github.io/language-server-protocol/specification) for the final say in general.

HighCommander4:
> I wouldn't consider anything final until [microsoft/language-server-protocol#468](https://github.com/microsoft/language-server-protocol/issues/468) has been closed.

Thanks. Do you know what is needed for that to happen? Is it just a matter of someone writing a PR?

And another question: is the proposed spec (the `protocol.callHierarchy.proposed.md` file from the commit in this PR) hosted in rendered form somewhere, for implementers / early adopters to look at?

rcjsuen:
@HighCommander4

1. From what I understand, yes, it's just a matter of someone writing up the PR. But note that at the end of the day it is Microsoft's call regarding when to merge it and make it an official new LSP release (be it 3.x or 4.0). Note that I am not expecting the next release to be 4.0 but that's just me and I don't speak for anyone else in the LSP community.
1. You can find that file [rendered online in GitHub](https://github.com/microsoft/vscode-languageserver-node/blob/4cee63ff54eb16587e9015f80a94b7873123ca18/protocol/src/protocol.callHierarchy.proposed.md) but I do not believe it is viewable through any "non-GitHub website" otherwise.

KamasamaK:
As a consequence of it not being finalized yet, it's important to understand that it's still subject to change. There continue to be discussions on how best to finalize this API, but since this PR is merged this is probably not the best place. Maybe create a new issue.

Also, while there is nothing to actually dictate this, it is generally the case that the LSP API is closely aligned with the VS Code API, so you can see microsoft/vscode#70231 for how the API may change as well.

PR #425, opened 2018-10-15
Added an optional `uri` property to the `DocumentSymbol`.

kittaakos:
Clients can use this when the URI of the text document
cannot be inferred from the request context.

Issue: https://github.com/Microsoft/language-server-protocol/issues/582

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -365,7 +365,7 @@
 		},
 		"event-stream": {
 			"version": "3.3.4",
-			"resolved": "http://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
+			"resolved": "https://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
 			"integrity": "sha1-SrTJoPWlTbkzi0w02Gv86PSzVXE=",
 			"dev": true,
 			"requires": {
--- a/types/src/main.ts
+++ b/types/src/main.ts
@@ -1870,6 +1870,16 @@ export class DocumentSymbol {
 	 * Children of this symbol, e.g. properties of a class.
 	 */
 	children?: DocumentSymbol[];
+
+	/**
+	 * The URI of the text document this symbol belongs to. Can be set by the language server,
+	 * if the URI of the text document is not available or cannot be inferred from the request context.
+	 *
+	 * If not defined, it can be inferred from the context of the request. For example, when calling the
+	 * `textDocument/documentSymbol` method, the `DocumentUri` (`string`) can be inferred from the request
+	 * parameter: `DocumentSymbolParams.textDocument.uri`.
+	 */
+	uri?: string;
 }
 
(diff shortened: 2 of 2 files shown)


dbaeumer:
I understand why you are requesting this (because of the type hierarchy request) but IMO this is not the right way to go. DocumentSymbols are as the name implies scoped to a document and we shouldn't start changing this. I would rather opt to have a new type `TypeHierarchyItem`. This ensures we can evolve them independently. In addition we can make these items lazy to better support large type hierarchies.

Any objection to close this ?

kittaakos:
> because of the type hierarchy request

Yes :)

> I would rather opt to have a new type `TypeHierarchyItem`.
> This ensures we can evolve them independently.

This was the initial idea.

> Any objection to close this ?

I do not mind adopting the proposals. What's your take on that, @svenefftinge?

yyoncho:
>  I would rather opt to have a new type TypeHierarchyItem. 

IMO the name `TypeHierarchyItem` is not good since the class could be usefull for cases not related to type hierarchy. E. g. in Eclipse you could list the class symbols + inherited members.

kittaakos:
> IMO the name `TypeHierarchyItem` is not good since the class could be usefull for cases not related to type hierarchy. E. g. in Eclipse you could list the class symbols + inherited members.

It should be still possible with the `TypeHierarchyItem`.

Presumably, you are talking about the <kbd>Ctrl/Cmd</kbd>+<kbd>o</kbd> functionality in Eclipse. It is your outline: `textDocument/documentSymbol`.

If you want the inherited members (<kbd>Ctrl/Cmd</kbd>+<kbd>o</kbd> and <kbd>Ctrl/Cmd</kbd>+<kbd>o</kbd> again), you can get the supertypes of the current document: `textDocument/superTypes`, and execute `textDocument/documentSymbol` for each type.  Finally, you merge the result on the client. Although, I did not check with JDT, I would assume, Eclipse does it in two steps too. @yyoncho, correct me if I've overlooked something.

svenefftinge:
> If you want the inherited members (Ctrl/Cmd+o and Ctrl/Cmd+o again), you can get the supertypes of the current document: textDocument/superTypes, and execute textDocument/documentSymbol for each type. Finally, you merge the result on the client. Although, I did not check with JDT, I would assume, Eclipse does it in two steps too. @yyoncho, correct me if I've overlooked something.

Deciding on whether a symbol is inherited or overridden requires knowledge of language semantics, so I'd say this is something to be provided by the language server and not the client.

dbaeumer:
The LSP actually tries hard to stay way from this. All the types it declares are bound to a tool action and usually return UI data types. The `documentSymbols` request would have better be named `documentOutline` and returned `OutlineItems`. 

For most servers that request doesn't return all symbols defined in a document (for example locals are usually not part of this).

I will close the issue. Please ping if someone disagrees.

PR #426, opened 2018-10-15
Proposal of the type hierarchy protocol extension.

kittaakos:
**Note**:
Another related PR in this repository: https://github.com/Microsoft/vscode-languageserver-node/pull/346
The original GH issue for the PR: https://github.com/Microsoft/language-server-protocol/issues/136

----

Type hierarchy in action with [Eclipse Theia](https://github.com/theia-ide/theia/pull/3148) and the [JDT language server](https://github.com/eclipse/eclipse.jdt.ls/pull/823).

46807217-b4993600-cd69-11e8-8fd6-cdd5771004f2

CC: @svenefftinge

### DIFF
--- a/protocol/src/protocol.typeHierarchy.proposed.md
+++ b/protocol/src/protocol.typeHierarchy.proposed.md
@@ -1,0 +1,222 @@
+#### Type Hierarchy
+
+The type hierarchy language feature provides type hierarchy information for languages supporting inheritance and nominal typing.
+
+_Client Capabilities_:
+
+```ts
+/**
+ * Client capabilities specific to the type hierarchy feature.
+ */
+export interface TypeHierarchyCapabilities {
+
+  /**
+   * The text document client capabilities.
+   */
+  textDocument?: {
+
+    /**
+     * Capabilities specific to the `textDocument/typeHierarchy`.
+     */
+    typeHierarchy?: {
+
+      /**
+       * Whether implementation supports dynamic registration. If this is set to `true`
+       * the client supports the new `(TextDocumentRegistrationOptions & StaticRegistrationOptions)`
+       * return value for the corresponding server capability as well.
+       */
+      dynamicRegistration?: boolean;
+
+    }
+
+  }
+
+}
+```
+
+_Server Capabilities_:
+
+Capability whether the language server capability can provider type hierarchy information to the client.
+
+```ts
+/**
+ * Type hierarchy language server capability.
+ */
+export interface TypeHierarchyServerCapabilities {
+
+  /**
+   * Server capability for calculating super- and subtype hierarchies.
+   */
+  typeHierarchyProvider?: boolean | (TextDocumentRegistrationOptions & StaticRegistrationOptions);
+
+}
+```
+
+_Requests_:
+
+The `Type Hierarchy Request` is sent from the client to the server to retrieve a `TypeHierarchyItem` at a given cursor position. If computing full type hierarchy items is expensive, servers can additionally provide a handler for the type hierarchy item resolve request (`typeHierarchy/resolve`). This request would also allow to specify if the item should be resolved and whether sub- or supertypes are to be resolved. If no item can be retrieved for a given text document position, returns with `null`.
+
+#### Type Hierarchy Request:
+
+ * method: `textDocument/typeHierarchy`
+ * params: `TypeHierarchyParams`
+
+```ts
+/**
+ * The type hierarchy params is an extension of the `TextDocumentPositionParams` with optional properties
+ * which can be used to eagerly resolve the item when requesting from the server.
+ */
+export interface TypeHierarchyParams extends TextDocumentPositionParams {
+
+  /**
+   * The hierarchy levels to resolve. `0` indicates no level. When not defined, it is treated as `0`.
+   */
+  resolve?: number;
+
+  /**
+   * The direction of the hierarchy levels to resolve.
+   */
+  direction?: TypeHierarchyDirection
+
+}
+
+export namespace TypeHierarchyDirection {
+
+  /**
+   * Flag for retrieving/resolving the subtypes.
+   */
+  export const Children = 0;
+
+  /**
+   * Flag to use when retrieving/resolving the supertypes.
+   */
+  export const Parents = 1;
+
+  /**
+   * Flag for resolving both the super- and subtypes.
+   */
+  export const Both = 2;
+
+}
+export type TypeHierarchyDirection = 0 | 1 | 2;
+
+/**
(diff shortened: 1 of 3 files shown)


kittaakos:
I have updated the proposal based on https://github.com/Microsoft/vscode-languageserver-node/pull/346#issuecomment-440283344.

protocol/src/protocol.typeHierarchy.proposed.md
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * The hierarchy levels to resolve. `0` indicates no level.
+   */
+  resolve?: number;
+
+  /**
+   * The direction of the hierarchy levels to resolve. Valid values are: `parents`, `children`, or `both`.
+   * Defaults to `children`.
+   */
+  direction?: 'parents' | 'children' | 'both';
```
dbaeumer: I think we need to clarify what should happen if a hierarchy item has already the parents resolved and is asked to resolve its children. Should the client send the parents again or is it allowed to leave them out. I would opt for the second.

dbaeumer: I do like strings literals over numbers but so far we used numbers for all enum like types. To keep things consistent I would like to opt to say with this unless we have a strong reason to do otherwise.

dbaeumer: I know I proposed this earlier but now I am more for consistency :-)

kittaakos: > I would opt for the second.

+1

@dbaeumer, I would have a follow-up question on this: what should happen in general, if something is already resolved and the client requests to resolve it again?

In the JDT reference implementation, I did for this language feature, I blindly resolve everything from scratch, because:
 - the text document that is referenced by the `TypeHierarchyItem` might have changed meanwhile, so the item I am trying to resolve is not at the same cursor position anymore,
 - the type hierarchy of my `TypeHierarchyItem` might have changed.

Thank you!

protocol/src/protocol.typeHierarchy.proposed.md
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * The hierarchy levels to resolve. `0` indicates no level. When not defined, it is treated as `0`.
+   */
+  resolve?: number;
+
+  /**
+   * The direction of the hierarchy levels to resolve.
+   */
+  direction?: TypeHierarchyDirection
```
HighCommander4: Does this have a default? What should happen if the client provides a value for `resolve` but not for `direction`?

protocol/src/protocol.typeHierarchy.proposed.md
```diff
@@ -0,0 +1,222 @@
...
+   * Server capability for calculating super- and subtype hierarchies.
+   */
+  typeHierarchyProvider?: boolean | (TextDocumentRegistrationOptions & StaticRegistrationOptions);
+
+}
+```
+
+_Requests_:
+
+The `Type Hierarchy Request` is sent from the client to the server to retrieve a `TypeHierarchyItem` at a given cursor position. If computing full type hierarchy items is expensive, servers can additionally provide a handler for the type hierarchy item resolve request (`typeHierarchy/resolve`). This request would also allow to specify if the item should be resolved and whether sub- or supertypes are to be resolved. If no item can be retrieved for a given text document position, returns with `null`.
```
sam-mccall: This seems pretty complicated - is there evidence that a) lazy resolving is needed and b) eager resolving is *also* needed?

protocol/src/protocol.typeHierarchy.proposed.md
```diff
@@ -0,0 +1,222 @@
...
+/**
+ * The type hierarchy params is an extension of the `TextDocumentPositionParams` with optional properties
+ * which can be used to eagerly resolve the item when requesting from the server.
+ */
+export interface TypeHierarchyParams extends TextDocumentPositionParams {
+
+  /**
+   * The hierarchy levels to resolve. `0` indicates no level. When not defined, it is treated as `0`.
+   */
+  resolve?: number;
```
sam-mccall: can we specify that the server may resolve more than this number of levels, e.g. if it's cheap to do so?

Client doesn't know whether it's cheap or expensive, so may specify low resolve (e.g. 1). However for some servers (in particular for clangd), resolving parent links one-at-a-time in a loop is *much* more expensive than doing it all at once - the cost is in finding the first parent, the rest are free.

HighCommander4:
Note that there is now a server implementation of this in [clangd](https://clang.llvm.org/extra/clangd/Features.html#complete-list-of-features) (C++ language server), using the protocol proposed here.

yudar1024:
@HighCommander4 I wanna know weather this spec is finalized? wait this feature long time and an implement pr merge [823](https://github.com/eclipse/eclipse.jdt.ls/pull/823) depends on this spec

HighCommander4:
@yudar1024 I'm not the spec's author, nor a maintainer of this repo. I just implemented the spec in the clangd language server.

My understanding is the spec continues to be subject to change until it merges. However, clients and servers may choose to implement it eagerly (as we did in clangd), with the understanding that if the spec changes the implementation will need to be updated as well.

yudar1024:
@fbricon can you merge first to let developer can use it in early time like clangd,  merge it at this time or after spec be finalize, we both need change the code if the spec make some change. but if we merge at this time , at least it can make vsc more efficient early

HighCommander4:
@kittaakos What should I make of this PR being closed? It does not appear to have been merged. Was there a decision to reject it somewhere?

Trias:
I would like you to consider making ´range` optional in the request. This could be used for displaying document level type requests (i. e inheritance chain of the current class).

I would also like to know the status of this proposal, as it seemed to be rather popular.

dbaeumer:
@Trias there is currently no active work happening to support type hierarchies in LSP.

HighCommander4:
> @Trias there is currently no active work happening to support type hierarchies in LSP.

To clarify: is that just because there's no one driving it forward at the moment? Or is it because the LSP maintainers feel there is no value in this feature?

That is, if someone were to dust off this PR, rebase it to be current, and address any outstanding feedback -- would you / other maintainers continue to entertain it?

dbaeumer:
Such a request makes sense in LSP. But please keep in mind that beside having a spec part we need an implementation in a client library and client as well.

Trias:
client meaning vscode? There apparently is already an existing implementation as shown in the first post but in the theia IDE.

HighCommander4:
We also have a vscode implementation in [vscode-clangd](https://github.com/clangd/vscode-clangd/blob/11f5b680647de3c195997515ecc85d0983fe4395/src/type-hierarchy.ts).

PR #431, opened 2018-10-20
Update version in lockfile to match package.json

thibaudcolas:
Spotted while running `npm install` in local development. npm automatically syncs the `package.json` and the lockfile, which creates this change.

The 5.1.1 version bump is from 9c26d2b58a441698a8e7790ce46c667dc4e2e199.


dbaeumer:
Thanks!

PR #432, opened 2018-10-21
Add support for url attribute of Diagnostic API

thibaudcolas:
See https://github.com/Microsoft/vscode/pull/61436. This adds support for the new `url` attribute of Diagnostic objects in the converters, so they can be forwarded to extensions using language servers like `vscode-eslint` (https://github.com/Microsoft/vscode-eslint/pull/562).

Edit: as far as I can see, the CI failure is because the Diagnostic `url` doesn't exist _yet_.


thibaudcolas:
Closing because of https://github.com/Microsoft/vscode/pull/61436#issuecomment-437942788.

PR #444, opened 2018-12-26
Support using a seperate output channel for trace messages

DJMcNab:
This allows setting `traceOutputChannelName` in the `LanguageClientOptions` to use a different output channel. Otherwise, this falls back onto the default output channel. `traceOutputChannel` can also be set directly.

Just a note: The current formatting settings are not actually consistently applied on all of the files, so I have been caught out by format on save whilst creating this PR.

TODO: test this - I couldn't get `npm` to work properly (it was giving me really obtuse error messages) when I tried to install my local copy of the client into an extension. (@matklad, how did you do it?)

TODO: determine if the output channel should be automatically shown. My inclination is to say not, but I'm not sure.

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -452,6 +452,8 @@ export interface LanguageClientOptions {
 	diagnosticCollectionName?: string;
 	outputChannel?: OutputChannel;
 	outputChannelName?: string;
+	traceOutputChannel?: OutputChannel;
+	traceOutputChannelName?: string;
 	revealOutputChannelOn?: RevealOutputChannelOn;
 	/**
 	 * The encoding use to read stdout and stderr. Defaults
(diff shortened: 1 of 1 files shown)


dbaeumer:
@DJMcNab thanks for the PR. I would like to request one change. Instead of having both a name and a channel we should support having a channel only. The reason why we have this for the normal output channel is API compatibility. In a first version you could only specify a name and not a concrete channel.

client/src/client.ts
```diff
@@ -2367,6 +2368,8 @@ export abstract class BaseLanguageClient {
 	private _initializeResult: InitializeResult | undefined;
 	private _outputChannel: OutputChannel | undefined;
 	private _disposeOutputChannel: boolean;
+	private _traceOutputChannel: OutputChannel | undefined;
+	private _disposeTraceOutputChannel: boolean;
```
dbaeumer: I think we don't need this boolean. Since the creator of the client passes in a channel in the constructor the client never owns the channel and  therefore never disposes it. The channel should be disposed by the extension itself since it created it.

PR #446, opened 2019-01-02
Fix configuration not always using resource scope

LaurentTreguier:
When requesting a nested configuration section (e.g. `a.b.c`), the configuration returned isn't scoped to the requested uri, although it does so when requesting something like `a`, with no dot inside.


dbaeumer:
Good catch. Thanks !

PR #459, opened 2019-02-11
Add generic capabilities type

johnwiseheart:
We have a variety of experimental capabilities that we add using the `ServerCapabilities` type, but are unable to type them nicely because `experimental` is typed as `any`. By adding this generic, it would enable people to reuse this type, without having to hack around the `any`, without changing how any back-compatibility of existing code works (I think).

Thoughts?

### DIFF
--- a/protocol/src/protocol.ts
+++ b/protocol/src/protocol.ts
@@ -869,7 +869,7 @@ export interface TextDocumentSyncOptions {
  * Defines the capabilities provided by a language
  * server.
  */
-export interface _ServerCapabilities {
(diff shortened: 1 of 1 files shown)


dbaeumer:
Actually that makes sense now that TS support default values for generics.

johnwiseheart:
Hey @dbaeumer would you mind publishing a new version to npm? Would like to use this in https://www.npmjs.com/package/vscode-base-languageclient

dbaeumer:
I released new verion as next. However `vscode-base-languageclient` is not maintained by us.

PR #474, opened 2019-03-27
adjust selectionRange API

matklad:
Tested manally with insiders code and [this commit](https://github.com/rust-analyzer/rust-analyzer/commit/ac095ffd93fd6ae544af46c02d8034fe8baef67d) in rust-analyzer. Seems to work :)

### DIFF
--- a/client/src/selectionRange.proposed.ts
+++ b/client/src/selectionRange.proposed.ts
@@ -7,11 +7,11 @@
 import * as UUID from './utils/uuid';
 import * as Is from './utils/is';
 
-import { languages as Languages, Disposable, TextDocument, ProviderResult, Position as VPosition, SelectionRange as VSelectionRange, SelectionRangeKind as VSelectionRangeKind } from 'vscode';
+import { languages as Languages, Disposable, TextDocument, ProviderResult, Position as VPosition, SelectionRange as VSelectionRange } from 'vscode';
 
 import {
 	ClientCapabilities, CancellationToken, ServerCapabilities, TextDocumentRegistrationOptions, DocumentSelector, StaticRegistrationOptions,
-	SelectionRangeRequest, SelectionRangeProviderOptions, SelectionRangeKind, SelectionRange, SelectionRangeParams
+	SelectionRangeRequest, SelectionRangeProviderOptions, SelectionRange, SelectionRangeParams
 } from 'vscode-languageserver-protocol';
 
 import { TextDocumentFeature, BaseLanguageClient } from './client';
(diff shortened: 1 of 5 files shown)


client/src/selectionRange.proposed.ts
```diff
@@ -84,84 +84,89 @@ export class SelectionRangeFeature extends TextDocumentFeature<TextDocumentRegis
...
-				return VSelectionRangeKind.Declaration;
-			default:
-				return VSelectionRangeKind.Empty
+	private asSelectionRange(selectionRange: SelectionRange): VSelectionRange | undefined {
+		let range = this._client.protocol2CodeConverter.asRange(selectionRange.range);
+		let parent = undefined;
+		if (selectionRange.parent) {
+			parent = this.asSelectionRange(selectionRange.parent);
+			if (!parent || !(parent.range.contains(range) && !parent.range.isEqual(range))) {
+				return undefined
```
matklad: What is the proper way to report invalid data? eg, when parent is smaller than a child?

client/src/selectionRange.proposed.ts
```diff
@@ -84,84 +84,89 @@ export class SelectionRangeFeature extends TextDocumentFeature<TextDocumentRegis
...
+			}
 		}
+		return new VSelectionRange(range, parent);
 	}
 }
 
 declare module 'vscode' {
```
matklad: note that the API is not proposed in insiders, so we should remove this once we update vscode dependency

client/src/selectionRange.proposed.ts
dbaeumer: @jrieken any recommendation?

jrieken: Yeah, child must be contained by parents

matklad: Yep, and this code checks this conditions. The question is, what to do if the server implementation returns wrong data? 

The behavior should be such that:

* server implementer ideally should see a loud error and fix the problem, instead of spending time with debugger to figure out where exactly the info is dropped
* editor user should ideally get *some* message (such that they can open an issue for the corresponding extension), but it should not be distracting, and it should not be repeating. 


Is there some function like `show_error_dialog_if_in_dev_mode_otherwise_log_a_warning`?

dbaeumer:
I created a branch with the updated API

https://github.com/Microsoft/vscode-languageserver-node/tree/dbaeumer/selectionRange

Regarding error checking:

- since the VS Code API already checks this there is IMO no need to check it in the LSP client again.
- checking this in the server is up to the server implementation / library. If in dev mode you could use the showMessage request to show a message to the user.

Let me know if the changes look OK?

dbaeumer:
Sorry for the overlap but this PR was closed for me :-)

matklad:
Changes in https://github.com/Microsoft/vscode-languageserver-node/tree/dbaeumer/selectionRange look good to me! I would prefer if client did more validation/error reporting, to make it easier to debug a buggy server, and to make protocol more strict, but this would be only a small quality of life improvement, nothing too important.

dbaeumer:
I see the point. Could we do the following. I merge the update into master and then we take it from there?

matklad:
I wasn't clear enough: I would be more more than happy to have any version of this in master, I don't think validation should block anything (or, for that matter, I am only 80% sure that it is worth having validation at all).

dbaeumer:
OK. Great.

matklad:
I guess this can be closed? The impl is already im master!

dbaeumer:
Yes. Thanks for closing.

PR #479, opened 2019-04-10
onImplementation and onTypeDefinition can return DefinitionLink[]

bmewburn:
Implementation and type definition requests can return `DefinitionLink[]`, can't they?


dbaeumer:
Good catch. Thanks!

PR #481, opened 2019-04-14
Fix client capabilities in window/progress proposed doc

Xanewok:
Fixes wrong `ClientCapabilities` extension from #261. 

@dbaeumer by the way, what is the workflow for proposed features? Do they become a part of a regular protocol version but noted that they're unstable and only proposed at the moment?

### DIFF
--- a/protocol/src/protocol.progress.proposed.md
+++ b/protocol/src/protocol.progress.proposed.md
@@ -8,11 +8,11 @@ The client sets the following capability if it is supporting notifying task prog
 
 ```ts
 /**
- * Experimental client capabilities.
+ * Window specific client capabilities.
  */
-experimental: {
+window?: {
   /**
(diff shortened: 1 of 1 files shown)


dbaeumer:
@Xanewok thanks. I already started fixing this on the correspoding progress branch I have. 

While implementing cancel support I separated the notification into start, report, done and cancel. Mainly because it was very hard to descibe in which cases which property combinations were valid. Was a lot easier having separate literals per request. I hope this is OK with you.

For going forward with the proposal: I added a comment here https://github.com/Microsoft/language-server-protocol/issues/70#issuecomment-483252666 to get additional feedback on the proposal. If there is no strong objection I usually merge the proposal into the final version.

I will also publish new next version of the node libraries containing the implementation.

PR #489, opened 2019-04-25
Add isPreferred to the CodeAction protocol.

Jason3S:
See [vscode isPreferred discussion](https://github.com/Microsoft/vscode/issues/62110#issuecomment-454182230)

### DIFF
--- a/types/src/main.ts
+++ b/types/src/main.ts
@@ -2206,12 +2206,12 @@ export interface CodeAction {
 	command?: Command;
 
 	/**
-     * Marks this as a preferred action. Preferred actions are used by the `auto fix` command and can be targeted
-     * by keybindings.
-     *
-     * A quick fix should be marked preferred if it properly addresses the underlying error.
-     * A refactoring should be marked preferred if it is the most reasonable choice of actions to take.
-	*/
+	 * Marks this as a preferred action. Preferred actions are used by the `auto fix` command and can be targeted
+	 * by keybindings.
+	 *
+	 * A quick fix should be marked preferred if it properly addresses the underlying error.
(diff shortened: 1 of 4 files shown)


dbaeumer:
For this to be merged we need to update the spec as well (see https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#L1). In addition I think this needs to be guarded by a client capability so the server can decided if it is worthwhile to compute the property (in this case inexpensive but this is how the spec works today).

Jason3S:
I have updated the spec and I added `isPreferred` to the Code Action capabilities. I hope this is what you meant.

Jason3S:
@dbaeumer Is there more I need to do?

types/src/main.ts
```diff
@@ -2204,6 +2204,15 @@ export interface CodeAction {
 	 * executed and then the command.
 	 */
 	command?: Command;
+
+	/**
+     * Marks this as a preferred action. Preferred actions are used by the `auto fix` command and can be targeted
+     * by keybindings.
+     *
+     * A quick fix should be marked preferred if it properly addresses the underlying error.
+     * A refactoring should be marked preferred if it is the most reasonable choice of actions to take.
```
rcjsuen: Is it just me or does this indenting look off?

Jason3S: It is a tab vs spaces problem. Apparently this files uses tabs. I'll fix it.

Jason3S:
Just waiting on: [Update specification.md Pull Request #736 · microsoft/language-server-protocol](https://github.com/microsoft/language-server-protocol/pull/736)

dbaeumer:
Merged by hand with some minor modifications for consistency (e.g. `isPreferredSupport`).

PR #490, opened 2019-04-26
[RFC] Implement ShowTextDocumentRequest

darkbasic:
See https://github.com/Microsoft/language-server-protocol/issues/612

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -54,7 +54,7 @@ import {
 	ExecuteCommandRequest, ExecuteCommandParams, ExecuteCommandRegistrationOptions,
 	ApplyWorkspaceEditRequest, ApplyWorkspaceEditParams, ApplyWorkspaceEditResponse,
 	MarkupKind, SymbolKind, CompletionItemKind, Command, CodeActionKind, DocumentSymbol, SymbolInformation, Range,
-	CodeActionRegistrationOptions, TextDocumentEdit, ResourceOperationKind, FailureHandlingKind
+	CodeActionRegistrationOptions, TextDocumentEdit, ResourceOperationKind, FailureHandlingKind, ShowTextDocumentRequest
 } from 'vscode-languageserver-protocol';
 
 import { ColorProviderMiddleware } from './colorProvider';
@@ -2758,6 +2758,10 @@ export abstract class BaseLanguageClient {
 				}
 				let actions = params.actions || [];
 				return messageFunc(params.message, ...actions);
+			});
+			connection.onRequest(ShowTextDocumentRequest.type, async (params) => {
+				const document = await Workspace.openTextDocument(params.fileName);
+				return Window.showTextDocument(document, this.protocol2CodeConverter.asTextDocumentShowOptions(params.options));
 			});
 			connection.onTelemetry((data) => {
 				this._telemetryEmitter.fire(data);
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -2760,13 +2760,9 @@ export abstract class BaseLanguageClient {
 				return messageFunc(params.message, ...actions);
 			});
 			connection.onRequest(ShowTextDocumentRequest.type, async (params) => {
-				try {
-					const { textDocument: { uri }, options } = params;
-					const document = await Workspace.openTextDocument(Uri.parse(uri));
-					return Boolean(await Window.showTextDocument(document, this.protocol2CodeConverter.asTextDocumentShowOptions(options)));
-				} catch (err) {
-					return false;
-				}
+				const { textDocument: { uri }, options } = params;
+				const document = await Workspace.openTextDocument(Uri.parse(uri));
+				await Window.showTextDocument(document, this.protocol2CodeConverter.asTextDocumentShowOptions(options));
 			});
 			connection.onTelemetry((data) => {
 				this._telemetryEmitter.fire(data);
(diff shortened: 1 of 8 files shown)


protocol/src/protocol.ts
```diff
@@ -1207,6 +1207,26 @@ export namespace ShowMessageRequest {
 	export const type = new RequestType<ShowMessageRequestParams, MessageActionItem | null, void, void>('window/showMessageRequest');
 }
 
+export interface ShowTextDocumentRequestParams {
+	/**
+	 * A name of a file on disk.
+	 */
+	fileName: string;
```
matklad: This should be 

```
textDocument: TextDocumentIdentifier;
```

types/src/main.ts
```diff
@@ -2611,6 +2611,92 @@ export interface TextDocumentContentChangeEvent {
 	text: string;
 }
 
+/**
+ * Denotes a location of an editor in the window. Editors can be arranged in a grid
+ * and each column represents one editor location in that grid by counting the editors
+ * in order of their appearance.
+ */
+export enum ViewColumn {
```
matklad: This exposes implementation details of VS Code at the protocol level, which doesn't feel right. 

If you really need to control VS Code-specific editor placement, a better approach could be to make a custom command in your VS Code extension.

This probably won't even work for VS Code, given the recent switch to grid layout.

protocol/src/protocol.ts
rcjsuen: Agreed. This needs to be a URI and not the name of a file.

protocol/src/protocol.ts
```diff
@@ -1207,6 +1207,26 @@ export namespace ShowMessageRequest {
...
+
+	/**
+	 * Editor options to configure the behavior of showing the editor.
+	 */
+	options: TextDocumentShowOptions;
```
rcjsuen: As all the properties within `TextDocumentShowOptions` is optional, should `options` itself be `options?` instead?

protocol/src/protocol.ts
darkbasic: Fixed, thanks.

types/src/main.ts
darkbasic: I don't really need a specific editor placement, so I'm simply omitting it for the moment.

protocol/src/protocol.ts
darkbasic: It's optional now, thanks.

client/src/client.ts
```diff
@@ -2759,6 +2759,15 @@ export abstract class BaseLanguageClient {
 				let actions = params.actions || [];
 				return messageFunc(params.message, ...actions);
 			});
+			connection.onRequest(ShowTextDocumentRequest.type, async (params) => {
+				try {
+					const { textDocument: { uri }, options } = params;
+					const document = await Workspace.openTextDocument(Uri.parse(uri));
+					return Boolean(await Window.showTextDocument(document, this.protocol2CodeConverter.asTextDocumentShowOptions(options)));
```
darkbasic: I've changed the return type to `boolean` to give the server some knowledge about the client successfully opening the document or not.
I don't know what's the best way to accomplish this with the language server protocol (maybe returning the error would be better?), so feedback is welcome.

rcjsuen: What do you anticipate servers doing if the client fails to open the file?

matklad: JSON-RPC supports returning either ok response or an error natively. There's no need to make the return type here non-void, the client can communicate the error using the protocol directly

See https://microsoft.github.io/language-server-protocol/specification, section `Response Message`.

rcjsuen: Just to follow-up on why I said what I said, I was wondering if this could be a notification instead of a response? It's not clear to me what a server would do in response to a client that didn't open the file.

dbaeumer:
@darkbasic I would like to ask you to provide this as a protocol extension like described in https://github.com/microsoft/language-server-protocol/blob/master/contributing.md. This has the advantage that we have separate files which eases the code review. We could then even have it in a next version of the LSP libraries.

You could take `protocol/src/protocol.configuration.ts` `client/src/configuration.ts` and `server/src/configuration.ts` as a blue print since it sends a request from the server to the client as well.

client/src/client.ts
darkbasic: Thanks, I'll simply take advantage of the JSON-RPC capability to natively return error messages then.

darkbasic:
@dbaeumer thanks, it's under its own standalone proposal now.

client/src/showTextDocument.proposed.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { Uri, workspace, window } from 'vscode';
+import { StaticFeature, BaseLanguageClient, ShowTextDocumentRequest } from './client';
+
+export class ShowTextDocumentFeature implements StaticFeature {
+
+	constructor(private _client: BaseLanguageClient) {
+	}
+
+	public fillClientCapabilities(): void {
```
darkbasic: What is `fillClientCapabilities` for?
Is `StaticFeature` the right interface to implement for this class?

rcjsuen: It's for a client to declare that it supports this feature. Otherwise, you have servers sending this request to the client when the client doesn't even support this. It does not seem like you have this feature declared as a client capability in the protocol at the moment though.

darkbasic: Thanks, I've declared the client capabilities!

darkbasic:
Client capabilities are now declared.
@rcjsuen I thinked about it and there is nothing the server can do if the client fails to open the editor. In fact I decided to not return a boolean anymore: if there is an error it can still be catched thanks to the protocol capabilities.

dbaeumer:
@darkbasic I re-implemented this since a lot has changed since the creation of the PR. I will still mention you in the N&N since you are the original creator of the code.

PR #583, opened 2020-03-04
Fix typo in name of semantic tokens file

rcjsuen:
Rename `protocol.sematicTokens.proposed.ts` to `protocol.semanticTokens.proposed.ts`.


dbaeumer:
Thanks!

PR #591, opened 2020-03-14
Fix leftover reference to WorkDoneProgressStart

Gama11:
I think `WorkDoneProgressStart` got removed during some of the initial experimentation / refactoring.


dbaeumer:
Actually a better fix is to remove the sentence due to the follow up sentence 

> Clients that don't support cancellation or don't support control the button's enablement state are allowed to ignore the setting.

dbaeumer:
Thanks for bringing this up!

PR #598, opened 2020-03-26
re-export MessageConnection so that vscode-languageserver/client is s…

heejaechang:
…elf contained

MessageConnection is exposed through CanellationSenderStrategy


heejaechang:
tag @dbaeumer

dbaeumer:
Thanks !

PR #609, opened 2020-04-07
Feature Proposal: Inlay Hints

kjeremy:
This is the start of an "inlay hints" feature proposal. It is essentially the interface we use in [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer) to achieve inlay hints (https://github.com/microsoft/vscode/issues/16221)

I am working on this in my spare time (hence just the md for now) but would like to get feedback.

### DIFF
--- a/protocol/src/protocol.inlayHints.proposed.md
+++ b/protocol/src/protocol.inlayHints.proposed.md
@@ -1,0 +1,109 @@
+
+#### Inlay Hints
+
+The LSP provides retrieving inlay hints information with the following request.
+
+_Client Capabilities_:
+
+```ts
+InlayHintsClientCapabilities {
+    /**
+     * The text document client capabilities
+     */
+    textDocument?: {
+        /**
+         * Capabilities specific to the `textDocument/inlayHints`
+         */
+        inlayHints?: {
+            /**
+             * Whether implementation supports dynamic registration. If this is set to `true`
+             * the client supports the new `(TextDocumentRegistrationOptions & StaticRegistrationOptions)`
+             * return value for the corresponding server capability as well.
+             */
+            dynamicRegistration?: boolean;
+        };
+    }
+}
+```
+
+_Server Capabilities_:
+
+```ts
+InlayHintsServerCapabilities {
+    /**
+     * The server provides Inlay Hint support.
+     */
+    inlayHintsProvider?: boolean | (TextDocumentRegistrationOptions & StaticRegistrationOptions);
+}
+```
+
+##### Inlay Hints Request
+
+_Request_:
+
+The `textDocument/inlayHints` request is sent from the client to the server to request the hints for a given
+text document.
+
+Returns a collection of hints for the text document.
+
+* method: ‘textDocument/inlayHints'
+* params: `InlayHintsParams` defined as follows:
+
+```ts
+export interface InlayHintsParams {
+    /**
+     * The text document
+     */
+    textDocument: TextDocumentIdentifier;
+}
+```
+
+_Response_:
+
+The server will send a collection of `InlayHint` objects, or `null` if no hints are found.
+
+Each `InlayHint` object defines a range and kind corresponding to the hint and a label
+to be displayed inline in the editor.
+
+* result: `InlayHint[]` | `null`
+
+```ts
+export interface InlayHint {
+
+    /**
+     * The source range of the hint.
+     */
+    range: Range;
+
+    /**
+    * The type of hint.
+     */
+    kind: InlayHintKind;
+
+    /**
+     * The label that is displayed the in editor
+    */
+    label: string;
+}
+
+/**
+ * A set of predefined hint kinds
+ */
+export interface InlayHintKind {
+
+	/**
+	 * The hint is for type information
+	 */
+	export const TypeHint: InlayHintKind = 'TypeHint';
+
+	/**
+	 * The hint corresponds to parameter information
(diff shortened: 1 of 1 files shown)


rcjsuen:
> I am working on this in my spare time (hence just the md for now) but would like to get feedback.

@kjeremy I suggest you share your proposal as an issue in the [microsoft/language-server-protocol repository](https://github.com/microsoft/language-server-protocol).

dbaeumer:
@jrieken and @mjbvz has something like this every come up in the VS Code API. I actually do like the feature especially for inferred types.

kjeremy:
A screen shot of this in action can be found at https://github.com/microsoft/vscode/issues/16221#issuecomment-570848310

jrieken:
https://marketplace.visualstudio.com/items?itemName=sokra.ts-inline-types 👈 an extension that does this for TypeScript - tho, by creating yet another instance of the compiler

Veetaha:
What we do in `rust-analyzer` is:
1. Send a request `rust-analyzer/inlayHints` on any source change (i.e. any character typed, any file moved/created/deleted etc.).
2. As a parameter we pass `{ textDocument: TextDocumentIdentifier }` for the particular file we need to get inlay hints for.
3. The server responds with an array of `InlayHint` objects for the requested file.

Here is a simplified, *stringly-typed* version of it ([we use a discriminated union at our repo instead](https://github.com/rust-analyzer/rust-analyzer/blob/4762c6d9c66dc1b6be9b9010dbe787ef8d69530a/editors/code/src/rust-analyzer-api.ts#L91-L108)):
```ts
import * as lc from "vscode-languageclient";

export interface InlayHint {
    range: lc.Range;
    label: string;
    kind: string;
}

export interface InlayHintsParams {
    textDocument: lc.TextDocumentIdentifier;
}

const inlayHints = new lc.RequestType<InlayHintsParams, InlayHint[], unknown>(
    "rust-analyzer/inlayHints"
);
```
[The client-side implementation for this feature is very straight-forward and actually fits into < 250 lines of code](https://github.com/rust-analyzer/rust-analyzer/blob/master/editors/code/src/inlay_hints.ts). It currently uses the vscode text decorations feature (more precise inlay hints are `after` and `before` css pseudoelements), but this feature is **very limited**.
Current limitations:
1. It feels unnatural when the cursor goes through the inlay hint, it's hard to describe it better be seen:
inlays_limitations
The cursor kind jumps through the hint and the nearby token. IntelliJ in this case emulates the inlay hint being a real holistic token in the editor.
2. As you can see on the *.gif* these `after` and `before` pseudoelements inherit the styling from the token they are attached to. This means whenever that token (e.g. the variable name, a paren, a coma etc.) changes its style (e.g. is highlighted, outlined, etc) the inlay hint changes its style along with it, which is unwanted...
3. Text decorations don't let you tweak too much of their style. For example, we cannot change their font size to make them a bit smaller and vertically center them in the middle of the line of code.

Our current implementation can be very easily generalized and integrated into LSP:
1. The inlay hint `kind` should not be a fixed enum (I propose to change it in this PR @kjeremy), but just some string identifier for which the user can customize display styling. This gives lsp server developers the flexibility to add new kinds of inlay hints they can that think of without changing the protocol, and the editor should be abstract and not provide any ad hoc implementation (but we might consider the protocol to specify the `well-known` inlay hint kinds where special support for them may be implemented (e.g. by clicking on the type hint the editor invokes goto-definition of the type, or by clicking on the parameter name hint it invokes goto-definition of the parameter on the function that is invoked, etc.)
2. The server itself should decide to which token and on **which side** (after/before) of the token the editor should place the inlay hint. So we might modify the `InlayHint` interface:
    ```ts
    interface InlayHint {
        range: lc.Range;
        kind: string;
        label: string;
        position: "after" | "before";
    }
    ```
3. The strategy for invalidating the inlay hints should also be tweakable. For example, the server should register the inlay hints capability with the following options for invalidation.
Just a naive scratch of invalidation events:
    ```ts
    enum InlayHintsInvalidationEvents {
        TextDocumentContentChanged = 0b0001,
        WorkspaceFileRemoved = 0b0010,
        WorkspaceFileAdded = 0b0100,
        WorkspaceFileMoved = 0b1000
    }  
    ```

So this describes the pull-based implementation of inlay hints that is currently used at `rust-analyzer`.
We might think of a push-based implementation instead, where the server can decide whether inlay hints are invalidated or not, but I guess we cannot easily implement it via only LSP protocol right now, this might require some more changes to the protocol to let the server always know which text documents are opened in the editor at any point of time.

Maybe @matklad  can elaborate on Intelij where they've natively implemented this feature and which model/API they've considered best to be used.

KamasamaK:
@mjbvz I see that you locked microsoft/vscode#16221. I agree that the nature of the comments mostly amounted to noise, but locking also prevents further actual +1s on the opening post.

mjbvz:
We are still tracking it. There were enough +1 and comments on the issue to know that it’s something users are interested in

matklad:
Something which I believe is missing form discussion so far is that one could (but not necessary should) go beyond plain text for hints:

image

Veetaha:
Maybe we should move the discussion to https://github.com/microsoft/language-server-protocol/issues/956?

lastthyme (approved):

dbaeumer:
Lets close this in favour of https://github.com/microsoft/vscode-languageserver-node/pull/772

PR #611, opened 2020-04-10
Aligns language server with protocol proposition

giltho:
Aligns the server with microsoft/language-server-protocol#953

### DIFF
--- a/jsonrpc/src/common/connection.ts
+++ b/jsonrpc/src/common/connection.ts
@@ -247,7 +247,7 @@ export interface SetTraceParams {
 }
 
 export namespace SetTraceNotification {
-	export const type = new NotificationType<SetTraceParams, void>('$/setTraceNotification');
+	export const type = new NotificationType<SetTraceParams, void>('$/setTrace');
 }
 
 export interface LogTraceParams {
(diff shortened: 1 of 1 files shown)


jsonrpc/src/main.ts
```diff
@@ -269,7 +269,7 @@ export interface LogTraceParams {
 }
 
 export namespace LogTraceNotification {
-	export const type = new NotificationType<LogTraceParams, void>('$/logTraceNotification');
+	export const type = new NotificationType<LogTraceParams, void>('logTrace');
```
dbaeumer, line 272: Actually we should name this `$/logTrace` since clients and servers don't need to implement this.

dbaeumer:
@giltho should we for consistency change this one as well?

https://github.com/microsoft/vscode-languageserver-node/blob/master/jsonrpc/src/main.ts#L262

giltho:
Yes I think it would be better indeed. Also, I could probably add it to the protocol proposition.

dbaeumer:
That would be great!

giltho:
I updated the draft here microsoft/language-server-protocol#953 :)

dbaeumer:
@giltho could you resolve the conflict please.

giltho:
@dbaeumer done :)

PR #613, opened 2020-04-16
npm audit fix

kjeremy:


dbaeumer:
I fixed this yesterday by moving the mocha version to 7.x. In general I like to update root dependencies :-)

PR #614, opened 2020-04-16
Stabilize call hierarchy

kjeremy:


protocol/src/protocol.ts
```diff
@@ -665,6 +679,11 @@ export interface _ServerCapabilities<T = any> {
 	 */
 	executeCommandProvider?: ExecuteCommandOptions;
 
+	/**
+	 * The server provides Call Hierarchy support.
+	 */
```
rcjsuen, line 684: Can we add an `@since 3.16.0` tag here too?

protocol/src/protocol.ts
```diff
@@ -436,6 +443,13 @@ export interface TextDocumentClientCapabilities {
 	 * Capabilities specific to `textDocument/publishDiagnostics`.
 	 */
 	publishDiagnostics?: PublishDiagnosticsClientCapabilities;
+
+	/**
+	 * Capabilities specific to the `textDocument/callHierarchy`.
+	 *
+	 * @since 3.16.0 - Proposed state
```
rcjsuen, line 450: Should the `Proposed state` comment be removed?

client/src/callHierarchy.ts
```diff
@@ -195,27 +200,27 @@ class CallHierarchyProvider implements VCallHierarchyProvider {
...
-		const capabilites: ClientCapabilities & Proposed.CallHierarchyClientCapabilities = cap as ClientCapabilities & Proposed.CallHierarchyClientCapabilities;
+		const capabilites: ClientCapabilities & CallHierarchyClientCapabilities = cap as ClientCapabilities & CallHierarchyClientCapabilities;
 		const capability = ensure(ensure(capabilites, 'textDocument')!, 'callHierarchy')!;
 		capability.dynamicRegistration = true;
 	}
 
-	public initialize(cap: ServerCapabilities, documentSelector: DocumentSelector): void {
-		const capabilities: ServerCapabilities & Proposed.CallHierarchyServerCapabilities = cap;
+	public initialize(capabilities: ServerCapabilities, documentSelector: DocumentSelector): void {
+		//const capabilities: ServerCapabilities & CallHierarchyServerCapabilities = cap;
```
rcjsuen, line 215: Should this line be deleted if it's not used?

protocol/src/protocol.ts
kjeremy: None of the other fields contain `@since` so I didn't add it here either.

kjeremy:
@dbaeumer Is this reasonable?

dbaeumer:
@kjeremy have not forgotten you but needed to finish things up for the endgame this week. Will take care of it this week.

protocol/src/protocol.ts
dbaeumer: I was sloppy on this. We should add @since 3.16 here. I tried to do this for all new features as well.

protocol/src/protocol.callHierarchy.ts
```diff
@@ -16,7 +16,7 @@ import {
  * Represents programming constructs like functions or constructors in the context
  * of call hierarchy.
  *
- * @since 3.16.0 - Proposed state
+ * @since 3.16.0
  */
 export interface CallHierarchyItem {
```
dbaeumer, line 21: For consistency reason this should go into the types package.

protocol/src/protocol.callHierarchy.ts
```diff
@@ -59,7 +59,7 @@ export interface CallHierarchyItem {
 /**
  * Represents an incoming call, e.g. a caller of a method or constructor.
  *
- * @since 3.16.0 - Proposed state
+ * @since 3.16.0
  */
 export interface CallHierarchyIncomingCall {
```
dbaeumer, line 64: For consistency reason this should go into the types package.

protocol/src/protocol.callHierarchy.ts
```diff
@@ -78,7 +78,7 @@ export interface CallHierarchyIncomingCall {
 /**
  * Represents an outgoing call, e.g. calling a getter from a method or a method from a constructor etc.
  *
- * @since 3.16.0 - Proposed state
+ * @since 3.16.0
  */
 export interface CallHierarchyOutgoingCall {
```
dbaeumer, line 83: For consistency reason this should go into the types package.

rcjsuen:
@dbaeumer If it becomes stable should the `protocol.callHierarchy.proposed.md` file be outright deleted?

dbaeumer:
@rcjsuen yes and no. The MD file needs to be merged into a non existing 3.16 version of the spec. I will start this on the protocol repository so that someone can create a PR for that as well.

Just to make things clear: moving it out of the proposed namespace doesn't mean we officially ship 3.16. This also means that the lsp node libraries will still be published as next version. We can only make this official with a official 3.16 release of the protocol.

dbaeumer:
I created the 3.16 specification document. The call hierarchy mark down can go here: https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md#L5634

PR #624, opened 2020-05-22
fix InsertTextFormat comment

h-michael:
The referenced file was moved.


rcjsuen:
@dbaeumer Looks like this should be a canonical link instead. Otherwise, the specification of a snippet changes as modifications are made to that Markdown file instead of a snapshot of it in time. 🤔

dbaeumer:
@h-michael thanks for the PR. @rcjsuen you are correct. I change it to point to the spec.

Issue #628, opened 2020-06-04
How to get the lsp response in vscode-language-client?


Abbyyan:
I'm working on a vscode lsp extension development and want to ask if there are some methods that i can get the lsp response in the vscode-language-client ? Hope for your help. Thanks a lot.

dbaeumer:
@Abbyyan the LSP client has a middleware which all requests and response go through.

Abbyyan:
How can i get the full completion items in the server response? Should i use `provideCompletionItem` or `resolveCompletionItem`  ? Thanks for your help.

Abbyyan:
Is there some documents shows how to use them please?

rcjsuen:
> How can i get the full completion items in the server response? Should i use `provideCompletionItem` or `resolveCompletionItem` ? Thanks for your help.

You probably want the former but you may want both depending on your exact use case.

> Is there some documents shows how to use them please?

Maybe the code from #322 may serve as a bit of a reference for you.

Abbyyan:
Learned from [lsp-example](https://github.com/microsoft/vscode-extension-samples/blob/1c9a14f3e481d8fbfeec4a645e476bbaf8333d7a/lsp-embedded-request-forwarding/client/src/extension.ts#L47) that i can change the input send to server (just as the uri in the example). But i still confused about get the response (or other info ) from server.  For example , how can i get the completion list in provideCompletionItem? Hope for your answer please. Thanks for your help！ 
```js
let clientOptions: LanguageClientOptions = {
		documentSelector: [{ scheme: 'file', language: 'html1' }],
		middleware: {
			provideCompletionItem: async (document, position, context, token, next) => {
				// If not in `<style>`, do not perform request forwarding
				if (!isInsideStyleRegion(htmlLanguageService, document.getText(), document.offsetAt(position))) {
					return await next(document, position, context, token);
				}

				const originalUri = document.uri.toString();
				virtualDocumentContents.set(originalUri, getCSSVirtualContent(htmlLanguageService, document.getText()));

				const vdocUriString = `embedded-content://css/${encodeURIComponent(
					originalUri
				)}.css`;
				const vdocUri = Uri.parse(vdocUriString);
				return await commands.executeCommand<CompletionList>(
					'vscode.executeCompletionItemProvider',
					vdocUri,
					position,
					context.triggerCharacter
				);
			}
		}
	};
```

dbaeumer:
[...]

PR #637, opened 2020-06-11
Add calls to validatePosition for some requests

davidanthoff:
This is an attempt to improve the situation described in https://github.com/microsoft/language-server-protocol/issues/946.

Be warned: I really don't know what I'm doing here. These diffs look plausible to me, but I have really no idea. I also don't know how to try the code in this repo, so as far as I can tell this compiles, but beyond that I haven't tested it :)

### DIFF
--- a/client/src/common/client.ts
+++ b/client/src/common/client.ts
@@ -1502,7 +1502,8 @@ class CompletionItemFeature extends TextDocumentFeature<CompletionOptions, Compl
 				const client = this._client;
 				const middleware = this._client.clientOptions.middleware!;
 				const provideCompletionItems: ProvideCompletionItemsSignature = (document, position, context, token) => {
-					return client.sendRequest(CompletionRequest.type, client.code2ProtocolConverter.asCompletionParams(document, position, context), token).then(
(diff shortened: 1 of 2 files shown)


dbaeumer:
We should never change the position of a request since the sender of the request might not be aware of this. All we can do is to reject the request if the position is outside the document range. And now that I think about this it should may be not be something that goes into the LSP libs right away since I never heard about such a problem. @davidanthoff have you tried to handle this in a custom middleware on your end to see what is actually the cause of this?

davidanthoff:
No, I haven't tried that. Is there some pointer how I might be able to add custom middleware?

rcjsuen:
> Is there some pointer how I might be able to add custom middleware?

@davidanthoff Maybe #322 and #628 can help?

davidanthoff:
Yes, those pointers were super helpful!

I've added diagnostic code now in our extension (see https://github.com/julia-vscode/julia-vscode/pull/1333), lets see what we get from that. I'll report back if we find anything interesting.

Thanks for the help! I'll close this PR here, the middleware route is clearly better.

Issue #418, opened 2018-09-26
Consider exposing the restartTimes to LanaguageClientOptions


hokein:
The languageclient restarts LSP server 5 times if LSP server crashes. The restart time is hard-coded in https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/client.ts#L272. 

Can we expose it as an option like `restartTimes` in `LanguageClientOptions`, so that LSP extension can customize it? I know LSP extension can provide their custom `ErrorHandler` to adjust the restart times, but I think it is still sensible to provide such option.

dbaeumer:
@hokein exposing it makes sense. A PR is welcome for this.

PR #658, opened 2020-08-12
Expose restartTimes to client.

hokein:
Implements https://github.com/microsoft/vscode-languageserver-node/issues/418

### DIFF
--- a/client/src/common/client.ts
+++ b/client/src/common/client.ts
@@ -289,12 +289,12 @@ class DefaultErrorHandler implements ErrorHandler {
 	}
 	public closed(): CloseAction {
 		this.restarts.push(Date.now());
-		if (this.restarts.length < this.maxRestartCount) {
+		if (this.restarts.length <= this.maxRestartCount) {
 			return CloseAction.Restart;
 		} else {
 			let diff = this.restarts[this.restarts.length - 1] - this.restarts[0];
 			if (diff <= 3 * 60 * 1000) {
-				Window.showErrorMessage(`The ${this.name} server crashed ${this.maxRestartCount} times in the last 3 minutes. The server will not be restarted.`);
+				Window.showErrorMessage(`The ${this.name} server crashed ${this.maxRestartCount+1} times in the last 3 minutes. The server will not be restarted.`);
 				return CloseAction.DoNotRestart;
 			} else {
 				this.restarts.shift();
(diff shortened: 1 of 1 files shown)


client/src/common/client.ts
```diff
@@ -276,7 +277,7 @@ class DefaultErrorHandler implements ErrorHandler {
 
 	private readonly restarts: number[];
 
-	constructor(private name: string) {
+	constructor(private name: string, private restartTimes: number) {
```
dbaeumer, line 280: I would suggest to name this maxRestartCount. We usually use count instead of times for this.

hokein: Done.

client/src/common/client.ts
```diff
@@ -163,7 +163,7 @@ interface ConnectionCloseHandler {
 
 interface ConnectionOptions {
     cancellationStrategy: CancellationStrategy
-    restartTimes: number
+    maxRestartCount: number
```
dbaeumer, line 166: Actually more stuff. Sorry for this. For backwards compatibility this should be in the whole call sequence be optional.

client/src/common/client.ts
```diff
@@ -2767,8 +2767,8 @@ export abstract class BaseLanguageClient {
 		return this._diagnostics;
 	}
 
-	public createDefaultErrorHandler(restartTimes: number): ErrorHandler {
-		return new DefaultErrorHandler(this._name, restartTimes);
+	public createDefaultErrorHandler(maxRestartCount: number): ErrorHandler {
```
dbaeumer, line 2770: See above.

client/src/common/client.ts
hokein: Done.

client/src/common/client.ts
hokein: Done.

client/src/common/client.ts
```diff
@@ -2767,8 +2767,8 @@ export abstract class BaseLanguageClient {
 		return this._diagnostics;
 	}
 
-	public createDefaultErrorHandler(maxRestartCount: number): ErrorHandler {
-		return new DefaultErrorHandler(this._name, maxRestartCount);
+	public createDefaultErrorHandler(maxRestartCount?: number): ErrorHandler {
+		return new DefaultErrorHandler(this._name, maxRestartCount || 5);
```
dbaeumer, line 2771: This is not correct. It will even default to 5 if maxRestartCount is 0 since `||` is coercive. Did you mean maxRestartCount ?? 5

hokein: Thanks! Ah, right, `??` is what I meant.

Also added a sanity check for invalid value, It should keep the original behavior now.

Zenix27 (approved):

PR #659, opened 2020-08-18
export common `createConnection` method

acao:
When I saw that there is a wildcard export of `server/common/api` in `server/node/main`, I assumed the intention might have been to allow people to create their own service interface, similar to `node/main` or `browser/main`. however, the `createConnection` method from `common` is not available in node obviously because it's overwritten!

If this was the original intention, to only export one `createConnection` interface, then here is our use case: Our language server that is used by a number of popular extensions originally adopted `vscode-jsonrpc` `createMessageConnection` directly in 2017. we've just recently hit a ceiling of course, as we have a lot of missing features from the `createConnection` interface in the docs. We have our ownt= similar [service interface for choosing the transport](https://github.com/graphql/graphiql/blob/main/packages/graphql-language-service-server/src/startServer.ts#L146), for example. 

Is there a reason those process argument flags (`--node-ipc`, etc) are baked in? Are those used by vscode, or *only* for the `server/node/main` `_createConnection`?

if it wasn't the original intention, then, and you _did_ in fact want to export the common `createServer` - well you have a bugfix!

### DIFF
--- a/server/src/browser/main.ts
+++ b/server/src/browser/main.ts
@@ -5,7 +5,7 @@
 
 import {
 	MessageReader, MessageWriter, Logger, ConnectionStrategy, ConnectionOptions, ProtocolConnection, WatchDog, InitializeParams, createProtocolConnection,
-	createConnection as createCommonConnection, Connection, Features, _Connection, _
+	createCommonConnection, Connection, Features, _Connection, _
 } from '../common/api';
 
 export * from 'vscode-languageserver-protocol/browser';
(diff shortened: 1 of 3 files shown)


dbaeumer:
I still think that it is enough to have one `createConnection`. If you call the function with a given input and output the whole code that interprets the command line args is skipped. Why doesn't that work for you?

acao:
@dbaeumer to be honest, I'm still in the process of discovering why our initial author chose to implement `createMessageConnection`, possibly he had the same mistaken assumption. I feel foolish after having stared at these very interfaces last night, haha! anyways, looks like it works a charm, i had just been passing a logger as a third argument accidentally.

PR #670, opened 2020-09-14
Proposed details for textDocument/resolveMoniker request

BertanAygun:
Language Server Index Format (LSIF) introduced the concept of symbol monikers to help associate symbols across different indexes. This request adds capability for LSP server implementations to provide the same symbol moniker information given a text document position. Clients can utilize this method to get the moniker at the current location in a file user is editing and do further code navigation queries in other services that rely on LSIF indexes and link symbols together.


protocol/src/common/protocol.resolveMoniker.proposed.md
```diff
@@ -0,0 +1,28 @@
+#### Resolve Symbol Moniker (textDocument/resolveMoniker)
+
+Language Server Index Format (LSIF) introduced the concept of symbol monikers to help associate symbols across different indexes. This request adds capability for LSP server implementations to provide the same symbol moniker information given a text document position. Clients can utilize this method to get the moniker at the current location in a file user is editing and do further code navigation queries in other services that rely on LSIF indexes and link symbols together.
+
+_Client Capabilities_:
```
BertanAygun, line 5: Do we need any client capabilities here since there is no optional behavior per client?

Also do we need server capabilities? Is it possible to add them as part of the protocol.resolveMoniker.ts file?

protocol/src/common/protocol.resolveMoniker.proposed.md
```diff
@@ -0,0 +1,28 @@
+#### Resolve Symbol Moniker (textDocument/resolveMoniker)
```
dbaeumer, line 1: Opt to name this textDocument/moniker. In LSP we use the name resolve to compute additional properties for a object returned from an LSP server. So for example completionItem/resolve which resolves additional properties for a completion item previously returned from a textDocument/completion request.

protocol/src/common/protocol.ts
```diff
@@ -2807,7 +2811,9 @@ export {
 	// Semantic Tokene
 	SemanticTokenTypes, SemanticTokenModifiers, SemanticTokensLegend, SemanticTokens, SemanticTokensPartialResult, SemanticTokensEdit, SemanticTokensDelta,
 	SemanticTokensDeltaPartialResult, TokenFormat, SemanticTokensClientCapabilities, SemanticTokensOptions, SemanticTokensRegistrationOptions, SemanticTokensParams,
-	SemanticTokensRequest, SemanticTokensDeltaParams, SemanticTokensDeltaRequest, SemanticTokensRangeParams, SemanticTokensRangeRequest, SemanticTokensRefreshNotification
+	SemanticTokensRequest, SemanticTokensDeltaParams, SemanticTokensDeltaRequest, SemanticTokensRangeParams, SemanticTokensRangeRequest, SemanticTokensRefreshNotification,
+	// Resolve moniker support
+	Moniker, ResolveMonikerParams, ResolveMonikerRegistrationOptions, ResolveMonikerRequest
```
dbaeumer, line 2816: You can't export this here since it is a proposed feature. This needs to go into the api.ts file first in the proposed namespace. You can look into this version https://github.com/microsoft/vscode-languageserver-node/blob/66f91b3583a19ae621214d908ce19c9b2d182626/protocol/src/common/api.ts

protocol/src/common/protocol.resolveMoniker.proposed.md
dbaeumer: I would add a client cap with a dynamiceRegistration property to keep it consistent. 

We do need a server capabilitz to indicate whether a server will provide a moniker request handler. Otherwise the client doesn't know if it can call the server.

server/src/common/resolveMoniker.proposed.ts
```diff
@@ -0,0 +1,32 @@
+/* --------------------------------------------------------------------------------------------
```
dbaeumer, line 1: You also need to register the feature as a proposed feature. Otherwise the code will not be available when running the server. See https://github.com/microsoft/vscode-languageserver-node/blob/66f91b3583a19ae621214d908ce19c9b2d182626/server/src/common/api.ts as an example.

dbaeumer (changes requested):

dbaeumer:
Nice work!

protocol/src/common/protocol.resolveMoniker.proposed.md
BertanAygun: Sure, will add them however I am not clear on what client capability would indicate in this case? Would it be for future use?

protocol/src/common/protocol.resolveMoniker.proposed.md
BertanAygun: Will do, I was going between textDocument/moniker, textDocument/symbolMoniker, textDocument/resolveMoniker

protocol/src/common/protocol.resolveMoniker.proposed.md
dbaeumer: It would indicate `dynamiceRegistration` for now

server/src/common/moniker.proposed.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import { Proposed } from 'vscode-languageserver-protocol';
+
+import type { Feature, _Languages, ServerRequestHandler } from './server';
+
+export interface MonikerFeatureShape {
+	resolveMoniker: {
```
dbaeumer, line 12: This should be named moniker as well.

dbaeumer (changes requested):

BertanAygun:
PR to update 3.16 spec is here https://github.com/microsoft/language-server-protocol/pull/1095

Now that we have the documentation there, should I remove the markdown file from here? What about the proposed suffixes on the files?

PR #686, opened 2020-11-07
Allow customizing typing latency

strager:
The default document change delay of 200 milliseconds is too high or too
low for some Visual Studio Code plugins. Allow users to customize the
document change delay to any number (including 0).

QUESTIONS FOR CODE REVIEWERS:

* Where should documentation for this option go?
* Is there a better spot for the changeDelay option? Is
  SynchronizeOptions a good place?
* Should we validate the option, disallowing negative numbers, NaN,
  Infinity, etc.?

### DIFF
--- a/client/src/common/client.ts
+++ b/client/src/common/client.ts
@@ -512,7 +512,8 @@ export interface LanguageClientOptions {
 	connectionOptions?: ConnectionOptions;
 	markdown?: {
 		isTrusted?: boolean;
-	}
+	};
+	changeDelay?: number;
 }
 
 interface ResolvedClientOptions {
@@ -535,7 +536,8 @@ interface ResolvedClientOptions {
 	connectionOptions?: ConnectionOptions;
 	markdown: {
 		isTrusted: boolean;
-	}
+	};
+	changeDelay: number;
 }
 
 export enum State {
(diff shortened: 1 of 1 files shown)


dbaeumer:
Instead of adjusting this I would rather opt that plugins having problems with this switch to an incremental buffer sync which is supported as well and has a delay of 0. 

Would this be an options for you?

strager:
> Instead of adjusting this I would rather opt that plugins having problems with this switch to an incremental buffer sync which is supported as well and has a delay of 0.

Are you talking about having the LSP server support TextDocumentSyncKind.Incremental? Sure, I can try that.

dbaeumer:
@strager yes. As two big advatanges:

- no additional latency
- minimal data transfer.

strager:
> Instead of adjusting this I would rather opt that plugins having problems with this switch to an incremental buffer sync which is supported as well and has a delay of 0.
> 
> Would this be an options for you?

Using LSP's incremental feature works great!

Do you think this PR's feature is worth adding regardless? Or should we discourage LSP servers from only supporting full document changes?

dbaeumer:
I actually would like to discourage full buffer sync. It is basically there to get started or to prototype something. But for real server we should promote incremental sync. So I opt to close the PR. Objections?

PR #687, opened 2020-11-10
Add {will,did}{Create,Rename,Delete}Files

DanTup:
~~This is still very much WIP, but opening to get some feedback/guidance.~~

Implementation for https://github.com/microsoft/language-server-protocol/pull/1074.


client-node-tests/src/integration.test.ts
```diff
@@ -619,6 +619,22 @@ suite('Client integration', () => {
...
+				{ uri: '???' }
+			];
+			const feature = client.getFeature(lsclient.WillCreateFilesRequest.method);
+			isDefined(feature);
+			feature.fillClientCapabilities(???);
+
+			// Trigger a VS Code file creation event
+
+			// Ensure it triggered a well-known file edit
+		});
```
DanTup, line 635: @dbaeumer I'm trying to understand how best to write a test for this. I think I need to:

1. Cause VS Code to call the onDidCreateFiles event (though if this only fires for its UI events, I'm not sure how to that programatically)
2. Either verify that the edit returned by the server was applied by VS Code, or at least that the `waitUntil` method was called on the event and passed the expected edit

Do you have any suggestions on the best way to do these? I couldn't find any similar tests to copy (I'd been using `WillSaveTextDocumentWaitUntil` mostly as a template, though I can't find tests for it - are they named something different?).

Thanks!

dbaeumer: @DanTup I actually assume in the integration tests that VS Code behaves correctly (e.g. if a user requests code complete they call the code complete handler). So what I usually do in the integration tests is to get the feature and then the provider and I then call the methods directly on the provider. See https://github.com/microsoft/vscode-languageserver-node/blob/master/client-node-tests/src/integration.test.ts#L663

You could do the same. It also allows you to test the middleware.

DanTup:
@dbaeumer I've implemented a test for `OnWillCreateFiles` - can you check that what's here seems sane before I implement the others (which are largely a copy/paste of this with Create swapped for Rename/Delete).

I made the test server register statically + dynamically (slightly different paths) so both can be tested together (I don't have middleware tests yet, but they should be easy).

Any preferences for how to test the Did{Create,Rename,Delete} notifications went through? (They don't have a response like the Will{...} requests.

Thanks!

client/src/common/client.ts
```diff
@@ -1196,6 +1210,95 @@ class WillSaveWaitUntilFeature implements DynamicFeature<TextDocumentRegistratio
...
+	}
+
+	public fillClientCapabilities(capabilities: ClientCapabilities): void {
+		let value = ensure(capabilities, 'files')!;
+		value.willCreate = true;
+	}
+
+	public initialize(capabilities: ServerCapabilities, _: DocumentSelector): void {
+		let syncOptions = capabilities.files;
+		if (syncOptions?.willCreate?.globPattern) {
```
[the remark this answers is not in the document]
dbaeumer: Regarding errors: during initialization we should check the pattern, and if not valid ignore the capability and add something to the output channel.

If an invalid pattern is received during dynamic registration we should throw. However we should always throw `new Error` and never a string.

dbaeumer (changes requested):

dbaeumer:
Regarding the test question: you can also throw an error on the server side which will propagate back to the client and will make the send call fail. Normally this will make the test fail.

client/src/common/client.ts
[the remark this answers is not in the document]
DanTup: Done both of these. We throw a custom error type in registration so it can be caught by static registration and ignored/logged instead. Let me know if this should done differently!

DanTup:
> Regarding the test question: you can also throw an error on the server side which will propagate back to the client and will make the send call fail. Normally this will make the test fail.

I think this would prevent checking that calling `waitUntil` is being handled correctly? Or at least, require separate tests (some that throw with the file list, for testing the correct files were sent, and another that returns a known edit that is passed to `waitUntil` to ensure that's being handled/called correctly). Any preference?

client/src/common/client.ts
```diff
@@ -1340,6 +1348,13 @@ class DidSaveTextDocumentFeature extends DocumentNotifications<DidSaveTextDocume
 	}
 }
 
+class InvalidGlobError extends Error {
```
dbaeumer, line 1351: I would leave this out and simply throw new Error. The client will not catch it anyways and the server will not care :-)

DanTup: Done! Though this does mean if another error occurs (that isn't converting the glob to a regex) inside the registration method, it may warn with a misleading message (and would also ignore the error). There's not a lot of code there to go wrong, but thought it worth mentioning.

DanTup:
@dbaeumer any other feedback on this, or is it worth me implementing the other methods too? (they're mostly copy/pastes of this, though with the `did` ones being slightly simpler than the `will` ones as they're just notifications with no responses - though we will need a way to confirm in the test that the server received the notification?).

dbaeumer:
@DanTup to be able to merge this we should have implementation for all notifications and requests. I will add my comments to the one implementation we have and the the rest should be copy paste.

DanTup:
@dbaeumer sgtm! I always planed to implement them all before anything is merged - I just wanted to check you're happy with this one before creating the others. Thanks!

protocol/src/common/protocol.ts
```diff
@@ -350,6 +350,37 @@ export interface WorkspaceClientCapabilities {
 	codeLens?: CodeLensWorkspaceClientCapabilities;
 }
 
+export interface FileClientCapabilities {
```
dbaeumer, line 353: Now that I think about the name we should may be call this FileManipulation to make it clear that these are not file events in the classical sense of file watching. We should also add a sentence to make this clear.

dbaeumer: You used `FileOperations` further down which is a good name.

protocol/src/common/protocol.ts
```diff
@@ -607,6 +638,13 @@ export interface _ClientCapabilities {
 	 * Experimental client capabilities.
 	 */
 	experimental?: object;
+
+	/**
+	 * The client has support for file requests/notifications.
+	 *
+	 * Since 3.16.0
+	 */
+	files?: FileClientCapabilities
```
dbaeumer, line 647: And then we can name this fileOperations as well. Same on the server side.

server/src/common/server.ts
```diff
@@ -1210,6 +1218,18 @@ export interface _Connection<PConsole = _, PTracer = _, PTelemetry = _, PClient
...
+	 * Installs a handler for the `WillCreateFiles` request.
+	 *
+	 * Note that this request is opt-in. The client will not send it unless
+	 * your server has the `files.willCreate` capability,
+	 * or you've dynamically registered for the `workspace/willCreateFiles`
+	 * method.
+	 *
+	 * @param handler The corresponding handler.
+	 */
+	onWillCreateFiles(handler: RequestHandler<CreateFilesParams, WorkspaceEdit | undefined | null, void>): void;
```
dbaeumer, line 1231: I moved away from putting the handlers on the connection namespace since it makes it hard to understand this. Since we already have a `window` namespace and that namespace is use for UI specific requests and notification we should add these to the `window` namespace. This makes it also clear that this has nothing to do with normal file events since those would be on the workspace namespace. All window stuff is UI related.

dbaeumer: This also allows you to implement this on the server side as a feature and not directly in the server since the feature can be added to the RemoteWindow type here https://github.com/microsoft/vscode-languageserver-node/blob/master/server/src/common/server.ts#L470 and to the implementation here: https://github.com/microsoft/vscode-languageserver-node/blob/master/server/src/common/server.ts#L512

I can help with this if it is not clear how to do this.

DanTup: I started doing this, but I've gotten confused. The example I was copying (ShowDocumentFeature) is a server->client request. The ones I'm implementing come from the client. So I don't have a server implementation that's shared, the implementation needs to come from the specific server implementation.

I've pushed my changes so far, there are two places I'm stuck:

1. What the (test) server implementation should look like now (this code doesn't compile):

https://github.com/microsoft/vscode-languageserver-node/pull/687/files#diff-18e38934adbfa54d730ee13bc65f20e81c71d6b5fcc526f1a6c543d9cb2995dbR243

2. What should the Feature look like (when the server code has no default implementation unless it's being contributed by the implementing server):

https://github.com/microsoft/vscode-languageserver-node/pull/687/files#diff-36df91f0cf05683eb4b817da9fe0d3f54e6110d749be36546e3e47ac09f72280R15

protocol/src/common/protocol.ts
DanTup: Done!

protocol/src/common/protocol.ts
DanTup: Done!

(Also pushed changes to the LSP spec PR to reflect the same)

server/src/common/server.ts
dbaeumer: You can copy from an example that implements a language request. It would work the same for the window namespace. A corresponding example is here: https://github.com/microsoft/vscode-languageserver-node/blob/master/server/src/common/onTypeRename.ts#L6

The code for the test server is correct assuming that you have it working for the window namespace. Is your code checked in. Then I can tweak it and push to your repository if you give me write access.

DanTup: Aha, I see! I've gotten further (and pushed my changes), though it seems like I'm missing a `this.attachWorkDoneProgress` from `_RemoteWindowImpl` (it's on `_Languages` already). What's the correct way to add this? (I can copy/paste from `_Languages` but I'm not sure if that's right - I'm also unsure if we should have progress here and if so whether I need some tweaks to the spec?).

I've given your access to my fork btw, so feel free to make changes if that's easier than describing. Thanks!

dbaeumer: OK. I did the changes and pushed them. Did some additional code cleanup One thing I didn't do to not surprise you is to rename protocol.window.fileOperations.ts to protocol.fileOperations. I usually don't put the namespace into the name. A also created a fileOperations.ts file in the client.

Regarding `this.attachWorkDoneProgress` since the request should be fast we shouldn't allow progress. So you use a normal RequestHandler.

DanTup: Great - thanks! I'll rename that file and fill in the rest of the notifications and tests.

DanTup:
@dbaeumer I've implemented the rest of these. I need to do some more testing (and ensure the protocol changes here match the spec exactly), but it should be good for review.

Some specific things I'm unsure of:

- The amount of copy/paste in the features - I wasn't sure if we should try to extract a common base or not (I tried, but it got messy - but my TS typing skills might not be complete :-))
- If I'm validating the results in the tests the best way

Feedback appreciated!

dbaeumer:
@DanTup I did the code restructuring to avoid the coping. I also move the operations back into the workspace namespace. Reason is that these notifications and requests should be called even if a workspace edit manipulates a file. So they are not only UI related. Sorry for the confusion. Can you please do the same for the LSP specification PR.

I also looked at the test and they look to me. I like the way how you assert the notifications (which by the way were sent as requests. I fixed that as well).

DanTup:
@dbaeumer 

> I did the code restructuring to avoid the coping

Great - thanks!

> I also move the operations back into the workspace namespace. [...]. Can you please do the same for the LSP specification PR.

Sure, will do!

> (which by the way were sent as requests. I fixed that as well).

Oops, thanks!

> Can you please check my code changes and then let me know when you are ready so that we can merge that in.

All seems good to me. I made one small tweak to the exception message.

I'm part-way through adding this to the Dart server, so I'll finish that off (+ update the spec to `window/`) and give it a test (I expect to get this all done today) as an additional check that the spec (which I generate all the Dart types/constants from) all matches up with this code here. I'll ping back to confirm if it all looks good (or I hit issues).

Thanks!

DanTup:
Tested `willRenameFiles` using the Dart server (all the interfaces/constants generated from the latest version at https://github.com/microsoft/language-server-protocol/pull/1074) and everything seemed to work as expected.

dart_lsp_rename

Client included capabilities:

```
"fileOperations":{"dynamicRegistration":true,"didCreate":true,"didRename":true,"didDelete":true,"willCreate":true,"willRename":true,"willDelete":true}}
```

Server asked for Dart/folders:

```
"fileOperations":{"willRename":{"globPattern":"**{/,*.dart}"}}
```

Rename sent nothing for a non-Dart file, and the expected request for a `.dart` file and applied the returned edits:

```
Content-Length: 211
{"jsonrpc":"2.0","id":73,"method":"workspace/willRenameFiles","params":{"files":[{"oldUri":"file:///Users/danny/Desktop/nulltest/bin/one.dart","newUri":"file:///Users/danny/Desktop/nulltest/bin/one1111.dart"}]}}

Content-Length: 260
Content-Type: application/vscode-jsonrpc; charset=utf-8
{"id":73,"jsonrpc":"2.0","result":{"documentChanges":[{"textDocument":{"version":11,"uri":"file:///Users/danny/Desktop/nulltest/bin/two.dart"},"edits":[{"range":{"start":{"line":0,"character":7},"end":{"line":0,"character":17}},"newText":"'one1111.dart'"}]}]}}

Content-Length: 287
{"jsonrpc":"2.0","method":"textDocument/didChange","params":{"textDocument":{"uri":"file:///Users/danny/Desktop/nulltest/bin/two.dart","version":12},"contentChanges":[{"range":{"start":{"line":0,"character":7},"end":{"line":0,"character":17}},"rangeLength":10,"text":"'one1111.dart'"}]}}
```

DanTup:
Actually, renames of folders aren't being sent - investigating (I suspect it's because we're assuming a trailing slash on the paths from VS Code, but not getting them).

There are also a couple of TODOs in the code I'm after opinions on (how to handle Windows backslash separators, and a matching question):

https://github.com/microsoft/vscode-languageserver-node/pull/687/files#diff-21298d0f48ea9cffb0b2b7bf3b870536b6122b77510a1f03d7d39cfb6f78f3b5R10

https://github.com/microsoft/vscode-languageserver-node/pull/687/files#diff-21298d0f48ea9cffb0b2b7bf3b870536b6122b77510a1f03d7d39cfb6f78f3b5R99

dbaeumer:
When I looked at the code today I was wondering how you made this work since I couldn't see code that does the actual checking. Now I know why. Your friend should be `vscode.workspace.fs#stat`. See https://github.com/microsoft/vscode/blob/master/src/vs/vscode.d.ts#L6822

DanTup:
@dbaeumer got it! I wasn't sure if calling that here would be ok (since it may hit the disk?), though thinking about it, that's likely negligible to the work the server may do to compute the edits. I'll fix it up, thanks!

The Windows thing is a non-issue, as it was already documented as using forward slashes and uses `.path` from the URI.

DanTup:
@dbaeumer I had a go at this but it's a little more involved than I'd expected so I wanted to check the approach seems sensible.

My plan was to "fix" the URIs to have trailing slashes for directories when the event is first raised. This means `send becomes `async` but also that middleware will always get the "fixed" version of the event (with the trailing `/` on directories).

However, I hit an issue that we can't call `waitUntil` asynchronously, so it may need some more restructuring. I've pushed what I've done so far (though this is broken, VS Code will throw on the `waitUntil` call) if you want to take a look. If you have a better idea than doing this (and reworking it so `waitUntil` isn't called asynchronously) let me know!

Also, as an aside - `addSlashesToDirectories` doesn't function in tests since we don't really have those files on disk, and I'm now wondering what happens if middlewares try to call `waitUntil` (asynchronously)... I'm wondering whether it should even be given to them?

client/src/common/fileOperations.ts
```diff
@@ -43,17 +60,20 @@ abstract class FileOperationFeature<I, E extends { readonly files: ReadonlyArray
...
 	private _listener: code.Disposable | undefined;
 	private _globPatterns: Map<string, RegExp> = new Map<string, RegExp>();
+	protected fixSlashes: (e: E) => Promise<E>;
 
 	constructor(client: BaseLanguageClient, event: code.Event<E>,
 		registrationType: proto.RegistrationType<proto.FileOperationRegistrationOptions>,
 		clientCapability: keyof proto.FileOperationClientCapabilities,
-		serverCapability: keyof proto.FileOperationOptions)
+		serverCapability: keyof proto.FileOperationOptions,
+		fixSlashes: (e: E) => Promise<E>)
```
dbaeumer, line 69: I don't think we need to make this a instance field. It is always the same function we use.

dbaeumer (changes requested):

client/src/common/fileOperations.ts
DanTup: It's slightly different for renames:

```ts
async (params) => ({ files: await Promise.all(params.files.map(addSlashesToDirectories)) })
// vs
async (params) => ({ files: await Promise.all(params.files.map(async (i) => ({ oldUri: await addSlashesToDirectories(i.oldUri), newUri: await addSlashesToDirectories(i.newUri) }))) })
```

I think I have a solution to my async stuff (including an easy way to remove the `waitUntil` from the middleware so it can't try to use it - since it would very likely break things, and it should just be returning the `Promise<WorkspaceEdit>`).

Just doing some more testing and I'll push it up for you to review.

dbaeumer: What I mean is that for example in a rename we only need to transform the oldUri since this is the URI we test again. The new URI shouldn't be transformed. In addition we shouldn't temper with the URI itself. The '/' shouldn't reach the server since URI for folders don't reach the server in that format in all other cases.

DanTup: Ah, got it! I'll fix :-)

DanTup:
@dbaeumer 

Ok, pushed some more changes that I *think* resolve the issues. I changed the middleware from `WillFoo` to just `Foo` events (which are the same, minus the `waitUntil`) so middleware can't try to call `waitUntil` (it's now called at the top level be us, wrapping the function that does the work and returns the `WorkspaceEdit`).

The directory-slash handling is done only during the filtering now, so the URIs provided to the middleware (and server) and unmodified (this slightly complicates things if the middleware wants to know if things are folders too - though I avoided this in the tests be putting trailing slashes in the source URIs for testing).

I need to re-test this in VS Code using the Dart server to ensure everything's working as intended, and also resolve the conflicts (will be tomorrow now). There's still a TODO about matching patterns, but besides those, we might be nearly there this time 😄

Thanks for your help and patience!

client/src/common/fileOperations.ts
dbaeumer: What we can do if necessary is to add a hint to the LSP params whether the URI represents a directory or file.

DanTup:
@dbaeumer 

> What we can do if necessary is to add a hint to the LSP params whether the URI represents a directory or file.

I'm not sure if it's necessary (servers can probably figure it out quite trivially), but if you think we should add it - let me know.

I've merged to master and made some final minor tweaks. I've re-tested the latest with the Dart server changes I was working on, and everything still looks good.

My only outstanding question is the one about the matching:

```ts
// **/folder2/**{/,*.js}
	// TODO(dantup): Do we expect folder2 included? It fails because although
	// ** matches 0 levels, we have an additional slash to match.
	// 'folder1/folder2/',
	'folder1/folder2/file3.js',
	'folder1/folder2/folder3/',
```

Right now, the `**` matches `folder1/` which means the the remaining `folder2` doesn't match the remaining `/folder2` in the path. It can be fixed by using `**folder2` in the pattern, though I'm not sure if this is intuitive. I couldn't come up with a way to tweak the regex to handle that.

dbaeumer:
Regarding the matching. IMO `folder2/` shouldn't be included. The pattern clearly asks for another folder using `**/`. What would be the argument that it should?

DanTup:
@dbaeumer because the path has `folder1` at the start.

It feels like `**` should match `folder1` in `folder1/folder2/`, so `**/folder2/` would match `folder1/folder2/`. It doesn't intuitive that `**` eats the `/`. This makes `***.dart` the correct pattern to match all dart files at any level, and it just seems a bit weird to me (but I'm not very familiar with how other glob patterns like this work, so if it's normal than I'm not objecting :-))

dbaeumer:
Actually I take it back. First IMO the pattern is wrong even for matching a JS file. It should be `**/folder2/**{/,/*.js}` (observe the additional `/` in `/*.js`. The pattern is then equal to `{**/folder2/**/,`**/folder2/**/*.js}`. Since `**` matches zero or more directories `**/folder2/**/` should match `folder1/folder2`. 

If the slash in a pattern gets too substile (and too complicated to parse and convert into a proper regexp) then I think we can also have two different patterns. A file and a directory pattern. The `FileOperationOptions` could then for each property accept an array so that with one registration a server can register n patterns. May be this makes it easier to understand than using a magic `/`. We could then even make use of this when filtering on the client. What do you think?

I also did some minor cleanup of the code (e.g. proper typing). You might want to have a look at it.

DanTup:
> I also did some minor cleanup of the code (e.g. proper typing). You might want to have a look at it.

Some of these changes (switching the events to the `WillXXX` variations) were deliberate, as it didn't seem to make sense to pass the `waitUntil` into middleware (they should never call it, and we must always call it synchronously at the top) - see https://github.com/microsoft/vscode-languageserver-node/pull/687#issuecomment-736744053.

> If the slash in a pattern gets too substile (and too complicated to parse and convert into a proper regexp) then I think we can also have two different patterns. A file and a directory pattern.

If we split the pattern, we could potentially use the `minimatch` package (which the `glob` package uses). The reason we had to skip that was that it didn't support selecting folders/directories (and in order to check for slashes ourselves, we needed to expand things in braces). If we have two patterns and we know whether the input is a file or directory, we can just select the correct pattern and send it through minimatch. Want me to do that?

If so, any preference for how to put the two patterns into `FileOperationRegistrationOptions`?

```ts
globPatterns: Array<{ pattern: string, files: boolean, directories: boolean }>;
globPatterns: Array<{ filePattern?: string, directoryPattern?: string }>;
```

dbaeumer:
> Some of these changes (switching the events away from the WillXXX variations) were deliberate, as it didn't seem to make sense to pass the waitUntil into middleware (they should never call it, and we must always call it synchronously at the top) - see #687 (comment).

That makes sense. I will make the necessary changes. But the feature in the client should be typed correctly.

dbaeumer:
I actually would defined a literal like this:

```
interface FileOperationPatten {
   glob: string;
   matches? 'file' | 'folder'
}
```

If matches is omitted it matches both.

DanTup:
@dbaeumer makes sense to me. I'll switch over to using the minimatch library and drop the pattern parsing stuff and go with your `FileOperationPattern` class from above. Thanks!

dbaeumer:
Made sure that waitUntil doesn't escape to the middle ware.. Please pull

DanTup:
@dbaeumer changes made, and custom pattern parsing has been removed for minimatch. I'll update the spec to match.

DanTup:
I've updated the spec, though when testing with the Dart server, something isn't right - folder renames aren't working. Running out of time, but will track this down tomorrow morning (and figure out why the tests aren't behaving the same).

dbaeumer:
@DanTup did some minor changes. Introduced `FileOperationPatternKind` and updated the third party notice file since we are now using minimatch.

DanTup:
I tracked down my folder issue - turns out if your pattern is `**/` then folders will only match if the path also ends with `/`.

So we have two options:

- We say you should use `**/*` to match all folders
- We allow `**/` to match all folders by putting back the code that appends slashes to folder paths during the match

I presume the latter is probably better so I'll start with that - but should if you think it'd be better the other way.

DanTup:
Ok, sorted. Re-tested agains the Dart server with various combinations of file/folder matching and they all work as expected, and the edits to fix up the imports are working correctly too.

I'll copy your `FileOperationPatternKind` over into the spec PR too, but otherwise as far as I'm aware this is all working and there's nothing outstanding :-)

DanTup:
@dbaeumer thanks!

Do you expect to push a new pre-release build that includes this (and the progress middleware changes) before 3.16 is finalised?

dbaeumer:
@DanTup thanks for providing the PR.

Regarding pre-release: yes, either today or tomorrow. I still need to do some polish and I wanted to better understand the minimatch directory issue to see if we need to spec more.

DanTup:
Great!

Do shout if anything isn't clear or you want me to make changes. I went with the second option above (we suffix `/` on the folder URIs for matching, so both `**/` and `**/*` will match all folders) so I think it should work as expected, though I'm not all that familiar with these glob patterns!

Thanks!

dbaeumer:
@DanTup I played a litte with minimatch and I think we don't need to add the slash to the pattern since we have the pattern kind. I tested that for example:

```js
let matcher = new Minimatch('a/**');
matcher.match('a/folder1/folder2/folder3');
```

evaluates to true.

So I would suggest that we do the following:

- if kind (e.g. folder or file) is undefined we take the pattern as is and match
- if kind is defined then we stat the URI and make sure that the type returned from the stat matches the requested kind. 

This means we still need to do that stat but we don't need to tweak the paths. Is there something you can think of that doesn't make this work?

dbaeumer:
The filter method would then look like this:

```typescript
	protected async filter(event: E, prop: (i: I) => code.Uri): Promise<E> {
		// (Asynchronously) map each file onto a boolean of whether it matches
		// any of the globs.
		const fileMatches = await Promise.all(event.files.map(async (item) => {
			const uri = prop(item);
			const path = uri.fsPath;
			for (const globs of this._globPatterns.values()) {
				for (const pattern of globs) {
					if (pattern.matcher.match(path)) {
						// The pattern matches. If kind is undefined then everything is ok
						if (pattern.kind === undefined) {
							return true;
						}
						const fileType = await FileOperationFeature.getFileType(uri);
						if (fileType === undefined || (fileType === code.FileType.File && pattern.kind === FileOperationPatternKind.file) || (fileType === code.FileType.Directory && pattern.kind === FileOperationPatternKind.folder)) {
							return true;
						}
					}
				}
			}
			return false;
		}));

		// Filter the files to those that matched.
		const files = event.files.filter((_, index) => fileMatches[index]);

		return { ...event, files };
	}

```

dbaeumer:
Tests are still green with that change. One thing we can debate is whether a failure of determining the file type should drop the event or sent it. Currently it is sent.

I have pushed the changes. Let me know if you see any problems with them.

DanTup:
> I played a litte with minimatch and I think we don't need to add the slash to the pattern since we have the pattern kind

The case I was trying to cover was if the pattern ended with a `/`, eg:

```ts
// true
console.log(new Minimatch('a/**').match('a/folder1/folder2/folder3'));
// false
console.log(new Minimatch('a/**/').match('a/folder1/folder2/folder3'));
// true
console.log(new Minimatch('a/**/').match('a/folder1/folder2/folder3/'));
```

The middle one fails there, as the pattern has a trailing slash. I don't know whether that's reasonable or not, but some of my previous example patterns were like:

```
**{/,/*.dart}
```

If it's ok to say there shouldn't be trailing slashes on the patterns for directories (we can't enforce this easily ourselves if we're not doing the brace expansion), then I think the changes are fine. It would make sense to highlight this in the spec if so, to avoid catching people out.

> One thing we can debate is whether a failure of determining the file type should drop the event or sent it. Currently it is sent.

I have no strong feeling on this, though since the server can easily drop it if it can't do anything with it (but can't handle what it doesn't get), maybe sending it makes most sense. Maybe worth being explicit about this in the spec whichever way you think is best?

dbaeumer:
OK. I do think we should cover `console.log(new Minimatch('a/**/').match('a/folder1/folder2/folder3'));`. I will look into fixing it.

dbaeumer:
Here is what I suggest we do: if the matching fails and kind is folder then we stat and if it is a directory we add a '/' to the end and rematch. The reasoning behind this is that the regular expression matching is a lot faster than stating the URI. 

Does that makes sense?

dbaeumer:
See https://github.com/microsoft/vscode-languageserver-node/commit/cd9b8e4baadb6e5df84346f69915defdb44aa85b

DanTup:
I can't think of any examples where that wouldn't work, so seems sensible to me.

The only question I have is whether minimatch cares that the trailing slash is forwards but the rest of the path might be backslashes now we're using `.fsPath`?

dbaeumer:
Minimatch actually doesn't care. But let me check again.

I used fsPath since we do use that in all other cases, even with document selector patterns. So this makes it at least consistent.

DanTup:
> Minimatch actually doesn't care

Great, all sounds good to me then :-)

dbaeumer:
It does take care of it but it depends on the platform :-). Only under Windows it treats `\` correctly and doesn't like '/' in a path. Need to see what we best do here to ensure everything works even with remote file systems.

DanTup:
Oh, I see. Can't we switch back to `Uri.path` and say that patterns always use `/`?

dbaeumer:
Using Uri.path is IMO confusing since VS Code always uses fsPath for patterns. This would result in strange behavior expecially under Windows with drive letters.

I did some more testing and it is actually OK. Under Windows:

```typescript
let matcher = new Minimatch('a/**/')
> matcher.match('a\\folder\\folder\\')
true
> matcher.match('a\\folder\\folder/')
true
>
```

Which is OK. But what I will do after checking with what we do elsewhere is to replace `\\` with `/` in the returned fsPath.

dbaeumer:
Done.

DanTup:
> This would result in strange behavior expecially under Windows with drive letters.

Fair - I was thinking patterns would always be relative since they're implicitly constrained by the workspace, but that probably isn't a valid assumption.

Oh, one final thing - minimatch seems to be case-sensitive by default. I think we should document this, or state that they're case-insensitive and set `nomatch`?

```ts
// true
console.log(new Minimatch('a/**').match('a/folder1/folder2/folder3'));
// false
console.log(new Minimatch('A/**').match('a/folder1/folder2/folder3'));
// true
console.log(new Minimatch('A/**', { nocase: true }).match('a/folder1/folder2/folder3'));
```

Let me know if you want me to do any of that. Thanks!

dbaeumer:
Actually, that is a good point. We should add an options property to `FileOperationPattern` with a property ignoreCase: boolean.

Something like 

```typescript
interface FileOperationPatternOptions {
    ignoreCase: boolean;
}
```

Does that make sense to you ?

DanTup:
Yep! I can update the spec/client tomorrow if you'd like.

Should the options and/or `ignoreCase` be optional? And if so, what should it default to?

dbaeumer:
I made it optional. Looks like this

```
export interface FileOperationPatternOptions {

	/**
	 * The pattern should be match ignoring casing.
	 */
	ignoreCase?: boolean;
}
```

Default should be false as with all optional values.

dbaeumer:
I update the spec as well.

DanTup:
I added a comment here about a typo + making the default clear from the description:

https://github.com/microsoft/vscode-languageserver-node/commit/943b2db291b8e958a793fce0ab2bbe73c6afc94d#r44910295

But otherwise, all looks good to me :-)

dbaeumer:
Published new next versions.

DanTup:
Thanks! I rebuilt everything from the latest version of the spec in the Dart server, and my renames are all working as expected! 😀

Looking at the stuff in the v3.16 milestone here and for LSP, I presume there aren't any significant changes planned before it's finalised? I'd like to start preparing PRs for the server changes (for this + semantic tokens) but have been holding off until the spec was less likely to change.

Thanks again for your help + patience with this!

dbaeumer:
Thanks a lot for all your help with this. I don't plan to add any new features for 3.16. As soon as VS Code 1.52 ships I will release a new LSP version.

DanTup:
Perfect, thanks!

Out of interest - is there anything specific in VS Code 1.52 that is required for v3.16? I'm wondering whether I should be increasing the minimum version of VS Code when I switch to this latest client or not - I haven't noticed any issues using it with v1.51, but my testing has been quite focused on the changes in this PR.

dbaeumer:
The new version of LSP will require 1.52 because of the linked range editing support.

DanTup:
Ah, I see - I'll bump my minimum version once VS code releases then too. Thanks!

PR #697, opened 2020-12-03
Use appName as client info instead of 'vscode'

henryju:
Using `env.appName` will return `Visual Studio Code` or `VSCodium` for example (and I hope another label for GitHub Codespaces). This is more useful IMHO than hardcoding `vscode`.


dbaeumer:
Makes sense.

PR #715, opened 2020-12-25
Fix hang when server crashes during shutdown

strager:
LanguageClient#stop sends a shutdown request to the language server. If
the server crashes/exits before sending a response, the promise returned
by LanguageClient#stop never settles. This happens because the
disconnect event is ignored during shutdown.

Fix the hang by not ignoring disconnects during shutdown. This will
cause LanguageClient#stop to throw an exception ("Connection got
disposed") if the server crashes or exits unexpectedly during shutdown.


dbaeumer:
Thanks a lot!.

PR #719, opened 2020-12-30
fix unknow parameter structure issue

OliverwengFiltered:
this is the fix for:
 Error: Unknown parameter structure byName
    at computeSingleParam (connection.js:776)
    at computeMessageParams (connection.js:787)
    at Object.sendRequest (connection.js:929)
    at Object.initialize (connection.js:20)
    at MonacoLanguageClient.doInitialize (client.js:2151)
    at MonacoLanguageClient.initialize (client.js:2147)
    at eval (client.js:2095)

### DIFF
--- a/jsonrpc/src/common/connection.ts
+++ b/jsonrpc/src/common/connection.ts
@@ -1046,7 +1046,7 @@ export function createMessageConnection(messageReader: MessageReader, messageWri
 	}
 
 	function computeSingleParam(parameterStructures: ParameterStructures, param: any): any | any[] {
-		switch(parameterStructures.kind) {
+		switch(parameterStructures.toString()) {
 			case 'auto':
 				if (isNamedParam(param)) {
 					return nullToUndefined(param);
--- a/jsonrpc/src/common/connection.ts
+++ b/jsonrpc/src/common/connection.ts
@@ -1046,20 +1046,20 @@ export function createMessageConnection(messageReader: MessageReader, messageWri
 	}
 
 	function computeSingleParam(parameterStructures: ParameterStructures, param: any): any | any[] {
-		switch(parameterStructures) {
-			case ParameterStructures.auto:
+		switch(parameterStructures.kind) {
+			case 'auto':
 				if (isNamedParam(param)) {
 					return nullToUndefined(param);
 				} else {
 					return [undefinedToNull(param)];
 				}
 				break;
-			case ParameterStructures.byName:
+			case 'byName':
 				if (!isNamedParam(param)) {
 					throw new Error(`Received parameters by name but param is not an object literal.`);
 				}
 				return nullToUndefined(param);
-			case ParameterStructures.byPosition:
+			case 'byPosition':
 				return [undefinedToNull(param)];
 			default:
 				throw new Error(`Unknown parameter structure ${parameterStructures.toString()}`);


dbaeumer:
@OliverwengFiltered can you please explain why you think this is necessary. The `ParameterStructures` are singletons and you should not create them outside. All my test cases work as expected. 

If you need to create a `ParameterStructures` from a string we should add a from method to `ParameterStructures`.

OliverwengFiltered:
@dbaeumer 

so this is for achieving a better compatibility. I am using a webpack compiled it into a bundled.js file.

and with latest and greatest language server relevant packages installed, it seems it's producing an error on that line even if it is a "byName" kind. see below gif.

Kapture 2021-01-04 at 14 23 38

dbaeumer:
I think web packing shouldn't change the semantic of a case statement and I actually webpack this for the ESLint extension as well. So it is either a problem if the execution environment or the parameter structure doesn't relate to one of the three singletons. How does the parameter structure come into the system in our example.

dbaeumer:
Coould it be that in our web pack file the class `ParameterStructures` exists multiple times?

OliverwengFiltered:
@dbaeumer 

emmm, the executing environment is just latest chrome.
and I don't see any additional ParameterStructures in this node_module: "vscode-jsonrpc"

here is the package zip I'm using.

[monaco-jsonrpc.zip](https://github.com/microsoft/vscode-languageserver-node/files/5771432/monaco-jsonrpc.zip)

dbaeumer:
I looked at it and I don't see any either.

Can you craft me a GitHub repository I can clone that demos what you are seeing. We need to understand why this singleton fails for you since we use it in other situations as well.

OliverwengFiltered:
Hi @dbaeumer 
so I have crafted a simple demo repo for reproducing the issue. here is the link: https://github.com/OliverwengFiltered/monaco-lsp-demo

steps to reproduce.
1. pull the code base.
2. `npm install`
3. `npx webpack`
4. `npm start`
5. go to browser and hit http://localhost:3000
6. click button "init monaco with language server"
7. check out browser console.

dbaeumer:
@OliverwengFiltered thanks for the test case. I looked into it and you do ship vscode-jsonrpc two times which results in two instances of the `ParamterStructures` class which makes the code incompatible. 

On instance come through 

capture

And one instance through

capture

And some of the calls go through the first one 

capture

and others through the second one hence resulting in the behavior you are seeing.

Such a setup is actually not supported (this is why protocol ships and re-exports vscode-jsonrpc) since the jsonrpc layer keeps quite some state (e.g. open request, ....)

Why is this necessary at all? The latest jsonrpc version has support for browser so IMO this could be achieved a lot easier.

rcjsuen:
@dbaeumer I am getting hit by this also (https://github.com/microsoft/vscode-docker/issues/2601).

Sounds like version discrepancies between the client and the server that's causing two different versions of `vscode-jsonrpc` to be needed...I guess?

dbaeumer:
@rcjsuen can you outline this. Client and server are running in separate processes so it is OK to have two separate versions of `vscode-jsonrpc`. You simply can't have two separate version in the client or the server.

rcjsuen:
> @rcjsuen can you outline this. Client and server are running in separate processes so it is OK to have two separate versions of `vscode-jsonrpc`. You simply can't have two separate version in the client or the server.

@dbaeumer That was just a guess. What you described about them running in separate processes makes sense so not sure why this is happening for the Docker extension then... :|

dbaeumer:
@rcjsuen I commented on https://github.com/microsoft/vscode-docker/issues/2601. Let me know if that helps.

dbaeumer:
Fixing this in the vscode-jsonrpc by a more forgiving equal test might make the error go away but the fact the vscode-jsonrcp is bundled two times might have other strange side effects.

dbaeumer:
Actually now I remember why this is a class and not simply and enum / string. It has something to do with signature overloads. Without this being a special type I couldn't distinguish this from a normal string parameter with a value `byName` or one of the other values.

OliverwengFiltered:
Hi @dbaeumer 

that makes sense now, let me use webpack to override the package version to load from one version only.

Thanks @dbaeumer

sharlatta:
I've run into a similar issue. 

I have a project that uses `vscode-languageclient` (client) and another one (server) that uses the `vscode-languageserver`. They both rely on `vscode-jsonrpc` lib for LSP communication.  My server package is an `npm` dependency  for the `client`.  When I've linked them locally with `npm link` the server was using `vscode-jsonrpc` lib from its `node_modules` whereas the client has its own instance of the library. When I created a new `RequestType` , its `_parameterStructures` is set to the `ParametersStructures.auto`.



When the client calls `sendRequest` with the `RequestType` defined on the server side, there is some code that checks request's object `_parameterStructures` value. And that's where it is comparing the static property of the `ParameterStructures` class from one location (`vscode-jsonrpc` lib in client't `node_modules`) with the similar property in the other location (`vscode-jsonrpc` lib in server's  `node_modules`).



And because they reference different objects, I get an error.

I was able to work fix this issue with `yalc` https://github.com/wclr/yalc. Unlike `npm link` it does not create a symlink to the folder but instead  grabs only files that should be published to NPM and puts them in a special global store. Thus the "linked" package does not have its own `node_modules` and when the dependency (`vscode-jsonrpc`) in the LS is resolved it is the same one that is being used  by the client .

Leaving it here in case someone else faces a similar issue. 

Also, `yalc` does not seem actively supported so if there are other ideas - I'd like to follow them.

PR #720, opened 2021-01-03
Make the moniker capability optional as per the spec

illright:
According to the [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize), the `moniker` property of the `TextDocumentCapabilities` may be undefined as all the other capabilities. Perhaps, the omission of the optionality marker here is a typo.

### DIFF
--- a/protocol/src/common/protocol.ts
+++ b/protocol/src/common/protocol.ts
@@ -519,7 +519,7 @@ export interface TextDocumentClientCapabilities {
 	 *
 	 * @since 3.16.0
 	 */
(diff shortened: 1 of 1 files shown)


dbaeumer:
@illright thanks. Good catch.

masad-frost:
Can we please get a 3.16.1 version published for the package?

dbaeumer:
There are already a couple of next version for 3.17 which will come soon (planned for end of April / beginning of May).

Issue #713, opened 2020-12-21
client/registerCapability sent for workspace/didChangeWorkspaceFolder even if dynamicRegistration set to false during initialization


mjlbach:
Moving from https://github.com/microsoft/pyright/issues/1302

**Describe the bug**

We have support for pyright as one of the language servers for neovim's built in language client which uses vscode-languageserver-node. One issue seems to be cropping up on node 15 where failed callbacks result in node terminating the process. The main callback that is failing is client/registerCapability, (our client doesn't support dynamic registration). When we report dynamicRegistration as false, pyright is still sending client/registerCapability requests

```
[ DEBUG ] 2020-12-20T13:55:08-0800 ] /home/michael/neovim/share/nvim/runtime/lua/vim/lsp/rpc.lua:489 ]	
"decoded"	{  id = 0,  jsonrpc = "2.0",  method = "client/registerCapability",  params = {    registrations = { {        
id = "807eedda-04cf-4637-9cb7-81d5f5d0e225",        method = "workspace/didChangeWorkspaceFolders",       
 registerOptions = vim.empty_dict()      } }  }}
```
We're initializing the server with all dynamicRegistration capabilities registered as false

<details open>
<summary>json of initialization params</summary>

```json
{
  "initialize_params": {
    "capabilities": {
      "callHierarchy": {
        "dynamicRegistration": false,
        "textDocument": {
          "codeAction": {
            "codeActionLiteralSupport": {
              "codeActionKind": {
                "valueSet": [
                  "",
                  "Empty",
                  "QuickFix",
                  "Refactor",
                  "RefactorExtract",
                  "RefactorInline",
                  "RefactorRewrite",
                  "Source",
                  "SourceOrganizeImports",
                  "quickfix",
                  "refactor",
                  "refactor.extract",
                  "refactor.inline",
                  "refactor.rewrite",
                  "source",
                  "source.organizeImports"
                ]
              }
            },
            "dynamicRegistration": false
          },
          "completion": {
            "completionItem": {
              "commitCharactersSupport": false,
              "deprecatedSupport": false,
              "documentationFormat": [
                "markdown",
                "plaintext"
              ],
              "preselectSupport": false,
              "snippetSupport": false
            },
            "completionItemKind": {
              "valueSet": [
                1,
                2,
                3,
                4,
                5,
                6,
                7,
                8,
                9,
                10,
                11,
                12,
                13,
                14,
                15,
                16,
                17,
                18,
                19,
                20,
[...]

PR #721, opened 2021-01-05
Remove did change workspace folders dynamic registration

bmewburn:
#713


dbaeumer:
@bmewburn removing this defeats the whole purpose of the listener. If a server doesn't want to be notified about workspace folder events since the client don't support them it should not call the method. The listener / event code is helper code to make the registration easier. I am closing the PR because without registering we should remove the whole code.

dbaeumer:
Actually thinking about it again we should guard this against the server capability to support the auto registration.

dbaeumer:
See comment https://github.com/microsoft/vscode-languageserver-node/issues/713#issuecomment-754510832. This requires a different change so. So I will close the PR since not registering at all makes the code wrong in other cases.

bmewburn:
No problem, thanks for reviewing @dbaeumer

PR #736, opened 2021-02-03
[Proposal] Document on Paste Formatting

333fred:
This adds a new proposed LSP endpoint, `textDocument/onPasteFormatting`, which follows the same general structure as the `onTypeFormatting` and `rangeFormatting` features, originally discussed here: https://github.com/microsoft/language-server-protocol/issues/767. I opted to go for the separate endpoint for on-paste formatting, as I don't believe fixing imports is the only thing that an on-paste handler would want to: it could, for example, try to figure out if the content that was pasted in is in a string and handle some escaping of special characters. Therefore, the separate request seemed the better approach to my mind. I could also see a context on a `rangeFormatting` request that mentions "this range was pasted in" working as well, but to be frank my knowledge of the codebase and typescript is not up to adding a new context to an existing type, so I'd need help implementing that in the client.

@dbaeumer, as it is I've made a stab at implementing this in the client by basing the design off the DocumentOnTypeFormattingFeature, but the issue I've run into is that that feature is mainly a delegation to the native vscode implementation. I've added stubs to this PR, but they obviously don't work (and/or don't compile because they have no bodies). What would the next step here be? Implement the feature in vscode? Implement the feature by essentially just forwarding to the rangeFormatting feature?


dbaeumer:
@333fred thanks for the PR. As you noticed yourself this highly depends on how the editor itself want to implement this. I reached out to the VS Code team to see if there is any discussion around this and they pointed me to https://github.com/microsoft/vscode/issues/30066. I think we should learn from their experience and co-ordinate and effort with them.

PR #744, opened 2021-02-11
Cache file system types in Will{Delete,Rename} handlers so they're available in Did{Delete,Rename} for filtering

DanTup:


dbaeumer:
@DanTup sorry for not looking into this but I was completely absorb with the new pull model for diagnostics to overcome the problem that servers can't distinguish between a file being open and being visible in the client. I will start looking into the PR in the week from 1st of Match.

DanTup:
> completely absorb with the new pull model for diagnostics to overcome the problem that servers can't distinguish between a file being open and being visible in the client

I feel your pain ;) This still causes me issues (and in some cases turns small issues into larger issues), especially around Ctrl+Click on symbols, which (due to the hover preview) sends the server "Open fileA, Close fileA, Open fileA".

Out of interest - have you come up with any ideas/solution that might be useful in other places? (or is it finally time for VS Code to get a fix? 😀).

> I will start looking into the PR in the week from 1st of Match.

👍

dbaeumer:
The current proposed implementation is to have an explicit pull of diagnostics. In the proposal it is implemented using some new proposed API from VS Code.

client/src/common/fileOperations.ts
```diff
@@ -221,6 +225,38 @@ abstract class NotificationFileOperationFeature<I, E extends { readonly files: R
 	protected abstract doSend(event: E, next: (event: E) => void): void;
 }
 
+abstract class CachingNotificationFileOperationFeature<I, E extends { readonly files: ReadonlyArray<I>; }, P> extends NotificationFileOperationFeature<I, E, P> {
+	protected _willListener: code.Disposable | undefined;
+	private _fsPathFileTypes: { [key: string]: code.FileType } = {};
```
dbaeumer, line 230: This is an ever growing cache which we should definitely avoid. When the actual did event fires we should clear the cache. In addition we should make this a normal ES6 Map.

client/src/common/fileOperations.ts
```diff
@@ -249,6 +285,22 @@ export class DidRenameFilesFeature extends NotificationFileOperationFeature<{ ol
...
 
+	public register(data: RegistrationData<proto.FileOperationRegistrationOptions>): void {
+		if (!this._willListener) {
+			this._willListener = code.workspace.onWillRenameFiles(this.willRename, this);
+		}
+		super.register(data);
+	}
+
+	private willRename(e: code.FileWillRenameEvent): void {
+		// Calling filter will force the matching logic to run. For any item
```
dbaeumer, line 296: We shouldn't make the will rename wait on this. Instead we should cache promises and make the actual did wait for the promise to resolve. This simply slows down the VS Code operation for no need.

dbaeumer (changes requested):

client/src/common/fileOperations.ts
DanTup: Is it guaranteed that these events can't overlap? (eg. clearing the cache in `did` can't impact other concurrent calls to these events?).

client/src/common/fileOperations.ts
DanTup: Similar to above - can there be concurrent requests? Will we need to track multiple promises, or can we just track one? (if it's called again, we could always replace the promise with another chained into the end - that would avoid any issues if there were concurrent requests though could result in the earlier requests waiting for the later requests depending on timing).

client/src/common/fileOperations.ts
dbaeumer: Since it is all async I guess they theoretical could. But I think it will never happen in practice

client/src/common/fileOperations.ts
dbaeumer: I would argue that we can track the last assuming that it reflects the current state in the file system.

DanTup: > We shouldn't make the will rename wait on this. Instead we should cache promises and make the actual did wait for the promise to resolve. This simply slows down the VS Code operation for no need.

I had a go at this, but I'm seeing failures and I don't think this is possible. If we don't _block_ VS Code from performing the operation while we read the types, the files may be moved/deleted _before_ our async work to get the file types has completed and we'll fail to get them.

For rename, we could perhaps work around it by using the _new_ locations, but that wouldn't be consistent/possible with delete.

(I've done the other changes to clear cache + change to Map though).

dbaeumer:
@DanTup looked at the code. Is it ready for another review. I ask since it is still caching the result and not the promise producing the result to un-block the wait call from VS Code?

DanTup:
@dbaeumer I mentioned this in the comment at https://github.com/microsoft/vscode-languageserver-node/pull/744#discussion_r587552499.

Unblocking VS Code causes it to immediately perform the rename/delete which then "breaks" the promise running asynchronously to fetch the file types. My understanding is that we _need_ to block VS Code while we read the file types, because otherwise we're back at square one where the items as missing from disk and we can't get them?

DanTup:
@dbaeumer any thoughts on the above?

PR #764, opened 2021-05-10
Add more info to client not ready error

yanmofeixi:
I'm having trouble debugging my 'Language client is not ready yet'. Adding the method to help debug which request has caused this error.


dbaeumer:
Thanks.

A small tip: you should only send requests after client.onReady() got fullfilled.

PR #772, opened 2021-05-29
Add proposed textDocument/inlayHints to protocol & client.

sam-mccall:
This addresses the FR https://github.com/microsoft/language-server-protocol/issues/956
And corresponds to the spec in https://github.com/microsoft/language-server-protocol/pull/1249

### DIFF
--- a/client/src/common/commonClient.ts
+++ b/client/src/common/commonClient.ts
@@ -55,10 +55,13 @@ export abstract class CommonLanguageClient extends BaseLanguageClient {
 // Exporting proposed protocol.
 
 import { DiagnosticFeature } from './proposed.diagnostic';
+import { InlayHintsFeature } from './proposed.inlayHints';
+
 export namespace ProposedFeatures {
 	export function createAll(_client: BaseLanguageClient): (StaticFeature | DynamicFeature<any>)[] {
 		let result: (StaticFeature | DynamicFeature<any>)[] = [
-			new DiagnosticFeature(_client)
+			new DiagnosticFeature(_client),
+			new InlayHintsFeature(_client),
 		];
 		return result;
 	}
--- a/client/src/common/proposed.inlayHints.ts
+++ b/client/src/common/proposed.inlayHints.ts
@@ -1,0 +1,92 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import { languages as Languages, Disposable, TextDocument, Range, ProviderResult, InlayHintKind, InlayHint as VInlayHint, InlayHintsProvider } from 'vscode';
+
+import {
+	ClientCapabilities, CancellationToken, ServerCapabilities, DocumentSelector, Proposed
+} from 'vscode-languageserver-protocol';
+
+import { TextDocumentFeature, BaseLanguageClient, Middleware } from './client';
+import * as p2c from './protocolConverter';
+
+function ensure<T, K extends keyof T>(target: T, key: K): T[K] {
+	if (target[key] === void 0) {
+		target[key] = {} as any;
+	}
+	return target[key];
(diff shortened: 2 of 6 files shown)


sam-mccall:
@dbaeumer I'm guessing that as a new feature, this should all go under "proposed"?
Does this mean I should fold the spec text (i.e. https://github.com/microsoft/language-server-protocol/pull/1249) into this PR as a `protocol.inlayHints.proposed.md` file?

dbaeumer:
@sam-mccall yes, this all has to go under proposed. And it is easier to keep the markdown spec part in sync if it is part of this repository as well. Adding it to the spec can be done when we think that the proposed API has reach some stability.

sam-mccall:
@dbaeumer Gentle ping to get this back on your radar :-)

We've got a server implementation of this in clangd which is coming up to a release cycle, and are looking to add matching support in a client soon. Having the protocol landed in some form (proposed, or even just rough agreement on the protocol structure) would let us align on that. IME it's hard to rip out nonstandard extensions later especially if they're "richer" than the standard replacement.

dbaeumer:
I was out on vacation. I will try to find out where VS Code is with that. What we can do is to add something that mirrors VS Code's proposed API. But then the proposed API might change if the VS Code API changes.

dbaeumer:
VS Code will finalize the API this month. After that I think it is a good time to look into the PR again.

rami3l:
@dbaeumer Gentle reminder: https://github.com/microsoft/vscode/issues/16221#issuecomment-1033827087 is now closed.

dbaeumer:
Anyone interested in preparing a new PR or updating this one? Based on reviews we did we changed the VS Code API to allow more features (like navigating on inlay hints) which resulted in changes in the API.

sam-mccall:
I can take another pass at it but probably for a few weeks.

If someone else wants to jump on it before then, please do!

dbaeumer:
A first draft version is available here: https://github.com/microsoft/vscode-languageserver-node/blob/582ea52183d8a7c56fa617974de5cce8a322f0f7/protocol/src/common/proposed.InlayHints.ts#L6

fwcd:
@dbaeumer Thanks for the nice work, just a question:

https://github.com/microsoft/vscode-languageserver-node/blob/37cf23b995720b53a19553c60c15a6fb04536a59/protocol/src/common/proposed.inlayHint.ts#L254

In the proposed LSP API, the range parameter of `InlayHintParams` is called `viewPort` whereas both the [new VSCode API](https://github.com/microsoft/vscode/blob/778b74a2159983ba09568d069ef9c9ecd858f783/src/vscode-dts/vscode.d.ts#L4719) and [`rust-analyzer`](https://github.com/rust-analyzer/rust-analyzer/blob/3f4a095d3e8166d710919e06e21fc4cbc056b0ee/editors/code/src/lsp_ext.ts#L105) call it `range`. Any reason for not picking `range` for LSP too?

matklad:
Yeah, `viewPort` would also be inconsistent with LSP's semantic highlighhting: https://github.com/microsoft/vscode-languageserver-node/blob/37cf23b995720b53a19553c60c15a6fb04536a59/protocol/src/common/protocol.semanticTokens.ts#L230

fwcd:
Another slight inconsistency is `textDocument/inlayHint` (singular) in the current LSP proposal vs. `textDocument/semanticTokens` (plural). A lot of other `textDocument` requests are singular too though, so perhaps naming the request in singular is more idiomatic?

dbaeumer:
I called it `viewPort` since this is how we called it in the inline values. I chatted with @jrieken that we should also make it consistent in the VS Code API. 

In general all request are singular. The semantic token one slipped in and it was too late to correct it since it already moved out of proposed and I didn't want to break it.

dbaeumer:
If we choose range we could make it consistent since inline values is still proposed.

dbaeumer:
OK. We will use range since a client can ask for a range that is not the viewPort.

dbaeumer:
I will close the PR. The feature has shipped in 3.17

PR #776, opened 2021-06-19
Ensure exit notification is sent before closing connection

### LINKED ISSUE #726: Orphaned server processes when VS Code client process is killed
I've been trying to figure this out for a while, and I'm running out of ideas. I've read through this related thread: https://github.com/microsoft/vscode/issues/35196

When I start the server from the client, I use a `withProgress` to monitor the progress. If close the VS Code client window right after the client start process begins, before `client.onReady()` returns, the server process is left orphaned/running.

```typescript
    vscode.window.withProgress({
        location: vscode.ProgressLocation.Notification,
        title: "my title",
        cancellable: false
    }, (_progress, _token) => {
        return client.onReady();
    });
    client.start();
```

My extension's `deactivate` function looks like this:

```typescript
function deactivate(): Promise<void> | undefined {
    // ... other unrelated cleanup code
    if (lspClient) {
        return lspClient.stop();
    }
}
```

I've also made sure not to add the returned disposable from `client.start()` to the extension context subscriptions as noted in the linked thread.

If I wait for the client to fully start, then close the VS Code window, the server is shutdown fine. The problem is only when VS Code is closed right after the server begins to start. I see in the server logs that `shutdown` is called in this case, but `exit` is never called. It seems the VS Code client process is terminated before the response from `shutdown` is received, therefore `exit` is never called.

The server's shutdown method isn't doing anything time-consuming. My only other suspicion is that the server's startup functionality causes the incoming `shutdown` request's processing to be delayed, so while `shutdown` doesn't take long to return, it may be a while before it's called. I have no proof of this though, just a guess.

Is there anything else I can try from the client's perspective?

dsherret:
Currently the exit notification is not sent because the connection gets closed before the message is sent (at least on my machine in WSL). This change waits for the exit notification to be sent before closing the connection.

Fixes #726 (exit notification not received part).

### DIFF
--- a/client/src/common/client.ts
+++ b/client/src/common/client.ts
@@ -3346,7 +3346,7 @@ export abstract class BaseLanguageClient {
 		// unhook listeners
 		return this._onStop = this.resolveConnection().then(connection => {
 			return connection.shutdown().then(() => {
-				connection.exit().then(() => {
+				return connection.exit().then(() => {
 					connection.end();
 					connection.dispose();
 					this.state = ClientState.Stopped;
--- a/client/src/common/client.ts
+++ b/client/src/common/client.ts
@@ -97,13 +97,13 @@ interface Connection {
 	onRequest<R, E>(method: string, handler: GenericRequestHandler<R, E>): Disposable;
 	onRequest<R, E>(method: string | MessageSignature, handler: GenericRequestHandler<R, E>): Disposable;
 
-	sendNotification<RO>(type: ProtocolNotificationType0<RO>): void;
-	sendNotification<P, RO>(type: ProtocolNotificationType<P, RO>, params?: P): void;
-	sendNotification(type: NotificationType0): void;
-	sendNotification<P>(type: NotificationType<P>, params?: P): void;
-	sendNotification(method: string): void;
-	sendNotification(method: string, params: any): void;
-	sendNotification(method: string | MessageSignature, params?: any): void;
+	sendNotification<RO>(type: ProtocolNotificationType0<RO>): Promise<void>;
+	sendNotification<P, RO>(type: ProtocolNotificationType<P, RO>, params?: P): Promise<void>;
+	sendNotification(type: NotificationType0): Promise<void>;
+	sendNotification<P>(type: NotificationType<P>, params?: P): Promise<void>;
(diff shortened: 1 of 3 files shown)


dbaeumer:
Good catch. Thanks for the PR.

bknaysi:
Hello @dbaeumer and @dsherret, the fix doesn't appear to be in the [7.0.0 client release](https://github.com/microsoft/vscode-languageserver-node/blob/release/client%2F7.0.0/client/src/common/client.ts) or [7.1.0.next.5](https://github.com/microsoft/vscode-languageserver-node/commits/release/client/7.1.0-next.5/client/src/common/client.ts) tag. When do you think there will be a client release at version 7.0.0 or higher containing this fix?

dbaeumer:
The fix is in the 8.x-next.y release of the library

bknaysi:
@dbaeumer Thanks for the reply. It looks like `8.x-next.y` [requires](https://github.com/microsoft/vscode-languageserver-node#3170-next-protocol-700-next-json-rpc-800-next-client-and-800-next-server) the `3.17.0` LSP Protocol. However, there doesn't exist an [LSP4J release](https://github.com/eclipse/lsp4j#supported-lsp-versions) supporting it yet, which we use to implement our server. Would it be possible for our client to use `8.x-next.y` while our server only supports the `3.16.0` LSP Protocol? I'm not sure which direction the backwards compatibility exist. For example, could a client on an older protocol work with a server on a newer protocol? Likewise, could a server on an older protocol work with a client on a newer protocol?

If the later is not true, it would be amazing if this exit notification fix could be added to the future `7.1.0` release. I'm not sure we'll want to adopt any transitory `next` releases, but we can wait until another official one.

dbaeumer:
I think the 8.x-next.y client should be able to talk to a 3.16 server. The protocol is compatible if the server correctly handles capability flags.

bknaysi:
I think we'll wait for the official `8.0.0` release before adopting it into our product, but it's good to know they should be compatible. In general, how stable are the intermittent `-next` releases?

dbaeumer:
I usually adopt them in ESLint. The proposed feature might still change. All the other stuff is fairly stable.

PR #779, opened 2021-06-22
Add type hierarchy support

Eskibear:
This is a **prototype** to support type hierarchy proposed in https://github.com/microsoft/language-server-protocol/pull/1231

Because there is no implementation of type hierarchy providers and related APIs in VS Code, I have to implement a simple one here, immitating provider registration in VS Code core.

Changes are mainly two parts:
- Type hierararchy feature, similar with other features e.g. call hierarchy
- ~Implemetation of type hierarchy providers, which should eventually be moved to (re-implemented in) vscode core.~ already available in `vscode.proposed.d.ts` since v1.59

Status of this PR:
- [x] basic implementation
- [x] add integration test
- [x] verify E2E test for Java. (type hierarchy providers in vscode 1.59, peek view in vscode 1.60)
- [x] verify if it works with multiple LSs at the same time. 

~I'm still working on it. The draft PR is created for early review of the feasibility.~ Any critical problem or obvious mistake about the whole workflow, please let me know.


Eskibear:
Type hierarchy provider is already in vscode, I'm going to refine this PR based on that. 

One question is, do I have to move TypeHierarchy into proposed namespace (same as Diagnostics) or just keep it as is?

dbaeumer:
@Eskibear it has to go to the proposed namespace fist although it is available in VS Code.

dbaeumer:
@Eskibear in which state is this PR. It is marked as a draft and it does have duplicate files (e.g. server typeHierarchy.ts and proposed.typeHierarchy.ts)

Are you planning on finishing the PR. I think we can merge it this month and create a new LSP release to contain it.

Eskibear:
It's almost done. I just want to do some E2E verification before I mark it ready to review. 

BTW is there any convenient way to create a private npm package for testing purpose? Due to the dependencies (client -> protocol -> types), previously to verify it E2E, I have to bump versions and create npm packs one-by-one, install them locally... and change the relative path to version number accordingly.

Eskibear:
I just tested with a mock LS for plaintext. Both "Show Type Hierarchy" and "Peek Type Hierarchy" work fine. 

The only problem is, when I use `languageClient.registerProposedFeatures();` to enable typehierarchy support, there's an error from proposed.diagnostic , tracked in #811

Eskibear:
E2E smoke test against Java LS:

https://user-images.githubusercontent.com/2351748/133284770-a9b99130-d38f-4867-9fc6-006419aa89ec.mp4

dbaeumer:
Cool. Thanks for providing this.

protocol/src/common/protocol.ts
```diff
@@ -3283,8 +3304,12 @@ export {
 	DidRenameFilesNotification, RenameFilesParams, FileRename, WillRenameFilesRequest,
 	DidDeleteFilesNotification, DeleteFilesParams, FileDelete, WillDeleteFilesRequest,
 	// Monikers
-	UniquenessLevel, MonikerKind, Moniker, MonikerClientCapabilities, MonikerOptions, MonikerRegistrationOptions, MonikerParams, MonikerRequest
-
+	UniquenessLevel, MonikerKind, Moniker, MonikerClientCapabilities, MonikerOptions, MonikerRegistrationOptions, MonikerParams, MonikerRequest,
+	// Type Hierarchy
+	TypeHierarchyClientCapabilities, TypeHierarchyOptions, TypeHierarchyRegistrationOptions,
```
dbaeumer, line 3309: These shouldn't be here right now. The exports from protocol are re-exported in api so this makes them accessible as normal API now. We should only access them via Proposed (even us internally in the client for example) until they API is final (out of proposed).

dbaeumer (changes requested):
The rest looks good to me

protocol/src/common/protocol.ts
Eskibear: updated.

dbaeumer (approved):

PR #787, opened 2021-07-10
Fix "A request has failed" dialog that can not be dismissed

### LINKED ISSUE #786: "A request has failed. See the output for more information." dialog always shows output channel, even when being dismissed
When a request fail, typically because of an error on the language server side, client shows a popup offering to go to output channel to see the error:

image

Problem is when user close the dialog, this open the output channel anyway.

--- 

I prepared a reproduction if necessary:
 - use https://github.com/microsoft/vscode-extension-samples/tree/main/lsp-embedded-language-service
 - apply a patch so that the server always error on completions, for example:
```patch
diff --git a/lsp-embedded-language-service/server/src/server.ts b/lsp-embedded-language-service/server/src/server.ts
index 4093561..d9805d4 100644
--- a/lsp-embedded-language-service/server/src/server.ts
+++ b/lsp-embedded-language-service/server/src/server.ts
@@ -85,18 +85,7 @@ async function validateTextDocument(textDocument: TextDocument) {
 }
 
 connection.onCompletion(async (textDocumentPosition, token) => {
-	const document = documents.get(textDocumentPosition.textDocument.uri);
-	if (!document) {
-		return null;
-	}
-
-	const mode = languageModes.getModeAtPosition(document, textDocumentPosition.position);
-	if (!mode || !mode.doComplete) {
-		return CompletionList.create();
-	}
-	const doComplete = mode.doComplete!;
-
-	return doComplete(document, textDocumentPosition.position);
+	throw new Error("💥");
 });
 
 // Make the text document manager listen on the connection
```
 - build the extension and debug it
 - change file type to `html1` (this is what this extension sample uses)
 - trigger a completion
 - the "A request has failed. See the output for more information." dialog popup
 - close the dialog

Expected behaviour:

The Output channel is not opened, focus remain on the current editor pane

Actual behaviour:

The Output channel is opened

https://user-images.githubusercontent.com/521510/125153592-ce247200-e18f-11eb-8e40-04992dcfdec2.mov

---

I believe the problems is here : 

https://github.com/microsoft/vscode-languageserver-node/blob/cf5d1e4d46efd85af4812b9b5dda775067c753fa/client/src/common/client.ts#L3063-L3067

as we can see in vscode API ( https://code.visualstudio.com/api/references/vscode-api#window.showInformationMessage ) this returns a thenable that resolves to the selected item or undefined when being dismissed.

I believe in case the returned value is undefined the output channel should not be revealed.

perrinjerome:
Take into account the user choice and only open the output channel if
user has clicked "Go to output".

Fixes #786

### DIFF
--- a/client/src/common/client.ts
+++ b/client/src/common/client.ts
@@ -3061,8 +3061,10 @@ export abstract class BaseLanguageClient {
 	}
 
 	private showNotificationMessage() {
-		void Window.showInformationMessage('A request has failed. See the output for more information.', 'Go to output').then(() => {
-			this.outputChannel.show(true);
+		void Window.showInformationMessage('A request has failed. See the output for more information.', 'Go to output').then((selection) => {
+			if (selection !== undefined) {
+				this.outputChannel.show(true);
+			}
 		});
 	}
 


dbaeumer:
Thanks!

PR #812, opened 2021-09-15
feat: optional targetSelectionRange property in LocationLink

AkatQuas:
As the function [LocationLink::is](https://github.com/microsoft/vscode-languageserver-node/blob/main/types/src/main.ts#L244-L249) indicates, the property `targetSelectionRange` might be `undefined`.

So I think it might be acceptable to change the signature.

Please correct me if I'm wrong.


dbaeumer:
The problem is with the is method. The `targetSelectionRange` can't be optional.

dbaeumer:
I fixed the is method.

AkatQuas:
Thanks. and here is the [related commit](https://github.com/microsoft/vscode-languageserver-node/commit/3459ab751be078f24ba3328eeb404ca33b3fc94e) for further references.

PR #816, opened 2021-09-16
feat: update type definition

AkatQuas:
Update the type definition.

Since `typescript@4.4.0` , the TS1336 is not problem any more.

So this issue is acceptable if the typescript version is bumped in the root `package.json`.


dbaeumer (commented):
These changes require TS version 4.4.3 which not everybody might use. They produce a compile error when using < 4.4.x

capture

So I would like to hold off of them until this is more widely adopted.

PR #819, opened 2021-09-17
fix: fixed `computeSingleParam` judgement

johnsoncodehk:
If request type create by different `vscode-jsonrpc` module, `LanguageClient.sendRequest` always fails:

```log
Error: Unknown parameter structure auto
    at computeSingleParam (xxx/node_modules/vscode-jsonrpc/lib/common/connection.js:793:23)
    at computeMessageParams (xxx/node_modules/vscode-jsonrpc/lib/common/connection.js:804:26)
    at Object.sendRequest (xxx/node_modules/vscode-jsonrpc/lib/common/connection.js:946:33)
    at Object.sendRequest (xxx/node_modules/vscode-languageclient/lib/common/client.js:38:54)
    at LanguageClient.sendRequest (xxx/node_modules/vscode-languageclient/lib/common/client.js:1918:45)
```

### DIFF
--- a/jsonrpc/src/common/connection.ts
+++ b/jsonrpc/src/common/connection.ts
@@ -1063,19 +1063,19 @@ export function createMessageConnection(messageReader: MessageReader, messageWri
 	}
 
 	function computeSingleParam(parameterStructures: ParameterStructures, param: any): any | any[] {
-		switch(parameterStructures) {
-			case ParameterStructures.auto:
+		switch(parameterStructures.kind) {
+			case ParameterStructures.auto.kind:
 				if (isNamedParam(param)) {
 					return nullToUndefined(param);
 				} else {
 					return [undefinedToNull(param)];
 				}
-			case ParameterStructures.byName:
+			case ParameterStructures.byName.kind:
 				if (!isNamedParam(param)) {
 					throw new Error(`Received parameters by name but param is not an object literal.`);
 				}
 				return nullToUndefined(param);
-			case ParameterStructures.byPosition:
+			case ParameterStructures.byPosition.kind:
 				return [undefinedToNull(param)];
 			default:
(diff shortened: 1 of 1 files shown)


dbaeumer:
I see the problem but can you explain why this is happening. Having two version of json-rpc loaded and using them intermixed will ask for whole set of other problems.

dbaeumer:
Please be also aware that your PR introduces compile errors.

johnsoncodehk:
> I see the problem but can you explain why this is happening.

This is same version json-rpc, but come from different node_modules. Because I created multiple monorepo workapces in a single repo and referenced each other.

> Having two version of json-rpc loaded and using them intermixed will ask for whole set of other problems.

This is the only problem to me, when I change this code in node_modules it's works good.

> Please be also aware that your PR introduces compile errors (kind is private)

Updated, if the change is not appropriate, just feel free to close this.

dbaeumer:
The problem is that the library keeps some global state. If I accept the PR and allow passing in types from different node module locations then I need to start addressing bugs around the global state if they show up. And I don't actually know how to address them if they occur. 

I still don't understand why you can't share one version of the library. If you provide API that uses json-rpc types then you should export these types in your library instead of forcing consumers to install the json-rpc library as well.

johnsoncodehk:
> The problem is that the library keeps some global state. If I accept the PR and allow passing in types from different node module locations then I need to start addressing bugs around the global state if they show up. And I don't actually know how to address them if they occur.

I think no need to allow passing in types from different node module locations even if accept the PR. Not allow is good.

> I still don't understand why you can't share one version of the library. If you provide API that uses json-rpc types then you should export these types in your library instead of forcing consumers to install the json-rpc library as well.

This problem will only occur in local development. After the release, users will always use the json-rpc from the same source.

PR #862, opened 2021-11-25
correct engines field node 14+

CodingDoll:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

Optional chaining is supported by Node 14+.


dbaeumer:
Thanks!

PR #871, opened 2021-12-14
jsonrpc: fix uncaught exceptions in ReadableStreamMessageReader

### LINKED ISSUE #870: Stream read errors result in uncaught exceptions
`ReadableStreamMessageReader` throws exceptions if `Content-Length` is invalid, which end up as uncaught exceptions instead of going through `onError`. See the tests in #871 for an example. Without the try/catch, the test fails:

```
  1) Messages
       Read without Content-Length:
     Uncaught Error: Header must provide a Content-Length property.
      at StreamMessageReader.onData (lib/common/messageReader.js:137:27)
      at Readable.<anonymous> (lib/common/messageReader.js:121:18)
      at Readable.read (_stream_readable.js:512:10)
      at flow (_stream_readable.js:985:34)
      at resume_ (_stream_readable.js:966:3)
      at processTicksAndRejections (internal/process/task_queues.js:84:21)

  2) Messages
       Read with invalid Content-Length:
     Uncaught Error: Content-Length value must be a number.
      at StreamMessageReader.onData (lib/common/messageReader.js:141:27)
      at Readable.<anonymous> (lib/common/messageReader.js:121:18)
      at Readable.read (_stream_readable.js:512:10)
      at flow (_stream_readable.js:985:34)
      at resume_ (_stream_readable.js:966:3)
      at processTicksAndRejections (internal/process/task_queues.js:84:21)
```

This was previously reported in #270, but it has not been fixed.

mroch:
Fixes #870

### DIFF
--- a/jsonrpc/src/common/messageReader.ts
+++ b/jsonrpc/src/common/messageReader.ts
@@ -215,15 +215,20 @@ export class ReadableStreamMessageReader extends AbstractMessageReader {
 				if (!headers) {
 					return;
 				}
-				const contentLength = headers.get('Content-Length');
-				if (!contentLength) {
-					throw new Error('Header must provide a Content-Length property.');
+				try {
+					const contentLength = headers.get('Content-Length');
+					if (!contentLength) {
+						throw new Error('Header must provide a Content-Length property.');
+					}
+					const length = parseInt(contentLength);
+					if (isNaN(length)) {
+						throw new Error('Content-Length value must be a number.');
+					}
+					this.nextMessageLength = length;
+				} catch (err) {
+					this.fireError(err);
+					return;
 				}
-				const length = parseInt(contentLength);
-				if (isNaN(length)) {
-					throw new Error('Content-Length value must be a number.');
-				}
-				this.nextMessageLength = length;
 			}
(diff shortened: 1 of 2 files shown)


jsonrpc/src/common/messageReader.ts
```diff
@@ -215,15 +215,20 @@ export class ReadableStreamMessageReader extends AbstractMessageReader {
...
-				if (!contentLength) {
-					throw new Error('Header must provide a Content-Length property.');
-				}
-				const length = parseInt(contentLength);
-				if (isNaN(length)) {
-					throw new Error('Content-Length value must be a number.');
+				try {
+					const contentLength = headers.get('Content-Length');
+					if (!contentLength) {
+						throw new Error('Header must provide a Content-Length property.');
```
dbaeumer, line 221: Instead of throwing here and then catching again we can directly call `this.fireError(...)`. 

We should also but the reader into a state where it doesn't attempt to read any messages anymore since it will not be able to advance in this situation.

dbaeumer (changes requested):

dbaeumer:
Closing in favor of https://github.com/microsoft/vscode-languageserver-node/pull/1151

PR #888, opened 2022-02-02
feat: more descriptive error when childProcess.spawn fails

### LINKED ISSUE #887: Consider improving error message when launching process fails
It might be nice to improve the error message here:

https://github.com/microsoft/vscode-languageserver-node/blob/dda3acafd7575e6f4a291ec23bc4898c4cb7abfa/client/src/node/main.ts#L463-L465

For example, it could maybe say that the provided executable doesn't exist. We had reports from some users that they hit an error, but it's hard to know what went wrong.

dsherret:
Closes #887

### DIFF
--- a/client/src/node/main.ts
+++ b/client/src/node/main.ts
@@ -349,7 +349,7 @@ export class LanguageClient extends CommonLanguageClient {
 					if (transport === TransportKind.ipc || transport === TransportKind.stdio) {
 						const serverProcess = cp.spawn(runtime, args, execOptions);
 						if (!serverProcess || !serverProcess.pid) {
-							return Promise.reject<MessageTransports>(`Launching server using runtime ${runtime} failed.`);
+							return handleChildProcessStartError(serverProcess, `Launching server using runtime ${runtime} failed.`);
 						}
 						this._serverProcess = serverProcess;
 						serverProcess.stderr.on('data', data => this.outputChannel.append(Is.string(data) ? data : data.toString(encoding)));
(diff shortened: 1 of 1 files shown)


client/src/node/main.ts
```diff
@@ -602,4 +602,19 @@ export class SettingMonitor {
...
+
+function handleChildProcessStartError(process: ChildProcess, message: string) {
+	if (process == null) {
+		return Promise.reject<MessageTransports>(message);
+	}
+
+	return new Promise<MessageTransports>((_, reject) => {
+		process.on("error", (err) => {
+			reject(`${message} ${err}`);
+		});
```
dsherret, line 615: See https://nodejs.org/api/child_process.html#event-error

It's unfortunate that this is an event instead of it just being thrown from `spawn`.

client/src/node/main.ts
```diff
@@ -602,4 +602,19 @@ export class SettingMonitor {
...
\ No newline at end of file
+}
+
+function handleChildProcessStartError(process: ChildProcess, message: string) {
+	if (process == null) {
+		return Promise.reject<MessageTransports>(message);
+	}
+
+	return new Promise<MessageTransports>((_, reject) => {
+		process.on("error", (err) => {
```
dbaeumer, line 613: One minor thing: we use single quotes for constant string values (e.g. 'error')

dbaeumer (changes requested):
Rest looks good to me. Thanks for providing this.

client/src/node/main.ts
dsherret: Done!

dbaeumer (approved):

Issue #879, opened 2022-01-07
npm package distribution - missing source map


char0n:
Hello all,

I've noticed that some files in npm distribution package are referencing non-existing source maps. These non-existing source maps are causing webpack errors when bundling. 

Here are example of the the errors produced:

```
WARNING in ./node_modules/vscode-languageserver-protocol/lib/common/utils/is.js
Module Warning (from ./node_modules/source-map-loader/dist/cjs.js):
Failed to parse source map from '/home/char0n/Documents/GitHub/test/node_modules/vscode-languageserver-protocol/lib/common/utils/is.js.map' file: Error: ENOENT: no such file or directory, open '/home/char0n/Documents/GitHub/test/node_modules/vscode-languageserver-protocol/lib/common/utils/is.js.map'
 @ ./node_modules/monaco-languageclient/lib/monaco-converter.js 10:11-72
 @ ./src/plugins/monaco/adapters/DiagnosticsAdapter.js 4:0-87 51:22-47
 @ ./src/plugins/monaco/adapters/setup-mode.js 6:0-54 34:26-44
 @ ./src/plugins/monaco/adapters/setup.js 8:0-41 24:4-13
 @ ./src/plugins/monaco/components/MonacoEditor.jsx 14:0-69 187:4-21 188:4-17
 @ ./src/plugins/monaco/components/EditorPane.jsx 9:0-42 116:37-49
 @ ./src/plugins/monaco/index.js 3:0-49 8:6-16 11:6-16
 @ ./src/components/IdePreset.jsx 7:0-51 13:33-51
 @ ./src/App.jsx 5:0-41 9:11-14
 @ ./src/index.jsx 6:0-24 9:38-41
```

```
WARNING in ./node_modules/vscode-jsonrpc/lib/common/semaphore.js
Module Warning (from ./node_modules/source-map-loader/dist/cjs.js):
Failed to parse source map from '/home/char0n/Documents/GitHub/test/node_modules/vscode-jsonrpc/lib/common/semaphore.js.map' file: Error: ENOENT: no such file or directory, open '/home/char0n/Documents/GitHub/test/node_modules/vscode-jsonrpc/lib/common/semaphore.js.map'
 @ ./node_modules/vscode-jsonrpc/lib/common/messageWriter.js 16:20-42
 @ ./node_modules/vscode-jsonrpc/lib/common/api.js 227:24-58
[...]

PR #898, opened 2022-02-17
Don't ignore sourcemap files in .npmignore

zachkirsch:
# Fixes https://github.com/microsoft/vscode-languageserver-node/issues/879

_This PR should be considered mutually exclusive with https://github.com/microsoft/vscode-languageserver-node/pull/899 - i.e. only one of the two should be merged._

## The problem 
The packages in this repo are configured to produce source maps at compile-time, which includes:
1. a reference to a source map at the bottom of each generated JS file, e.g. 
    ```js
    // lib/browser/main.js
    ...

        return browser_1.createMessageConnection(reader, writer, logger, options);
    }
    exports.createProtocolConnection = createProtocolConnection;
    //# sourceMappingURL=main.js.map
    ```
1. a source map, e.g. `lib/browser/main.js.map`

However, when the packages are published, the source maps are not uploaded because they are explicitly ignored by npm: https://github.com/microsoft/vscode-languageserver-node/blob/f72faa0975cb2aaa85ce84da819944f84e1aac52/protocol/.npmignore#L4

Therefore, the published package:
1. does include the references at the bottom of each generated JS file
1. does **not** include the actual source maps

These dangling references are causing build errors:
```
Failed to parse source map from '/home/char0n/Documents/GitHub/test/node_modules/vscode-languageserver-protocol/lib/common/utils/is.js.map' file: Error: ENOENT: no such file or directory, open '/home/char0n/Documents/GitHub/test/node_modules/vscode-languageserver-protocol/lib/common/utils/is.js.map'
``` 

## The solution

There are two solutions that I can think of:
1. Don't produce source maps. This effectively maintains the current functionality, but removes the dangling references in the generated JS files.
1. Produce and publish source maps - i.e. don't ignore `lib/\*\*/\*.map` in `.npmignore`

While I think Option 2 is a better devx, I'd understand if you'd like to maintain the current functionality. To that end, I've created two PRs:
1. https://github.com/microsoft/vscode-languageserver-node/pull/899 does Option 1.
1. This PR does Option 2.

Let me know which is your preference and I'll close the other one.

Thanks!

### DIFF
--- a/client/.npmignore
+++ b/client/.npmignore
@@ -1,7 +1,6 @@
 .vscode/
 lib/**/test/
 .vscode-test/
-lib/**/*.map
 lib/**/*.tsbuildInfo
 src/
 test/
--- a/jsonrpc/.npmignore
+++ b/jsonrpc/.npmignore
@@ -1,7 +1,6 @@
 .vscode/
 dist/
 lib/**/test/
-lib/**/*.map
 lib/**/*.tsbuildInfo
 src/
 test/
--- a/protocol/.npmignore
+++ b/protocol/.npmignore
@@ -1,7 +1,6 @@
 .vscode/
 dist/
 lib/**/test/
-lib/**/*.map
 lib/**/*.tsbuildInfo
 src/
 test/
--- a/server/.npmignore
+++ b/server/.npmignore
@@ -1,6 +1,5 @@
 .vscode/
 lib/**/test/
-lib/**/*.map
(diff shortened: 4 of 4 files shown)


dbaeumer:
IMO shipping the source maps is not enough since they will reference the `ts` file. So if we ship the source maps then we need to ship the TS source files as well which I think is not a good idea. This is why I am actually against shipping them in the first place.

PR #910, opened 2022-03-24
Add collapsedText to FoldingRange

ris58h:
https://github.com/microsoft/language-server-protocol/issues/1426


ris58h:
@dbaeumer Do I need to generate meta-model and commit it too?

PR #916, opened 2022-03-31
Provides a way to register a file watcher relative to a given uri.

heejaechang:
Visual studio code added an ability to watch file/folders outside of workspace folder but LSP didn't expose the ability to server yet. This PR adds that support to LSP.

here are link to spec - https://github.com/microsoft/language-server-protocol/pull/1433

and here are link to vscode API - vscode.workspace.createFileSystemWatcher - https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.d.ts#L11516

### DIFF
--- a/protocol/src/common/protocol.ts
+++ b/protocol/src/common/protocol.ts
@@ -1807,6 +1807,12 @@ export interface DidChangeWatchedFilesClientCapabilities {
 	 * from the server side.
 	 */
 	dynamicRegistration?: boolean;
+	/**
+	 * A relative pattern is a helper to construct glob patterns that are matched
+	 * relatively to a base file path. The base path can either be an absolute file
+	 * path as string or uri.
+	 */
+	relativePatternSupport?: boolean;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


heejaechang:
@dbaeumer can you take a look? and if it looks okay, provide me some guidance for me to finish up the PR?

protocol/src/common/protocol.ts
```diff
@@ -1871,17 +1871,42 @@ export interface DidChangeWatchedFilesRegistrationOptions {
 	watchers: FileSystemWatcher[];
 }
 
+export interface GlobPattern {
```
dbaeumer, line 1874: I would name this RelativePattern and use the type GlobPattern as type GlobPattern = RelativePattern | string;

We might discover other pattern forms in the future.

dbaeumer (changes requested):
We also need to do the following things:

- add a client capability in DidChangeWatchedFilesClientCapabilities. Name proposal is `relativePatternSupport?: boolean`.
- we also need a PR against the spec to add this for 3.17.

protocol/src/common/protocol.ts
heejaechang: thank you for review @dbaeumer! will make the change!

heejaechang:
> we also need a PR against the spec to add this for 3.17.

what is the spec repo name?

heejaechang:
@dbaeumer can you take a look? I am not sure whether the doc string I added is good enough or I need to add something more

dbaeumer (approved):

dbaeumer:
I merged it in by hand and did some modifications to it.

PR #929, opened 2022-05-06
make sure to unsubscribe from file events

heejaechang:
I looked through vscode's code base and It doesn't look like ```fileSystemWatcher.dispose()``` through ```disposables.push(fileSystemWatcher);``` will unsubscribe each file watch subscription. 

each must be disposed individually by ```this.hookListeners(fileSystemWatcher, watchCreate, watchChange, watchDelete, **disposables**);``` like how ```registerRaw``` does.

here are relevant code in vscode

https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/extHostFileSystemEventService.ts#L22
https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/extHostFileSystemEventService.ts#L91
https://github.com/microsoft/vscode/blob/main/src/vs/base/common/event.ts#L617
https://github.com/microsoft/vscode/blob/main/src/vs/base/common/event.ts#L681
https://github.com/microsoft/vscode/blob/main/src/vs/base/common/lifecycle.ts#L356

basically, no one is calling dispose on listener.subscription which is disposable returned by fileWatcher.onXXX call.

### DIFF
--- a/client/src/common/fileSystemWatcher.ts
+++ b/client/src/common/fileSystemWatcher.ts
@@ -62,7 +62,7 @@ export class FileSystemWatcherFeature implements DynamicFeature<DidChangeWatched
 				watchDelete = (watcher.kind & WatchKind.Delete) !== 0;
 			}
 			const fileSystemWatcher: VFileSystemWatcher = Workspace.createFileSystemWatcher(globPattern, !watchCreate, !watchChange, !watchDelete);
-			this.hookListeners(fileSystemWatcher, watchCreate, watchChange, watchDelete);
+			this.hookListeners(fileSystemWatcher, watchCreate, watchChange, watchDelete, disposables);
 			disposables.push(fileSystemWatcher);
 		}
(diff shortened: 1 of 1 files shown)


dbaeumer (approved):

dbaeumer:
@aeschli , can you have a look as well.

aeschli:
@heejaechang Did you file a bug against vscode as well?

heejaechang:
@aeschli is it a bug in vscode? I don't see any other code that use Event class remembers all subscription made to the event and dispose them when event itself is disposed. instead, all other case, each caller that calls ```Event.event(...)``` remember returned ```disposable``` and explicitly dispose them.

so, it seems fs watcher behavior around event is just normal pattern in vscode?

you can search for ```Event.event``` usages and check how they are used in vscode.

heejaechang:
is there something I need to do to be checked in?

Issue #934, opened 2022-05-11
Definition mismatch and wrong conversion of InlineValueContext 


d-biehl:
Hi,

wanted to implement 'textDocument/inlineValue'  from version 3.17 of LSP Documentation in Python and use Version 8.0.0 of vscode-languageclient in my VSCode extension.

But struggeling with the different definition of InlineValueContext in [LSP Documentation](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#inlineValueContext) and the implementation in [vscode-languageserver-types](https://github.com/microsoft/vscode-languageserver-node/blob/f1f8e1a28a3cd35d21b41966724ac405ca52ce26/types/src/main.ts#L3872)  and the implementation of what I get in `textDocument/inlineValue` request:  

```json
{
  "jsonrpc": "2.0",
  "id": 21,
  "method": "textDocument/inlineValue",
  "params": {
    "textDocument": { "uri": "file:///c%3A/tmp/rt1/tests/w.robot" },
    "range": {
      "start": { "line": 0, "character": 0 },
      "end": { "line": 17, "character": 0 }
    },
    "context": {
      "frameId": 2381178444560,
      "stoppedLocation": [
        { "line": 6, "character": 1 },
        { "line": 6, "character": 1 }
      ]
    }
  }
}
```
 
`stoppedLocation` is not a Range structure, I think it is not converted correctly from vscode.Range to lsp.Range. The `frameID` is interessting, but I think in LSP implementation not relevant, maybe only together with a DAP implementation.

ok, so what is the correct defintion/implementation of InlineValueContext?

thanks
Daniel

dbaeumer:
I added the frameId to the spec. I think it could be useful in cases where a server knows about DAP which is OSS as well.

dbaeumer:
Done and ships with 8.0.1

PR #936, opened 2022-05-11
Convert range to protocol in InlineContext

dbaeumer:


dbaeumer:
See https://github.com/microsoft/vscode-languageserver-node/issues/934

aeschli (approved):

PR #938, opened 2022-05-12
Fix meta model

dbaeumer:
Minor fixes to the meta model

### DIFF
--- a/protocol/metaModel.json
+++ b/protocol/metaModel.json
@@ -49,7 +49,7 @@
 				"kind": "reference",
 				"name": "ImplementationRegistrationOptions"
 			},
-			"documentation": "A request to resolve the implementation locations of a symbol at a given text\ndocument position. The request's parameter is of type [TextDocumentPositionParams]\n(#TextDocumentPositionParams) the response is of type [Definition](#Definition) or a\nThenable that resolves to such."
+			"documentation": "A request to resolve the implementation locations of a symbol at a given text document position. The request's parameter is of type [TextDocumentPositionParams] (#TextDocumentPositionParams) the response is of type [Definition](#Definition) or a Thenable that resolves to such."
 		},
 		{
 			"method": "textDocument/typeDefinition",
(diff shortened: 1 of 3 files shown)


protocol/metaModel.json
```diff
@@ -5346,7 +5346,7 @@
 						}
 					},
 					"optional": true,
-					"documentation": "Marks that the code action cannot currently be applied.\nClients should follow the following guidelines regarding disabled code actions:\n  - Disabled code actions are not shown in automatic [lightbulb](https://code.visualstudio.com/docs/editor/editingevolved#_code-action)\n    code action menu.\n  - Disabled actions are shown as faded out in the code action menu when the user request a more specific type\n    of code action, such as refactorings.\n  - If the user has a [keybinding](https://code.visualstudio.com/docs/editor/refactoring#_keybindings-for-code-actions)\n    that auto applies a code action and only a disabled code actions are returned, the client should show the user an\n    error message with `reason` in the editor.\n@since 3.16.0",
```
DanTup, line 5349: I think the linebreak removal should be more selective. In this case, there are list items that got mashed together:

```
  /// Marks that the code action cannot currently be applied.
  ///
  /// Clients should follow the following guidelines regarding disabled code
  /// actions:
  ///
  /// - Disabled code actions are not shown in automatic lightbulbs code
  ///   action menus.
  ///
  /// - Disabled actions are shown as faded out in the code action menu when
  ///   the user request a more specific type of code action, such as
  ///   refactorings.
```

Ideally, the linebreaks before the list items would not be removed, but those that are only wrapping text would. This might be tricky if the source of truth requires the word-wrapping, though replacing out only `\b\n\b` might get most of the way (so any line ending with a fullstop would not have its newline stripped).

aeschli (dismissed):

dbaeumer:
@DanTup. Goo point. So, I will keep the line breaks and not remove any since this will very likely never be correct.

aeschli (approved):

DanTup:
@dbaeumer sounds fair to me. I was previously just replacing `\b\n\b` out myself and the results weren't bad (mostly I'm trying to minimise the diff between the TS version and JSON version so I can reasonably review all generated code before switching over). I can put that back in if it's not trivial to do here.

dbaeumer:
@DanTup: actually, I don't want to be responsible for any mis formatting. That is why I think keeping then `\n` as is is the best we can do right now.

DanTup:
@dbaeumer Yeah sorry - what I meant was I will go back to handling the re-wrapping in my code generation scripts in Dart (at least if the results all seem good when I review the full set). I'm already doing some other clean-up to make things work better in Dart so I'm not too worried about handling this (and if necessary, some exceptions) to improve the generated code.

Thanks!

PR #941, opened 2022-05-13
Dbaeumer/involved-goat-bronze

dbaeumer:
More small fixed to the meta model generation.


aeschli (approved):

rcjsuen:
@dbaeumer This is one heck of a branch name. I am confused... 🤔

dbaeumer:
@rcjsuen we will move to a new model where all code will come via PRs into the main branch. Since it is hard to come up with branch names in this model we have a branch name generator comparable to the docker name generator. So you will see more of these names.

PR #942, opened 2022-05-13
Add messageStrategy connection option

eddyerburgh:
This PR adds an option to configure a custom message handler.

# Motivation

I'm propagating a value between language servers by adding it to messages as they are written, and then extracting the value when the message is read and using [`AsyncLocalStorage.run`](https://nodejs.org/api/async_context.html#asynclocalstoragerunstore-callback-args) to run the message handler with the value in its context. 

Currently, there is no way to configure a message handler to run inside a given context. We have `messageReader.listen`, but the message handler isn't guaranteed to be run in the same context as `messageHandeler.listen` because `triggerMessageQueue` might already have schedule `processMessageQueue` to run on the next tick.


dbaeumer:
@eddyerburgh thanks for the PR. Can you please also open an issue with an example where this can be used to track this.

client/src/common/client.ts
```diff
@@ -331,7 +331,8 @@ export type LanguageClientOptions = {
 	};
 	workspaceFolder?: VWorkspaceFolder;
 	connectionOptions?: {
-		cancellationStrategy: CancellationStrategy;
```
eddyerburgh, line 334: I don't think this is required since connectionOptions is just passed in to createProtocolHandler which doesn't require cancellationStrategy. (maxRestartCount is used in BaseLanguageClient constructor though so I guess that's why connectionOptions doesn't have type ConnectionOptions)

eddyerburgh:
@dbaeumer I've created an issue: https://github.com/microsoft/vscode-languageserver-node/issues/943

eddyerburgh:
@dbaeumer do you need anything else from me here?

dbaeumer:
No nothing. The thing is that I just shipped 3.17 and I want to wait a little with new feature. Will look into this in June or is it urgent for you?

eddyerburgh:
Thanks @dbaeumer. The sooner the better for me. Do you have timelines on the next release?

dbaeumer:
~ Mid June

eddyerburgh:
@dbaeumer are you able to look into this now?

dbaeumer:
I need to get 8.0.2 out first before adding new features which is planned for this week. Then I will work on merging this one in.

eddyerburgh:
Hi @dbaeumer are you back from vacation yet? If not, when do you expect to be back?

jsonrpc/src/common/connection.ts
```diff
@@ -615,21 +626,28 @@ export function createMessageConnection(messageReader: MessageReader, messageWri
...
 		const message = messageQueue.shift()!;
 		try {
-			if (Message.isRequest(message)) {
-				handleRequest(message);
-			} else if (Message.isNotification(message)) {
-				handleNotification(message);
-			} else if (Message.isResponse(message)) {
-				handleResponse(message);
+			const messageStrategy = options?.messageStrategy;
+			if (messageStrategy && MessageStrategy.is(messageStrategy)) {
```
dbaeumer, line 647: IMo testing that `messageStrategy` is set is not necessary since `MessageStrategy.is(messageStrategy)` will fail if it is so.

client/src/common/client.ts
dbaeumer: It is used here: https://github.com/microsoft/vscode-languageserver-node/blob/8b56a58d0a706bc0cc8e947b42ad135faf488fe3/jsonrpc/src/common/connection.ts#L540 and travels down their via structural typing.

dbaeumer (changes requested):
Can you please look at the set test comment

jsonrpc/src/common/connection.ts
eddyerburgh: Good point, updated

alexr00 (approved):

dbaeumer:
@eddyerburgh I need to revert this since it makes a lot of test fails. I don't know yet why the build got green but the failing test cases can be seen here:

https://dev.azure.com/ms/vscode-languageserver-node/_build/results?buildId=349017&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=3f67303d-caa6-535b-2055-750b67f1a4bf&l=601

Can you please make sure that all tests pass. You can run them via `npm run test`.

dbaeumer:
The tests for the commit before still passed 

https://dev.azure.com/ms/vscode-languageserver-node/_build/results?buildId=343462&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=3f67303d-caa6-535b-2055-750b67f1a4bf&l=550

jsonrpc/src/common/connection.ts
```diff
@@ -393,14 +393,25 @@ export namespace CancellationStrategy {
...
 export interface ConnectionOptions {
 	cancellationStrategy?: CancellationStrategy;
 	connectionStrategy?: ConnectionStrategy;
+	messageStrategy?: MessageStrategy;
 }
 export namespace ConnectionOptions {
 	export function is(value: any): value is ConnectionOptions {
 		const candidate: ConnectionOptions = value;
-		return candidate && (CancellationStrategy.is(candidate.cancellationStrategy) || ConnectionStrategy.is(candidate.connectionStrategy));
+		return candidate && (CancellationStrategy.is(candidate.cancellationStrategy) || ConnectionStrategy.is(candidate.connectionStrategy)) || MessageStrategy.is(candidate.messageStrategy);
```
eddyerburgh, line 414: @dbaeumer this was the issue, sorry about the churn. I'll put up a fresh pr

PR #949, opened 2022-05-17
Fix meta model documentation for enum values

DanTup:
@dbaeumer feel free to close this if you're already looking at it (or it's not the right fix). I don't know why this `.parent` is required, but it appears to fix all of the missing enum docs.

### DIFF
--- a/tools/src/visitor.ts
+++ b/tools/src/visitor.ts
@@ -812,8 +812,7 @@ export default class Visitor {
 									break;
 								}
 								const entry: EnumerationEntry = { name: declaration.name.getText(), value: value };
-								// Use the declaration statement since enum declaration only have one declaration.
-								this.fillDocProperties(variable.declarationList.parent, entry);
+								this.fillDocProperties(variable, entry);
 								enumerations.push(entry);
 							}
 							if (isEnum) {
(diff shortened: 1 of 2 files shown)


tools/src/visitor.ts
```diff
@@ -813,7 +813,7 @@ export default class Visitor {
 								}
 								const entry: EnumerationEntry = { name: declaration.name.getText(), value: value };
 								// Use the declaration statement since enum declaration only have one declaration.
-								this.fillDocProperties(variable.declarationList, entry);
+								this.fillDocProperties(variable.declarationList.parent, entry);
```
dbaeumer, line 816: Shouldn't the then be `this.fillDocProperties(variable, entry)` instead of going down one and up one.

We should also fix the comment above.

dbaeumer (changes requested):

dbaeumer:
@DanTup thanks for the PR. See my comment.

tools/src/visitor.ts
DanTup: Oh, ofc - I totally misread that the first time. Fixed!

dbaeumer (approved):

aeschli (approved):

PR #951, opened 2022-05-17
fix: return `false` when `showDocument` fails

Vtec234:
If there is an error during `showDocument` execution, it sounds like the request should return `success: false`. On the other hand if this is intentional, feel free to close.


dbaeumer:
Good catch.

dbaeumer (approved):

aeschli (approved):

Issue #381, opened 2018-07-26
Understanding the executeCommand workflow


sanboxrunner:
Hi, 

Apologies if this is the wrong place to ask this, but the documentation is not clear on this. I'm trying to add a command to the VSCode Command Palette (or any other similar set up). I understand that the command "lives" client side, and simply when the user invokes it, the server picks it up and can respond on executeCommand. That works fine. But I want, when the user invokes the command, for a dialogue box (similar to a Git commit asking you for your username and password) to pop up and ask for additional information.

It seems, on the face of it, to be `connection.sendRequest` but I'm unsure how to use it, and the documentation, such as it is, around it is unclear. 

Any help pointing me in the right direction would be appreciated.

Gama11:
You might want to create some custom methods?

https://stackoverflow.com/questions/51041337/vscode-language-client-extension-how-to-send-a-message-from-the-server-to-the

sanboxrunner:
Hmm. That's kind of unfortunate -- that *really* chains the LSP to a specific client. Thanks though!

PR #957, opened 2022-05-24
Avoid circular dependencies in SemVer

samrose3:
## Problem

Consumers using bundlers like Rollup or Webpack are hitting a circular dependency error coming from [npm/node-semver](https://github.com/npm/node-semver). This is causing the `SemVer,satisfies` function to always return `false` when bundled, resulting in the following error:

semver-satisfies-error

## What this PR is doing

A circular dependency between the Range and Comparator classes can cause issues for consumers who are trying to bundle extensions with `vscode-languageclient` as a dependency.

See related GitHub issue from [npm/node-semver](https://github.com/npm/node-semver)

- [[BUG] the package isn't compatible with Rollup due to require cycle #381](https://github.com/npm/node-semver/issues/381)

### Using Rollup to bundle VS Code extension

| When importing all of SemVer | When only importing individual functions |
| --- | --- |
| Screen Shot 2022-05-23 at 7 15 24 PM | No circular deps and the extension works! ✨  Tested in external project using `yalc` |

Verified the build output of `client/lib/node/main.js` is correctly importing only the needed functions, avoiding the circular dependency:

```js
// client/lib/node/main.js

const semverParse = require("semver/functions/parse");
const semverSatisfies = require("semver/functions/satisfies");
```

### DIFF
--- a/client/src/node/main.ts
+++ b/client/src/node/main.ts
@@ -8,8 +8,6 @@ import ChildProcess = cp.ChildProcess;
 import * as fs from 'fs';
 import * as path from 'path';
 
-import * as SemVer from 'semver';
-
 import { workspace as Workspace, Disposable, version as VSCodeVersion } from 'vscode';
 
 import * as Is from '../common/utils/is';
@@ -17,6 +15,10 @@ import { BaseLanguageClient, LanguageClientOptions, MessageTransports } from '..
 
 import { terminate } from './processes';
 import { StreamMessageReader, StreamMessageWriter, IPCMessageReader, IPCMessageWriter, createClientPipeTransport, generateRandomPipeName, createClientSocketTransport, InitializeParams} from 'vscode-languageserver-protocol/node';
+
+// Import SemVer functions individually to avoid circular dependencies in SemVer
+import semverParse = require('semver/functions/parse');
+import semverSatisfies = require('semver/functions/satisfies');
 
 export * from 'vscode-languageserver-protocol/node';
 export * from '../common/api';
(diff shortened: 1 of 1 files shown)


dbaeumer:
@samrose3 I see the problem but I am not a fan of merging this in since it puts me at the mercy of the semver directory layout. IMO this should be addressed in semver directly. Wouldn't it make more sense to provide a PR there?

samrose3:
Hi @dbaeumer! Just following up on this 👋 

Given that accessing the `semverParse` and `semverSatisfies` functions are part of the [supported usage](https://github.com/npm/node-semver#usage) for the `node-semver` package, would it be alright to merge this fix in? I feel pretty confident it is safe using these function paths.

For more context, we are using `vscode-languageclient` in a VS Code extension. We are needing to bundle node modules with our extension and are using Rollup for all our monorepo package builds.

I have tested this change and verified that it only pulls in the small parts of `semver` needed for the `checkVersion` function and allows Rollup to properly tree-shake the dependency. This also results in a smaller bundle size for us (the consumer of `vscode-languageclient` which is another win 🎉 

Any other concerns I can help address? I've tried resolving the circular dep issue within `node-semver` but it is unfortunately a much bigger refactor than I have the context to solve :(

dbaeumer:
I looked at the semver documentation and they indeed advocate using the functions directly. Hope that keep that stable as well :-).

dbaeumer (approved):

aeschli (approved):

PR #945, opened 2022-05-16
Fix typos + minor doc tweaks

DanTup:
This is mostly typos and tweaks I noticed when diffing code generated from parsing the TypeScript spec to the new meta model, but also some additional things I spotted along the way.

Some of the changes are quite trivial like trailing full stops.. I mostly added these where they existed in the LSP spec (in an attempt to reduce the size of my diff to help me review the TS -> meta model migration), I didn't try to fix these in all places.

@dbaeumer

PR #973, opened 2022-05-31
Publish 8.0.2-next.3

dbaeumer:


aeschli (approved):

DanTup:
@dbaeumer I'm not sure why, but some of the changes from #945 seem to be missing here (but others are included). If I regen the model from current code, I get the updated docs:

Screenshot 2022-06-01 at 14 56 48

I can send a PR with the updated version, although I don't know if it's deliberate (or there's something different between our setups that is causing a difference?).

dbaeumer:
I will have a look.

DanTup:
Thanks! The PR I intended to post this one was https://github.com/microsoft/vscode-languageserver-node/pull/971, that seems to be the one that regenerated it with some (but not all) of thesse changes.

dbaeumer:
A new version should be available on the web.

PR #976, opened 2022-06-01
cleanup diagnostics in DiagnosticRequestor

### LINKED ISSUE #975: Pull Diagnostics does not remove old diagnostics if Language Server restarts
I have tried to implement [Pull Diagnostics](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_pullDiagnostics) for my language server. Sometimes it is necessary to restart the server instance, e.g. because of changed settings.
But when the server is restarted the old diagnostics are not deleted and are now double/triple/... present.

After analyzing the LS client, in my opinion, the "diagnostics" member in the DiagnosticRequestor must also be disposed.

d-biehl:
try to fix #975

### DIFF
--- a/client/src/common/diagnostic.ts
+++ b/client/src/common/diagnostic.ts
@@ -641,6 +641,9 @@ class DiagnosticRequestor implements Disposable {
 			}
 			this.openRequests.set(key, { state: RequestStateKind.outDated, document: request.document });
 		}
+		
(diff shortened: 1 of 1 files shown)


dbaeumer:
Good catch!

dbaeumer (approved):

PR #1007, opened 2022-06-21
Add information about message direction to metamodel

### LINKED ISSUE #998: metamodel: Add information about whether a request/notification is sent from the client or the server
This would be handy for implementations which treat these kinds of messages differently, e.g. the Haskell implementation uses this to make sure servers don't accidentally register handlers for messages that can't be sent to them.

michaelpj:
Fixes #998.

It would be good to get the info checked! I went through and everything that said it's sent from the server got `serverToClient`, `progress` and `cancelRequest` got `both`, and everything else got `clientToServer`.

### DIFF
--- a/protocol/src/common/protocol.$.ts
+++ b/protocol/src/common/protocol.$.ts
@@ -6,7 +6,7 @@
 import { LogTraceParams, SetTraceParams, ProgressToken, ErrorCodes } from 'vscode-jsonrpc';
 import { SemanticTokenTypes, SemanticTokenModifiers, LSPAny } from 'vscode-languageserver-types';
 
-import { ProtocolNotificationType } from './messages';
+import { MessageDirection, ProtocolNotificationType } from './messages';
 import { LSPErrorCodes } from './api';
 import { WorkDoneProgressBegin, WorkDoneProgressEnd, WorkDoneProgressReport } from './protocol.progress';
 
@@ -15,12 +15,16 @@ import { WorkDoneProgressBegin, WorkDoneProgressEnd, WorkDoneProgressReport } fr
 
 // @ts-ignore 6196
 namespace SetTraceNotification {
-	export const type = new ProtocolNotificationType<SetTraceParams, void>('$/setTrace');
+  export const method: '$/setTrace' = '$/setTrace';
+	export const messageDirection: MessageDirection = MessageDirection.clientToServer;
+	export const type = new ProtocolNotificationType<SetTraceParams, void>(method);
 }
 
 // @ts-ignore 6196
 namespace LogTraceNotification {
-	export const type = new ProtocolNotificationType<LogTraceParams, void>('$/logTrace');
+  export const method: '$/logTrace' = '$/logTrace';
+	export const messageDirection: MessageDirection = MessageDirection.serverToClient;
+	export const type = new ProtocolNotificationType<LogTraceParams, void>(method);
 }
 
 // @ts-ignore 6196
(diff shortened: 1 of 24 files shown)


michaelpj:
Okay, so the build fails because it can't build the visitor that auto-generates the metamodel, which isn't surprising since I'm not sure we *have* the information in the typescript implementation to use. This is really something we need to specify in the metamodel... Not sure what to do about that.

dbaeumer:
@michaelpj sorry for not mentioning that earlier but we can't add this to the metaModel.json since that one is generated using `npm run generate:metaModel`. We nee to annotate the Request definitions. So we would do something like this:

```
export namespace CompletionRequest {
        export const direction: MessageDirection = MessageDirection.clientToServer;
	export const method: 'textDocument/completion' = 'textDocument/completion';
	export const type = new ProtocolRequestType<CompletionParams, CompletionItem[] | CompletionList | null, CompletionItem[], void, CompletionRegistrationOptions>(method);
}
```

and then adopt the generator to honor this information. I can help with fixing the generator if you do a PR that adds the information to the request definitions.

michaelpj:
Okay, I think I've done that. I haven't touched the visitor or the generated metaModel. I can try and give that a shot, but I haven't actually managed to properly build the project yet...

dbaeumer:
To build the project:

- clone it
- run npm install
- run npm run symlink
- run npm run compile

dbaeumer:
Let me know if you need further help.

michaelpj:
Okay, I built the project and all the protocol part compiles, so it looks like I did all the adding of `messageDirection` fields okay. But I'm not a typescript dev and it's not obvious to me how to modify the visitor. Could you take a look? I suspect it's just a couple of lines... I just don't know what they are!

The branch here has my work so far.

michaelpj:
Done.

dbaeumer (dismissed):

protocol/src/common/protocol.ts
```diff
@@ -332,7 +332,9 @@ export interface RegistrationParams {
  * handler on the client side.
  */
 export namespace RegistrationRequest {
-	export const type = new ProtocolRequestType<RegistrationParams, void, never, void, void>('client/registerCapability');
+  export const method: 'client/registerCapability' = 'client/registerCapability';
```
aeschli, line 335: Looks like there's an unusual indent

aeschli (approved):

michaelpj:
@dbaeumer the build fails because the visitor isn't filling in the new field, but that requires fixing the visitor :( Feel free to push changes to my branch if you work on this!

protocol/src/common/protocol.ts
michaelpj: Fixed!

dbaeumer:
@michaelpj even if I make `messageDirection` optional in the meta model (which we should do to get the PR in a state were we can merge it) I still have compile errors in your repository. The first one is:

capture

which is actually correct :-)

dbaeumer:
`MessageDirection` should be defined on the protocol level and not the types level.

michaelpj:
Thanks, I'll take another look, I thought I was down to just the visitor :)

Should I make it optional in the meta model for now? I guess that would make things compile and then we can make it mandatory later.

michaelpj:
Alright, now it's green. I made `messageDirection` optional, but I think you should be able to make it mandatory when you update the visitor.

dbaeumer:
Yes, that is the right way to go. We keep it optional for now until the visitor is updated.

protocol/src/common/protocol.$.ts
```diff
@@ -15,12 +15,16 @@ import { WorkDoneProgressBegin, WorkDoneProgressEnd, WorkDoneProgressReport } fr
 
 // @ts-ignore 6196
 namespace SetTraceNotification {
-	export const type = new ProtocolNotificationType<SetTraceParams, void>('$/setTrace');
+  export const method: '$/setTrace' = '$/setTrace';
```
dbaeumer, line 18: @michaelpj these indents still seem to be off. Looks like that use 2 spaces, but in this repository we indent using tabs. Could you please address this as well. 

Thanks.

dbaeumer (changes requested):

protocol/src/common/protocol.$.ts
michaelpj: Argh sorry, I thought I'd got all of them >:(

Issue #1038, opened 2022-07-23
vscode-languageclient minmatch dependency version


Ehesp:
Currently `vscode-languageclient` depends on an old version of `minimatch`: https://github.com/microsoft/vscode-languageserver-node/blob/main/client/package.json#L31

Version 3 of minimatch requires the `path` dependency directly which can cause issues on the client when bundling, whereas the later version removes this and exports it's own path implementation.

Would it be possible to bump the dependency version? I'll take a look at doing it and running some tests in the week if possible.

dbaeumer:
See https://github.com/microsoft/vscode-languageserver-node/pull/1046.

Ehesp:
Thanks!

PR #1046, opened 2022-08-03
Update dependencies

dbaeumer:


dbaeumer:
Update dependencies to fix depbot allert.

alexr00 (approved):

Ehesp:
@dbaeumer @alexr00 would it be possible to publish a new release with this change?

dbaeumer:
@Ehesp could you explain why?

Ehesp:
@dbaeumer Sure, https://github.com/TypeFox/monaco-languageclient/blob/main/packages/client/package.json#L58 currently points to 8.0.2, which includes the v3 minimatch package, which causes this issue https://github.com/microsoft/vscode-languageserver-node/issues/1038

dbaeumer:
Will look into it after end game week (which is this week).

PR #1065, opened 2022-08-26
Fix double pull of documents

dbaeumer:


.vscode/settings.json
```diff
@@ -9,6 +9,9 @@
 	"files.eol": "\n",
 	"files.trimTrailingWhitespace": true,
 
+	"task.allowAutomaticTasks": "on",
```
aeschli, line 12: Is the` settings.json` change intended?

client/src/common/diagnostic.ts
```diff
@@ -402,7 +408,8 @@ class DiagnosticRequestor implements Disposable {
 	}
 
 	public knows(kind: PullState, document: TextDocument | Uri): boolean {
-		return this.documentStates.tracks(kind, document);
+		const uri = document instanceof Uri ? document : document.uri;
+		return this.documentStates.tracks(kind, document) || this.openRequests.has(uri.toString());
```
aeschli, line 412: I wonder if we should use `toString(true)`, just to set a good example.
It's of course not relevant here where all we need is a key for the URI.

.vscode/settings.json
dbaeumer: Yes this is intended to auto start the compiler / watch task

client/src/common/diagnostic.ts
dbaeumer: In general I agree however we would need to change at least all places in this file since `openRequests` is used at other places as well. Since I in LSP never used toString(true) I kept it that way. In new implementation I do I actually switch to toString(true). 

Hope that is OK for you.

aeschli (approved):

PR #1118, opened 2022-11-09
A more reliable way to check the document version when handling ApplyWorkspaceEdit

qoter:
Some language servers and vscode have a bit different ways to encode uri to string. For example [csharp-language-server-protocol](https://github.com/OmniSharp/csharp-language-server-protocol) encodes uri `C:\abc` to `file:///c:/abc` but vscode encodes this uri a bit more aggressive: `file:///c%3A/abc`. As a result we have a problem with lookup in a map in the method `handleApplyWorkspaceEdit`. To fix it I am going to call converter `asUri` before call map lookup by uri.


dbaeumer:
Good catch.

dbaeumer (approved):

aeschli (approved):

PR #1122, opened 2022-11-15
Adopt `@link` for jsdoc comments

### LINKED ISSUE #1015: Adopt JSDoc `@link` for comments
# Problem
Currently JSDoc comments in the generated `d.ts` files use markdown style syntax, such as `[Location](#Location)`:

https://github.com/microsoft/vscode-languageserver-node/blob/2e889fbffaccee0d054dc8af1feb356b27dc2873/types/src/main.ts#L241

This is rendered as a link by VS Code but is not actual valid. Trying to click it results in an error

# Proposal
Instead these comments should use JSDoc's `{@link}` syntax: https://jsdoc.app/tags-inline-link.html

Not only does this create links that VS Code can understand, it also makes TypeScript aware of which symbols are referenced in comments so that they are updated on rename and listed by find all references

mjbvz:
Fixes #1015

Converts the jsdoc comments to use `@link` syntax for links. This makes these links work inside editors like VS Code

### DIFF
--- a/jsonrpc/src/common/cancellation.ts
+++ b/jsonrpc/src/common/cancellation.ts
@@ -21,7 +21,7 @@ export interface CancellationToken {
 	readonly isCancellationRequested: boolean;
 
 	/**
-	 * An [event](#Event) which fires upon cancellation.
+	 * An {@link Event event} which fires upon cancellation.
 	 */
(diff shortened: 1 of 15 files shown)


dbaeumer:
Thanks for providing the PR!

dbaeumer (approved):

PR #1129, opened 2022-11-24
Fixes #1128: JSON no diagnostics

### LINKED ISSUE #1128: JSON no diagnostics
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.74.0
- OS Version: macOS

This problems exists from `1.68.0`. in `1.67.0` everything seems to be fine.

1. Extension code to reproduce:

```ts
// most probably you want to register it to some command
const testContent = '{test: 5}'

vscode.workspace
    .openTextDocument({
        content: testContent,
        language: 'json',
    })
    .then(async doc => {
        await vscode.window.showTextDocument(doc)
    })
```

2. Run this code twice

Actual: First editor has diagnostics and second not
Expected: Both editors have diagnostics

The second call can also open file instead. This was happening to me in stable builds in some cases without any extensions, but the only consistent way I found to reproduce is this ext code.

Attaching gif:

Text changes doesn't produce diagnostics in these editors. Of course, after reload diagnostics will be there.

dbaeumer:
#1128

### DIFF
--- a/testbed/client/src/extension.ts
+++ b/testbed/client/src/extension.ts
@@ -5,7 +5,7 @@
 'use strict';
 
 import * as path from 'path';
-import { commands, ExtensionContext, Uri } from 'vscode';
+import { commands, ExtensionContext, workspace, window } from 'vscode';
 import { LanguageClient, LanguageClientOptions, ServerOptions, TransportKind, NotificationType, SuspendMode, DidOpenTextDocumentNotification } from 'vscode-languageclient/node';
 
 let client: LanguageClient;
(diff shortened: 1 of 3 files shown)


testbed/client/src/extension.ts
```diff
@@ -78,6 +78,30 @@ export async function activate(context: ExtensionContext) {
 			console.log(value);
 		});
 	});
+
+	commands.registerCommand('testbed.openFile', () => {
+		// most probably you want to register it to some command
```
aeschli, line 83: this comment can be removed...

aeschli (approved):

PR #1148, opened 2022-12-13
Fixes #1057: Consider allowing suppression of CloseAction error/info message

### LINKED ISSUE #1057: Consider allowing suppression of `CloseAction` error/info message
Over in the PowerShell extension I was finally able to update to the latest version of this library, and it is _much_ improved, thank you so much. Thing is, we already have our own restart experience. For architectural reasons, the extension's LSP server _is also_ an extension-owned terminal. This means that users sometimes close the terminal, and then that shuts down the language server. So for a long time we've already thrown an error message that has to be very clearly worded that the terminal is a requirement for the server (and so for LSP features like completions etc.), and give a Yes/No prompt to restart it. This works great, but means that we use `DoNotRestart` for our `CloseAction`.

However, in the update this library now throws an error message:

https://github.com/microsoft/vscode-languageserver-node/blob/3b6f361635a1478c6e782e04d2833c73281d86aa/client/src/common/client.ts#L1469-L1486

and I can't suppress it! Which annoyingly to our users leads to _two_ error messages, and we'd really like to keep ours as it's an interactive prompt _asking_ if the user wants to restart the server.

I would like for you to consider implementing one of two options:

* Add a `PromptForRestart` option, that we would then move to instead of `DoNotRestart` and has a similar Yes/No option for the user, and then we just feed our existing message as customization
* Add a way to suppress the error/info message thrown by the library

For now, in https://github.com/PowerShell/vscode-powershell/pull/4141 I've customized the message in our `CloseAction` to refer to the prompt, and so the experience looks like:

Screen Shot 2022-08-17 at 5 54 05 PM

dbaeumer:
#1057

### DIFF
--- a/client/src/common/client.ts
+++ b/client/src/common/client.ts
@@ -155,6 +155,14 @@ export type ErrorHandlerResult = {
 	 * An optional message to be presented to the user.
 	 */
 	message?: string;
+
+	/**
+	 * If set to true the client assumes that the corresponding
+	 * error handler has presented an appropriate message to the
+	 * user and the message will only be log to the client's
+	 * output channel.
+	 */
+	handled?: boolean;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


aeschli (approved):

andschwa:
Thank you!

PR #1150, opened 2022-12-14
Fixes #1090: Return Promise from ErrorHandler.closed()

### LINKED ISSUE #1090: Return Promise from `ErrorHandler.closed()`
Would it be possible to change `ErrorHandler.close()` to return a `Promise<CloseHandlerResult>` instead of `CloseHandlerResult`? With it returning a `CloseHandlerResult` immediately it limits what extensions can do when the LSP server crashes.

When the LSP server crashes repeatedly I would like to display a dialog  asking the user what they would like to do. Do you want to restart the server, quit running the server, give them a chance to edit their code to fix the issue that caused the LSP server crash. I can't see how I can do this without `closed()` returning a promise.

This is to support using development versions of language compilers and LSP in particular the Swift compiler.

dbaeumer:
#1090

### DIFF
--- a/client-node-tests/src/integration.test.ts
+++ b/client-node-tests/src/integration.test.ts
@@ -1570,8 +1570,8 @@ class CrashClient extends lsclient.LanguageClient {
 		});
 	}
 
-	protected handleConnectionClosed(): void {
-		super.handleConnectionClosed();
+	protected async handleConnectionClosed(): Promise<void> {
+		await super.handleConnectionClosed();
 		this.resolve!();
 	}
 }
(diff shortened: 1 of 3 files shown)


aeschli (approved):

adam-fowler:
@dbaeumer any idea when this will make it into an official 8.1 release? Given the 8.1.0.next is already out

dbaeumer:
After we have shipped a new VS Code version which will be end of Jan / beginning of Feb.

adam-fowler:
Thanks for the prompt update 👍

PR #1158, opened 2022-12-20
Implement general middleware for requests and notifications

### LINKED ISSUE #918: Allow middleware that can handles all messages
When debugging performance, it would be useful to be able to time how long a request takes from being sent to the language server to the response being available (specifically including all time spend serialisation/transmission/etc.).

Right now I think to do this you'd need to implement all middleware functions and wrap them with timings. It would be nice if there was a single hook that could be used for _all_ requests to simplify capturing this (and include any requests that don't specifically have middleware).

@dbaeumer slightly related, pre-LSP I would attach system timestamps to all outbound messages to the language server. The server would read them and record the difference in the current system time and the provided timestamp, which a user could view in the servers diagnostics to help track down performance issues.

Since switching to LSP - in addition to the above - there isn't really a place I can stash this timestamp to provided to the server. Would it be reasonable to have LSP support attaching some arbitrary data to each message, or some specific (optional) support for clients to add this?

vinistock:
Closes #918

Allow adding a general middleware `provideRequestOrNotification` to client options that is used for all requests and notifications. The attached issue has a bit more context, but this is useful for measuring performance and even appending additional parameters to requests if desired.

Please let me know if this is the right direction for the implementation and if I'm following the right style and naming conventions. I wasn't sure if this was the right level for the implementation, but the general middleware can alternatively be implemented as a connection option instead.

### DIFF
--- a/client/src/common/client.ts
+++ b/client/src/common/client.ts
@@ -319,6 +319,23 @@ interface _Middleware {
 	window?: WindowMiddleware;
 }
 
+// A general middleware is applied to both requests and notifications
+interface GeneralMiddleware {
+	sendRequest?<R>(
+		this: void,
+		type: string | MessageSignature,
+		next: (type: string | MessageSignature, ...params: any[]) => Promise<R>,
+		...params: any[]
+	): Promise<R>;
+
+	sendNotification?<R>(
+		this: void,
+		type: string | MessageSignature,
+		next: (type: string | MessageSignature, params?: R) => Promise<void>,
+		params: R
+	): Promise<void>;
+}
+
 /**
  * The Middleware lets extensions intercept the request and notifications send and received
  * from the server
@@ -327,7 +344,7 @@ export type Middleware = _Middleware & TextDocumentSynchronizationMiddleware & C
 DocumentHighlightMiddleware & DocumentSymbolMiddleware & WorkspaceSymbolMiddleware & ReferencesMiddleware & TypeDefinitionMiddleware & ImplementationMiddleware &
 ColorProviderMiddleware & CodeActionMiddleware & CodeLensMiddleware & FormattingMiddleware & RenameMiddleware & DocumentLinkMiddleware & ExecuteCommandMiddleware &
 FoldingRangeProviderMiddleware & DeclarationMiddleware & SelectionRangeProviderMiddleware & CallHierarchyMiddleware & SemanticTokensMiddleware &
-LinkedEditingRangeMiddleware & TypeHierarchyMiddleware & InlineValueMiddleware & InlayHintsMiddleware & NotebookDocumentMiddleware & DiagnosticProviderMiddleware;
+LinkedEditingRangeMiddleware & TypeHierarchyMiddleware & InlineValueMiddleware & InlayHintsMiddleware & NotebookDocumentMiddleware & DiagnosticProviderMiddleware & GeneralMiddleware;
 
 export type LanguageClientOptions = {
 	documentSelector?: DocumentSelector | string[];
@@ -691,7 +708,12 @@ export abstract class BaseLanguageClient implements FeatureClient<Middleware, La
 		if (this._didChangeTextDocumentFeature!.syncKind === TextDocumentSyncKind.Full) {
 			await this.sendPendingFullTextDocumentChanges(connection);
 		}
-		return connection.sendRequest<R>(type, ...params);
+
+		const _sendRequest = this._clientOptions.middleware?.sendRequest;
+
+		return _sendRequest
+			? _sendRequest<R>(type, connection.sendRequest.bind(connection), ...params)
+			: connection.sendRequest<R>(type, ...params);
 	}
 
 	public onRequest<R, PR, E, RO>(type: ProtocolRequestType0<R, PR, E, RO>, handler: RequestHandler0<R, E>): Disposable;
@@ -772,7 +794,12 @@ export abstract class BaseLanguageClient implements FeatureClient<Middleware, La
 		if (openNotification !== undefined) {
 			this._pendingOpenNotifications.delete(openNotification);
 		}
-		return connection.sendNotification(type, params);
+
+		const _sendNotification = this._clientOptions.middleware?.sendNotification;
+
+		return _sendNotification
+			? _sendNotification(type, connection.sendNotification.bind(connection), params)
+			: connection.sendNotification(type, params);
 	}
 
 	public onNotification<RO>(type: ProtocolNotificationType0<RO>, handler: NotificationHandler0): Disposable;
--- a/client/src/common/client.ts
+++ b/client/src/common/client.ts
@@ -35,7 +35,8 @@ import {
 	ImplementationRequest, SelectionRangeRequest, TypeDefinitionRequest, CallHierarchyPrepareRequest, SemanticTokensRegistrationType, LinkedEditingRangeRequest,
 	TypeHierarchyPrepareRequest, InlineValueRequest, InlayHintRequest, WorkspaceSymbolRequest, TextDocumentRegistrationOptions, FileOperationRegistrationOptions,
 	ConnectionOptions, PositionEncodingKind, DocumentDiagnosticRequest, NotebookDocumentSyncRegistrationType, NotebookDocumentSyncRegistrationOptions, ErrorCodes,
-	MessageStrategy, DidOpenTextDocumentParams, CodeLensResolveRequest, CompletionResolveRequest, CodeActionResolveRequest, InlayHintResolveRequest, DocumentLinkResolveRequest, WorkspaceSymbolResolveRequest
+	MessageStrategy, DidOpenTextDocumentParams, CodeLensResolveRequest, CompletionResolveRequest, CodeActionResolveRequest, InlayHintResolveRequest, DocumentLinkResolveRequest, WorkspaceSymbolResolveRequest,
+	CancellationToken as ProtocolCancellationToken
 } from 'vscode-languageserver-protocol';
 
 import * as c2p from './codeConverter';
@@ -321,11 +322,12 @@ interface _Middleware {
 
 // A general middleware is applied to both requests and notifications
 interface GeneralMiddleware {
-	sendRequest?<R>(
+	sendRequest?<P, R>(
 		this: void,
 		type: string | MessageSignature,
-		next: (type: string | MessageSignature, ...params: any[]) => Promise<R>,
-		...params: any[]
+		param: P | undefined,
+		token: CancellationToken | undefined,
+		next: (type: string | MessageSignature, param?: P, token?: CancellationToken) => Promise<R>,
 	): Promise<R>;
 
 	sendNotification?<R>(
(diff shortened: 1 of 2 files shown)


client/src/common/client.ts
```diff
@@ -317,6 +317,15 @@ interface _Middleware {
 	window?: WindowMiddleware;
 }
 
+// A general middleware is applied to both requests and notifications
+interface GeneralMiddleware {
+	provideRequestOrNotification?<R>(
```
dbaeumer, line 322: I would prefer having a separate hook for notifications and requests. I would also like to suggest to not name it provide since they don't provide anything. Better would be something like:

```
  sendRequest(....)
  sendNotification(...)
```

client/src/common/client.ts
```diff
@@ -317,6 +317,15 @@ interface _Middleware {
...
+		this: void,
+		sendRequestOrNotification: (type: string | MessageSignature, ...params: any[]) => Promise<R>,
```
dbaeumer, line 324: We should also follow the style of the other middleware and make the next handler the last argument and call it next

client/src/common/client.ts
```diff
@@ -750,6 +766,13 @@ export abstract class BaseLanguageClient implements FeatureClient<Middleware, La
 				exclude = (params as DidOpenTextDocumentParams)?.textDocument.uri;
 			}
 			await this.sendPendingFullTextDocumentChanges(connection, exclude);
+
+			const generalMiddleware = this._clientOptions.middleware?.provideRequestOrNotification;
+
+			if (generalMiddleware) {
```
dbaeumer, line 772: I started to make checks explicit (e.g. generalMiddleware !== undefined) since it is easier to read.

dbaeumer (changes requested):

client/src/common/client.ts
vinistock: Separated the two and renamed them.

client/src/common/client.ts
vinistock: I did rename it to `next`, however I kept it second to last to be able to continue to use the rest parameters `...params`.  I made an attempt to not use a rest parameter, but ended up hitting [this error in connection](https://github.com/microsoft/vscode-languageserver-node/blob/84b4af76874ef32a0edb62e33dd2ad9a700c33ea/jsonrpc/src/common/connection.ts#L1230).

Let me know if you have an idea on how to restructure so that the rest parameter is not needed.

dbaeumer:
@vinistock this looks all good to me. Could you please rebase the PR on latest from main. I needed to change the two methods a little bit due to a serious bug with document synchronization. Thanks Dirk

client/src/common/client.ts
```diff
@@ -689,6 +706,13 @@ export abstract class BaseLanguageClient implements FeatureClient<Middleware, La
 		if (this._didChangeTextDocumentFeature!.syncKind === TextDocumentSyncKind.Full) {
 			await this.sendPendingFullTextDocumentChanges(connection);
 		}
+
+		const generalMiddleware = this._clientOptions.middleware?.sendRequest;
```
dbaeumer, line 710: To make this consistent with other implementation can we do the following:

- rename to _sendrequest
- then use a ? operation like here: https://insiders.vscode.dev/github/microsoft/vscode-languageserver-node/blob/d859bb14d1bcb3923eecaf0ef587e55c48502ccc/client/src/common/completion.ts#L130

client/src/common/client.ts
```diff
@@ -770,6 +794,13 @@ export abstract class BaseLanguageClient implements FeatureClient<Middleware, La
 		if (openNotification !== undefined) {
 			this._pendingOpenNotifications.delete(openNotification);
 		}
+
```
dbaeumer, line 797: Same as above

dbaeumer (changes requested):

client/src/common/client.ts
dbaeumer: I meant _sendRequest

vinistock: Done!

vinistock:
Not sure why the windows build timed out, but I assume it's not related to these changes. Let me know if that's not the case.

client/src/common/client.ts
```diff
@@ -317,6 +317,23 @@ interface _Middleware {
 	window?: WindowMiddleware;
 }
 
+// A general middleware is applied to both requests and notifications
+interface GeneralMiddleware {
+	sendRequest?<R>(
+		this: void,
+		type: string | MessageSignature,
+		next: (type: string | MessageSignature, ...params: any[]) => Promise<R>,
+		...params: any[]
```
dbaeumer, line 326: Actually thinking about this signature again it is not very client friendly. The reason is that the params array is hard to interpret in that form for a middleware since it can have the following combinatation:

- length 0: no param not token
- length 1: (no param and a token) or (a param no token)
- lenght 2: param and token

I have no good idea yet on how to improve this.

If all we want is to support some logging when sendRequest / sendNotification is invoked we might want to add a different API than a middleware which should allow interprestation / modification.

dbaeumer: @vinistock any idea regarding the above comment.

dbaeumer: One idea could be to have a signature like

```ts
(type: string | MessageSignature, param: any | undefined, token: CancellationToken | undefined) => Promise<R>
```

and then unfold the `params` and fold them again in the `next` handler.

vinistock: The main use case for us is for doing performance and error metrics. Something like this
```typescript
const clientOptions = {
  middleware: {
    sendRequest: () => {
      try {
        // Benchmark running the request (invoke next with params)
      } catch (error) {
        // Report error returned by the server
      }
    }
  }
}
```
For this scenario, I do think a middleware would make the most sense. It provides a lot of flexibility on what clients can do around requests and notifications.

In terms of a different API, one possibility would be to "break" the middleware into parts. This is not as flexible, and frankly feels less convenient than a middleware, but it might be easier to define an interface for. For example,
```typescript
const clientOptions = {
  beforeRequest: (type, params) => { /* start measuring performance */ },
  afterRequest: (type, params) => { /* stop measuring performance and log */ },
  onRequestError: (error, type, params) => { /* log error */ } // Could potentially be a part of `afterRequest` to avoid an extra handler
}
```

> and then unfold the params and fold them again in the next handler.

Can you explain with an example? I'm not sure I understand the approach.

dbaeumer: Not syntax or semantic checked but here is what I mean:

```ts
interface GeneralMiddleware {
	sendRequest?<P, R>(
		this: void,
		type: string | MessageSignature,
		param: P | undefined,
		token: CancellationToken | undefined,
		next: (type: string | MessageSignature, param: P | undefined, token?: CancellationToken) => Promise<R>,
	): Promise<R>;

	sendNotification?<P>(
		this: void,
		type: string | MessageSignature,
		params: P,
		next: (type: string | MessageSignature, params?: P) => Promise<void>
	): Promise<void>;
}
```

and in sendRequest

```ts
		const _sendRequest = this._clientOptions.middleware?.sendRequest;
		if (_sendRequest !== undefined) {
			const param: any | undefined = undefined;
			const token: CancellationToken | undefined = undefined;
			if (params.length === 1) {
				if (CancellationToken.is(params[0])) {
					token = params[0];
				} else {
					param = params[0];
				}
			} else if (params.length === 2) {
				param = params[0];
				token = params[1];
			}
			return _sendRequest(type, param, token, (type, param, token) => {
				const params: any[] = [];
				if (param !== undefined) {
					params.push(param);
				}
				if (token !== undefined) {
					params.push(token);
				}
				return connection.sendRequest<R>(type, ...params);
			});
		} else {
			return connection.sendRequest<R>(type, ...params);
		}
```

vinistock: I added your suggestion to the latest commit, so that it's easier to analyze and drop if necessary.

I wonder if there's a type signature we can come up for `params` that would improve the client interface without requiring this folding and unfolding of params. Something like
```typescript
// Params can be a CancellationToken, an `any` set of params, both of those or undefined
type ParamType = [CancellationToken] | [any] | [any, CancellationToken] |  undefined;

interface GeneralMiddleware {
	sendRequest?<R>(
		this: void,
		type: string | MessageSignature,
		next: (type: string | MessageSignature, params: ParamType) => Promise<R>,
		...params: ParamType,
	): Promise<R>;
}
```

client/src/common/client.ts
```diff
@@ -691,7 +709,45 @@ export abstract class BaseLanguageClient implements FeatureClient<Middleware, La
...
+
+		const _sendRequest = this._clientOptions.middleware?.sendRequest;
+		if (_sendRequest !== undefined) {
+			let param: any | undefined = undefined;
+			let token: CancellationToken | undefined = undefined;
+
+			// Separate cancellation tokens from other parameters for a better client interface
+			if (params.length === 1) {
+				// CancellationToken is an interface, so we need to check if the first param complies to it
+				if ('onCancellationRequested' in params[0] && 'isCancellationRequested' in params[0]) {
```
dbaeumer, line 721: In general we do these checks with TS type check functions. IMO you should be able to simply use: https://insiders.vscode.dev/github/microsoft/vscode-languageserver-node/blob/fca08a7e9f9bd9beeb687a5ae5529139783e915b/jsonrpc/src/common/cancellation.ts#L43

dbaeumer (changes requested):

client/src/common/client.ts
vinistock: I imported and used it instead. Let me know if you believe this is indeed the way to go and if we should do the same for notifications.

client/src/common/client.ts
```diff
@@ -38,6 +38,7 @@ import {
 	MessageStrategy, DidOpenTextDocumentParams, CodeLensResolveRequest, CompletionResolveRequest, CodeActionResolveRequest, InlayHintResolveRequest, DocumentLinkResolveRequest, WorkspaceSymbolResolveRequest
 } from 'vscode-languageserver-protocol';
 
+import { CancellationToken as JSONRPCCancellationToken } from 'vscode-jsonrpc';
```
dbaeumer, line 41: Protocol reexport vscode-jsonrpc. So you should get that type from `vscode-languageserver-protocol`

dbaeumer (changes requested):

client/src/common/client.ts
vinistock: Done.

dbaeumer (approved):

aeschli (approved):

vinistock:
@dbaeumer do we need to do the same separation for notifications as well? Let me know and I can follow up with another PR.

dbaeumer:
@vinistock what exactly do you mean?

vinistock:
Sorry, I just realized I got confused. I meant if we needed to do the folding and unfolding of parameters for notifications too, but I don't think notifications have a cancellation token.

dbaeumer:
Correct, notifications have no token.

vinistock:
Wanted to report back that I tested this on `8.2.0-next.0` and it's working well for our purposes. [Here](https://github.com/Shopify/vscode-ruby-lsp/pull/571) are the changes in our client if you'd like to see an example usage.

ejgallego:
That's exactly what we needed for coq-lsp to actually instrument the calls to our server worker, thanks!

PR #1160, opened 2022-12-25
Better error stack for response errror

chemzqm:
It's hard to trace which request causes the response error when the error is not handled.


dbaeumer:
@chemzqm to my understanding `Error.captureStackTrace` is not general available in JS runtimes. The code that is in common should only make use of the standard ES features. Normally you should have gotten a compile error when using it, but there was an oversight in the tsconfig files that I now fixed.

To move forward we can do the following:

- create our own definition of `Error.captureStackTrace` which makes the function optional
- check if the function exist before calling it

dbaeumer (changes requested):
See comment above.

PR #1164, opened 2023-01-03
Fix resolveCodeAction middleware call

alanz:
Change the name of the local resolveCodeAction function so it is actually called, otherwise the original is called instead, and the middleware is not invoked.

This matches what happens for the provideCodeAction call too, in the section above it.

### DIFF
--- a/client/src/common/codeAction.ts
+++ b/client/src/common/codeAction.ts
@@ -99,7 +99,7 @@ export class CodeActionFeature extends TextDocumentLanguageFeature<boolean | Cod
 				? (item: VCodeAction, token: CancellationToken) => {
 					const client = this._client;
 					const middleware = this._client.middleware;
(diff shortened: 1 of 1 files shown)


dbaeumer:
Which `original` do you mean. The `resolveCodeAction` here https://github.com/microsoft/vscode-languageserver-node/blob/bb6a6ac6b70dd876d2ea90f12d86a1be5414f440/client/src/common/codeAction.ts#L98  is a property which you need to call on the provider.

In addition there is a test case here https://github.com/microsoft/vscode-languageserver-node/blob/bb6a6ac6b70dd876d2ea90f12d86a1be5414f440/client-node-tests/src/integration.test.ts#L424 that ensure that the middleware is called.

Am I missing something?

alanz:
I just realised I am using an internal package manager giving me version 8.0.0.next.4.  I will update and see if I still have issues after that.

Issue #1133, opened 2022-12-01
formatting requests/didChange/didOpen appear to be sent out of order for global search-and-replace + formatOnSave


laurit17:
Hello!

I am an author of the `trunk` extension (https://marketplace.visualstudio.com/items?itemName=Trunk.io). We have users reporting bugs that when they set the trunk extension as the default formatter and run enable formatOnSave, they have global search-and-replace reverted by the extension.

It appears this is due to requests and notifications sent out of order. Here is an example trace (on a previously closed file) -
```
[Trace - 11:39:28 AM] Sending request 'textDocument/formatting - (207)'.
Params: {
    "textDocument": {
        "uri": "file:///home/lauri/work/trunk/trunk/cli/actions/foo.hh"
    },
    "options": {
        "tabSize": 4,
        "insertSpaces": true
    }
}


[Trace - 11:39:28 AM] Sending notification 'textDocument/didChange'.
Params: {
    "textDocument": {
        "uri": "file:///home/lauri/work/trunk/trunk/cli/actions/foo.hh",
        "version": 84
    },
    "contentChanges": [
        {
            "text": "#pragma once\n\n// asdfkjghatjhkjghakasdlfkjaslrakjglaksdjtlawkjglaskdjgasdgaljtlkaewjglakjgawlekasdlgkjasdglkj asdfkjghatjhkjghakasdlfkjaslrakjglaksdjtlawkjglaskdjgasdgaljtlkaewjglakjgawlekasdlgkjasdglkj asdfkjghatjhkjghakasdlfkjaslrakjglaksdjtlawkjglaskdjgasdgaljtlkaewjglakjgawlekasdlgkjasdglkj\nnamespace trunk::bar {}  // namespace trunk::bar\n"
        }
    ]
}


[Trace - 11:39:28 AM] Sending notification 'textDocument/didOpen'.
Params: {
    "textDocument": {
        "uri": "file:///home/lauri/work/trunk/trunk/cli/actions/foo.hh",
        "languageId": "cpp",
        "version": 83,
        "text": "#pragma once\n\n// abcdef abcdef abcdef\nnamespace trunk::bar {}  // namespace trunk::bar\n"
    }
}


[Trace - 11:39:29 AM] Received response 'textDocument/formatting - (207)' in 1132ms.
Result: [
    {
        "newText": "#pragma once\n\n// abcdef abcdef abcdef\nnamespace trunk::bar {}  // namespace trunk::bar",
        "range": {
[...]

PR #1168, opened 2023-01-10
Client: Fix documentSelector being overwritten

fhammerschmidt:
Thank you for fixing #1133 and releasing it with "8.1.0-next.4". It fixed the search & replace functionality for [rescript-vscode](https://marketplace.visualstudio.com/items?itemName=chenglou92.rescript-vscode).

Unfortunately that version broke semantic highlighting for us. We supply a `documentSelector` object from the client side which got overwritten by the server side `capability.documentSelector` which is set to `null` in our case.

I think this is the correct way to merge the capabilities with the `documentSelector`, with the client being last, because it says in the docs: 
> "A document selector to identify the scope of the registration. If set to null\nthe document selector provided on the client side will be used."

Source: https://github.com/microsoft/vscode-languageserver-node/blob/main/protocol/metaModel.json#L2386

### DIFF
--- a/client/src/common/features.ts
+++ b/client/src/common/features.ts
@@ -532,7 +532,7 @@ export abstract class TextDocumentLanguageFeature<PO, RO extends TextDocumentReg
 			const id = StaticRegistrationOptions.hasId(capability) ? capability.id : UUID.generateUuid();
 			const selector = capability.documentSelector || documentSelector;
 			if (selector) {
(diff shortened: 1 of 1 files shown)


fhammerschmidt:
I see you already did a more thorough fix in #1171, thank you!
Will close this one then.

dbaeumer:
Yes. I also wrote test to not break it.

Issue #1101, opened 2022-10-13
Feature request: mechanism for intercepting `client/registerCapability`


EhabY:
This is similar to https://github.com/microsoft/vscode-languageserver-node/issues/671 as we need to intercept the register capability notification for integration tests.

If there is already a way to do that then please point me in the right direction.

EhabY:
Fixed in https://github.com/microsoft/vscode-languageserver-node/pull/1179

PR #1179, opened 2023-01-15
Added registerCapability and unregisterCapability to the client middleware

EhabY:
As mentioned in https://github.com/microsoft/vscode-languageserver-node/issues/1101, currently there is no way to intercept `client/registerCapability`/`client/unregisterCapability`. This PR adds both of these requests to the middleware.

Not entirely sure if if these calls should be placed directly in `_Middleware` so any help is appreciated 🙏

### DIFF
--- a/client/src/common/client.ts
+++ b/client/src/common/client.ts
@@ -1817,14 +1817,14 @@ export abstract class BaseLanguageClient implements FeatureClient<Middleware, La
 
 	private async handleRegistrationRequest(params: RegistrationParams): Promise<void> {
 		const middleware = this.clientOptions.middleware?.handleRegisterCapability;
-		if(middleware) {
-			return middleware(params, nextParams => this.tryToRegisterCapability(nextParams));
+		if (middleware) {
+			return middleware(params, nextParams => this.doRegisterCapability(nextParams));
 		} else {
-			return this.tryToRegisterCapability(params);
+			return this.doRegisterCapability(params);
 		}
 	}
 
-	private async tryToRegisterCapability(params: RegistrationParams): Promise<void> {
+	private async doRegisterCapability(params: RegistrationParams): Promise<void> {
 		// We will not receive a registration call before a client is running
 		// from a server. However if we stop or shutdown we might which might
 		// try to restart the server. So ignore registrations if we are not running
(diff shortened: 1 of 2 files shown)


EhabY:

client/src/common/client.ts
```diff
@@ -313,6 +313,8 @@ export interface HandleWorkDoneProgressSignature {
 interface _Middleware {
 	handleDiagnostics?: (this: void, uri: Uri, diagnostics: VDiagnostic[], next: HandleDiagnosticsSignature) => void;
 	handleWorkDoneProgress?: (this: void, token: ProgressToken, params: WorkDoneProgressBegin | WorkDoneProgressReport | WorkDoneProgressEnd, next: HandleWorkDoneProgressSignature) => void;
+	registerCapability?: (this: void, params: RegistrationParams, next: RegistrationRequest.HandlerSignature) => Promise<void>;
```
dbaeumer, line 316: To keep things consistent we should name these `handle*`

dbaeumer (changes requested):

dbaeumer:
The rest looks fine with me.

client/src/common/client.ts
EhabY: Changed it to `handle*`

EhabY:
@dbaeumer Please re-review 🙏

client/src/common/client.ts
```diff
@@ -1814,6 +1816,15 @@ export abstract class BaseLanguageClient implements FeatureClient<Middleware, La
 	}
 
 	private async handleRegistrationRequest(params: RegistrationParams): Promise<void> {
+		const middleware = this.clientOptions.middleware?.handleRegisterCapability;
+		if(middleware) {
```
dbaeumer, line 1820: Formatting :-) We use space after if.

client/src/common/client.ts
```diff
@@ -1814,6 +1816,15 @@ export abstract class BaseLanguageClient implements FeatureClient<Middleware, La
...
+			return middleware(params, nextParams => this.tryToRegisterCapability(nextParams));
+		} else {
+			return this.tryToRegisterCapability(params);
+		}
+	}
+
+	private async tryToRegisterCapability(params: RegistrationParams): Promise<void> {
```
dbaeumer, line 1827: I would not call this `try*` since it is not a try in the classical sense. I would suggest `do*`

client/src/common/client.ts
```diff
@@ -1846,7 +1858,16 @@ export abstract class BaseLanguageClient implements FeatureClient<Middleware, La
...
-		for (let unregistration of params.unregisterations) {
+		const middleware = this.clientOptions.middleware?.handleUnregisterCapability;
+		if(middleware) {
+			return middleware(params, nextParams => this.tryToUnregisterCapability(nextParams));
+		} else {
+			return this.tryToUnregisterCapability(params);
+		}
+	}
+
+	private async tryToUnregisterCapability(params: UnregistrationParams): Promise<void> {
```
dbaeumer, line 1869: Same as above

dbaeumer (changes requested):

dbaeumer:
Soory for the delay. But I first wanted to get 8.1.0 out of the door before adding new things.

client/src/common/client.ts
EhabY: Fixed!

dbaeumer (approved):

aeschli (approved):

PR #1193, opened 2023-02-19
example console.log

sean-codes:
hello, I am still learning to understand how RPC works. And trying this example I was not able to see the console.log. 

Then I tried to send data back from the server and was able to log it!

I was curious would this be an okay change to the readme?


dbaeumer:
Actually this is not a good change to the readme. The example should log since it is the message that comes from the server.

sean-codes:
Thanks, I think I was confused because the comment says `// This prints Hello World`. When you run the example locally you will not see that console.log.

PR #1221, opened 2023-03-24
Fixes 1215: Language client does not automatically send file operation notification/request when working on certain path

dbaeumer:


dbaeumer:
@DanTup can you please have a look as well.

aeschli (approved):

PR #1237, opened 2023-04-21
Add 'data' to the list of item defaults for completion items.

dbaeumer:


aeschli (approved):

DanTup:
@dbaeumer is the intention here that `data` on a completion item completely replaces `itemDefaults.data`, or will the objects be merged?

If the entire object is replaced (which I suspect is the case), would you consider support for a way to merge it? In our data, we currently have some fields that are the same for a large number of completions and I'm looking at reducing payload sizes again because when running in the browser it can make large completion lists very slow (isIncomplete has some issues and VS Code doesn't support streaming results, so we tend to produce large lists because users want every symbol, imported or not).

dbaeumer:
The current strategy is that if a item has no data the one from the defaults is taken. There is currently no merge. If we want to provide merge then I would rather add another defaults field to make this very clear.

Issue #553, opened 2020-01-06
Investigate into Message Pack and GZIP support


dbaeumer:
Currently the transport uses a JSON string as a payload. To reduce the payload we could:

- zip the content of the string
- use Message Pack instead of JSON

We should investigate both options and measure the saving we can have.

This must be optional and would best be done using header properties to negotiate this as in HTTP

artas90:
CBOR format which was inspired by MessagePack seems to be better choice.
As is it a web standard https://tools.ietf.org/html/rfc7049.

dbaeumer:
@artas90 thanks for pointing this out.

AArnott:
We're looking at moving more of our remotable services in VS from JSON to MessagePack. I'm concerned about losing interoperability with node-based processes. I'd love to see `vscode-jsonrpc` support a pluggable formatter (like StreamJsonRpc has) so that the encoding can be whatever the user wants. JSON can be the default formatter, and we can supply the MessagePack one if it isn't in-box as an option as well.

AArnott:
@artas90 As I already have a great deal of investment and code based on MessagePack, I'm biased. But I wonder if you want to persuade me as to why CBOR is superior to MessagePack. They are both spec'd standards. One being RFC and the other spec'd elsewhere isn't particularly compelling. JSON itself at least started as a spec on https://www.json.org/ as I understand it.

AArnott:
> This must be optional and would best be done using header properties to negotiate this as in HTTP

FWIW, the format we use tends to be important to maintain across an entire RPC session. While we can change text encodings from message to message, switching between JSON and MessagePack isn't supported in StreamJsonRpc. So while you could declare the encoding to be messagepack in your HTTP headers, at least for our purposes we need to know whether MessagePack or JSON will be used up front when establishing the connection.

artas90:
@AArnott Honestly I never used neither CBOR nor MessagePack. Just wanted to mention that such standard as CBOR exists. And it might be considered as well

dbaeumer:
I rebased the work I started on master today and a first version of this is here: https://github.com/microsoft/vscode-languageserver-node/tree/dbaeumer/encodingSupport2

Implementation wise this works as follows:

- support pluggable encoders and decoders (gzip, ...)
- content format is still JSON
- handshake is done using HTTP header properties (e.g. `Accept-Encoding` and `Content-Encoding`). This only support encoding responses (which usually are the big payload)
[...]

PR #1250, opened 2023-06-02
Add content encoding to server options

vinistock:
For more context, see #553.

This PR adds allows clients to define `contentEncoder` and `contentDecoder` in their server options. This addition makes it easier for clients to opt-in to different content encodings, such as msgpack, without having to worry about the life cycle of the process spawned - which is necessary when passing a function.

There's a simple example usage in the integration tests, which doesn't actually do anything, but demonstrates that clients will be able to hook their own content encoders and decoders.


dbaeumer (changes requested):

dbaeumer:
I will close the PR for now. Please ping if you decide to continue working on this.

PR #1251, opened 2023-06-05
Fixes #183573: No diagnostics in JSON

dbaeumer:
Fixes #183573


client/src/common/diagnostic.ts
```diff
@@ -907,6 +907,34 @@ class DiagnosticFeatureProviderImpl implements DiagnosticProviderShape {
 			}
 		}));
 
+		tabs.onOpen((opened) => {
```
chrmarti, line 910: This looks like it might return a disposable. Should that be handled?

chrmarti (changes requested):

client/src/common/diagnostic.ts
dbaeumer: Great catch. Added to the list

chrmarti (approved):

PR #1253, opened 2023-06-07
Add formatMultipleRanges to DocumentRangeFormattingFeature

c-claeys:
Add multirange formatting support via new request textDocument/rangesFormatting
---
~This does nothing besides break `npm run compile` ATM, but I'm looking for advice. Am I missing some hook to correctly include the proposed API? I received the proposed .d.ts via `npx vscode-dts dev`, but `compile` doesn't seem to pick it up AFAICT.~


c-claeys:
Sounds good. I was just looking at past discussions. Initially we were thinking of putting additional ranges in the formatting options but VSCode's API sync resulted in parallel functions exposed on that API. Any opinion over whether the formatting options should be modified to include ranges, if a new formatting method (`textDocument/rangesFormatting`) should be added, or something else?

protocol/src/common/protocol.ts
```diff
@@ -3448,10 +3448,34 @@ export interface DocumentRangeFormattingParams extends WorkDoneProgressParams {
 	options: FormattingOptions;
 }
 
+/**
```
dbaeumer, line 3451: We need a @since tag here and a @proposed tag

protocol/src/common/protocol.ts
```diff
@@ -3448,10 +3448,34 @@ export interface DocumentRangeFormattingParams extends WorkDoneProgressParams {
...
+}
+
 /**
  * Provider options for a {@link DocumentRangeFormattingRequest}.
  */
 export interface DocumentRangeFormattingOptions extends WorkDoneProgressOptions {
+	/**
+     * Whether the server supports formatting multiple ranges at once.
+     */
+	canFormatMultipleRanges?: boolean;
```
dbaeumer, line 3478: to make things consitent with other name and the client capability I would name it `multipleRangesSupport` or `rangesSupport`

dbaeumer (changes requested):

protocol/src/common/protocol.ts
c-claeys: Done.

protocol/src/common/protocol.ts
c-claeys: Done.

c-claeys:
> Otherwise (as always) great PR.

Cheers! Thanks for the reviews

dbaeumer:
PR looks good. @c-claeys are you providing a spec change as well?

dbaeumer (approved):

jrieken (approved):

c-claeys:
Sure, happy to follow up with a spec change PR

PR #1264, opened 2023-07-11
Add `MessageType.Debug`

### LINKED ISSUE #1255: Feature request: Improve formatting of debug messages in `window/logMessage`
Currently, the only modes supported by LSP are `Error`, `Warning`, `Info` and `Log` (see [the spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessage)). While `Error`, `Warning` and `Info` messages are logged with a timestamp and their log level, `Log` messages, which seem to be the only option for debug logging, are forwarded directly to the output channel:

https://github.com/microsoft/vscode-languageserver-node/blob/43b7c7265de6b0662cfe58576c9d7b894a946ce3/client/src/common/client.ts#L1096-L1109

This unfortunately causes the output to look a bit messy due to inconsistent formatting. For example, my language server often produces logs like this:

```
[Info  - 2:05:32 PM] An info message
A debug message
Another debug message
[Info  - 2:05:33 PM] Another info message
[Info  - 2:05:35 PM] Another info message
```

Perhaps this could either be addressed here on the VSCode-side by formatting generic log messages in a similar style, e.g.

```
[Info  - 2:05:32 PM] An info message
[Log   - 2:05:32 PM] A debug message
[Log   - 2:05:32 PM] Another debug message
[Info  - 2:05:33 PM] Another info message
[Info  - 2:05:35 PM] Another info message
```

or by introducing a dedicated `MessageType.Debug` level on the LSP-side:

```
[Info  - 2:05:32 PM] An info message
[Debug - 2:05:32 PM] A debug message
[Debug - 2:05:32 PM] Another debug message
[Info  - 2:05:33 PM] Another info message
[Info  - 2:05:35 PM] Another info message
```

The latter solution has the additional advantage that it would leave servers that rely on the current formatting (e.g. because they already format messages in their own way) intact.

fwcd:
This fixes #1255 by adding `MessageType.Debug` (along with `RevealOutputChannelOn.Debug`) for debug messages.

The new enum cases are added in a backwards-comaptible way that shouldn't break any assumptions about ordering (e.g. debug level is conceptually "more fine-grained" than info/warn/error).


protocol/src/common/protocol.ts
```diff
@@ -1538,9 +1538,13 @@ export namespace MessageType {
 	 * A log message.
 	 */
 	export const Log = 4;
+	/**
```
dbaeumer, line 1541: We need to add a @since 3.18.0 tag here.

server/src/common/server.ts
```diff
@@ -162,6 +162,13 @@ export interface RemoteConsole extends FeatureBase {
 	 * @param message The message to log.
 	 */
 	log(message: string): void;
+
+	/**
+	 * Log a debug message.
+	 *
+	 * @param message The message to log.
```
dbaeumer, line 169: same here

dbaeumer:
Besides a PR for the implementation we also need one for the spec which is hosted here: https://github.com/microsoft/language-server-protocol. The branch to make the modifications in is gh-pgaes. Can you please provide a corresponding PR as well.

dbaeumer (changes requested):

dbaeumer (approved):

aeschli (approved):

Issue #1272, opened 2023-07-17
Feature request: mechanism for intercepting `workspace/applyEdit`


EhabY:
This is needed in our case because we want to save the changes when applying edits as VS Code does not offer that possibility (sadly).

dbaeumer:
PR welcome.

PR #1273, opened 2023-07-17
add applyEdit to the workspace middleware

EhabY:
Fix for https://github.com/microsoft/vscode-languageserver-node/issues/1272


davidmedinasigasi:
I would really like this to be merged

dbaeumer:
Was out on vacation for a while. Will have a look in August.

EhabY:
Pinging @dbaeumer since it's a smol change with big impact (for us) 🙏

client/src/common/client.ts
```diff
@@ -297,6 +297,7 @@ export interface DidChangeWatchedFileSignature {
 
 type _WorkspaceMiddleware = {
 	didChangeWatchedFile?: (this: void, event: FileEvent, next: DidChangeWatchedFileSignature) => Promise<void>;
+	applyEdit?: (this: void, params: ApplyWorkspaceEditParams, next: ApplyWorkspaceEditRequest.HandlerSignature) => Promise<ApplyWorkspaceEditResult>;
```
dbaeumer, line 300: For consistency the middleware should be called handleApplyEdit

dbaeumer:
Ad could you please add a test. The client part for integration tests is here https://github.com/microsoft/vscode-languageserver-node/blob/4b77bb4d2e47dbfeeed6c263c28054eced60e030/client-node-tests/src/integration.test.ts#L1 and the server part is here https://github.com/microsoft/vscode-languageserver-node/blob/4b77bb4d2e47dbfeeed6c263c28054eced60e030/client-node-tests/src/servers/testServer.ts#L1

An example on how to implement a test for a request that flows from server to client is here: https://github.com/microsoft/vscode-languageserver-node/blob/4b77bb4d2e47dbfeeed6c263c28054eced60e030/client-node-tests/src/integration.test.ts#L532

dbaeumer (changes requested):

client/src/common/client.ts
EhabY: Changed!

client-node-tests/src/integration.test.ts
```diff
@@ -1515,6 +1515,43 @@ suite('Client integration', () => {
...
+			setImmediate(currentProgressResolver);
+			/**
+			 * TODO signatures need to be updated also for showDocument, handleRegisterCapability, handleUnregisterCapability
+			 * The use of `HandlerSignature` makes it that you have to pass a dummy cancellation token and handle `ResponseError`
+			 */
+			const resultOrError = await next(params, tokenSource.token);
+			if(resultOrError instanceof lsclient.ResponseError) {
+				return Promise.reject(resultOrError.message);
+			}
+			return resultOrError;
```
EhabY, line 1536: It seems using `HandlerSignature` is not ideal for middlewares. We essentially have to pass a null `CancellationToken`  and handle `ResponseError`. Requests sent from the server to the client like `applyEdit`, `showDocument`, `handleRegisterCapability`, `handleUnregisterCapability` now have to do this workaround...

Am I using it wrong or is this expected?

dbaeumer: The token we have to provide since this is how things are speced right now. There could be servers that provide cancellation for message send from the server.

For the return value I think we should simply declare the middleware as returning a `HandlerResult`. The a middle ware implementation doesn't need to handle the ResponseError.

Let me know if this works

EhabY: Is it possible to make the token optional then? Requiring the token makes the code a bit awkward.

Just returned a `HandlerResult` and updated the middleware handling such that it returns a rejected promise. Should I update the middlewares of `showDocument`, `handleRegisterCapability`, `handleUnregisterCapability` such that they return a `HandlerResult` too (+ simple tests for sanity)?

dbaeumer (approved):

client-node-tests/src/integration.test.ts
dbaeumer: I decided to leave the middle ware as is for now since technically this might be breaking and I am not sure if it is worth the effort right now.

IMO the token is only a problem when writing the tests. In a real world scenario the token should already be provided.

aeschli (approved):

hediet (approved):

hediet (approved):

client-node-tests/src/integration.test.ts
EhabY: That's fair, but I tried implementing the middleware `handleRegisterCapability` in production and there is no way to propagate the cancellation token so I have to create a throwaway one.

I would expect the middleware to be able to follow this format: `handleX: (...args, next) => return next(...args)`.

For example, the `didOpen`:

```javascript
didOpen: async (data, next) => {
    return next(data)
}
```

PR #1296, opened 2023-08-18
Update to package-lock.json v3

remcohaszing:
v2 is an interop format that includes both v1 and v3. Updating to v3 literally means to delete half the lockfile. As of npm 9, v3 is the default.


dbaeumer:
We don't take PR against look files since they are extremely hard to review. I re-generated the files myself.

PR #1308, opened 2023-09-06
Adding names to literal types

karthiknadig:


karthiknadig:
@dbaeumer I have added names to a few types. Please let me know if I the change I have made are going in the right direction. For example, I see that the changes I made in `types/main.ts` don't get picked up by meta mode generator until types is published and the types in protocol is updated. Is that the right way to update that?

dbaeumer:
They do go into the right direction and they should be picked up in the generation. Did you run `npm run symlink` in the root of the directory?

karthiknadig:
Yes, I did. I am doing this on Windows.

When i run `npm install` for `protocol` it still pulls in the published version of the package.

dbaeumer:
> When i run npm install for protocol it still pulls in the published version of the package.

`npm install` removes all symbolic links again. So you always need to do `npm install & npm run symlink`

karthiknadig:
That worked. Thanks, I will update rest of the literal types.

karthiknadig:
I have converted all of the "literal" kinds except for the one that are like below:
```
                                        "name": "range",
					"type": {
						"kind": "or",
						"items": [
							{
								"kind": "base",
								"name": "boolean"
							},
							{
								"kind": "literal",
								"value": {
									"properties": []
								}
							}
						]
					},
```
I am assuming the intent here was the allow future extensibility of type for this field. We could leave it as is.

dbaeumer:
@karthiknadig could you please undo the change in the package.json. In general I only change version numbers when releasing stuff in our releaser pipeline. And although in the metaModel.json since I generate that when I release a new version.

This makes reviewing the PR and merging a lot easier. Thanks!

types/src/main.ts
```diff
@@ -2389,6 +2413,14 @@ export namespace CompletionList {
 	}
 }
 
+/**
+ * @deprecated use MarkupContent instead.
+ */
+export interface MarkedStringDetail {
```
dbaeumer, line 2419: Detail is not a good name. MarkedStringWithLanguage ?

And although deprecated we should add @since and @proposed.

protocol/src/common/protocol.ts
```diff
@@ -1275,28 +1343,49 @@ export interface ServerCapabilities<T = LSPAny> {
...
+	/**
+	 * The server's version as defined by the server.
+	 */
+	version?: string;
+}
+
+
+/**
+ * Information about the client
+ *
```
dbaeumer, line 1374: We should add @since 3.18.0 own type name or something to indicate the change we did.

dbaeumer: Same for ServerInfo and all other types we introduce that have already a @since tag

dbaeumer (changes requested):
In general please add @proposed to all

protocol/src/common/protocol.ts
```diff
@@ -127,45 +127,66 @@ import { InlineCompletionClientCapabilities, InlineCompletionOptions, InlineComp
...
 	pattern?: string;
-} | {
+}
+
+/**
+ * A document filter where `pattern` is required field.
+ *
+ * @since 3.18.0
+ */
+export interface TextDocumentFilterPattern {
```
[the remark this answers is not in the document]
karthiknadig: Yep. Fixed this.

karthiknadig:
I think we might have to update the `since` type in the schema to be `string[]`. I can do that in a separate PR.

michaelpj:
I'm not sure that marking these as "proposed" was a good idea, see: https://github.com/microsoft/vscode-languageserver-node/issues/1416

Issue #1306, opened 2023-08-30
Folding range doesn't offer /refresh


c-claeys:
Several methods language feature APIs in VSCode, written as providers, have did change events as options: `onDidChangeSemanticTokens`, `onDidChangeCodeLenses`, `onDidChangeInlineValues`, `onDidChangeFoldingRanges`, and `onDidChangeInlayHints`. All of these, except for folding range, have corresponding refresh support in the spec and this repo.

Are you open to a PR to add this to foldingRange.ts etc?

dbaeumer:
Absolutely. This is a pure oversight.

PR #1309, opened 2023-09-12
Add the workspace/foldingRange/refresh method

c-claeys:
Relating to #1306

server.ts has `onFoldingRanges` already which I've left alone for backwards compatibility but this PR also adds `languages.foldingRange.on` which should be identical.


dbaeumer:
@c-claeys very nice work. Thanks a lot. One minor thing. Can you mark the @since 3.18 as proposed as well. This will allow me to ship it  with 3.17.4 as a preview.

And are you working on a PR for the spec as well?

c-claeys:
Added the proposed tag. 
> And are you working on a PR for the spec as well?

Yeah it should be out for review shortly. I figured I'd ensure this goes smoothly first.

dbaeumer (approved):

alexr00 (approved):

aeschli (approved):

PR #1326, opened 2023-09-21
fix: ensure bundlers pick up esm version

AviVahl:
up until the addition of "exports", bundlers found the root "module" field and resolved the the esm version.

this change fixes a regression from https://github.com/microsoft/vscode-languageserver-node/pull/1297 where cjs browser consumers now get resolved to the umd, getting a warning about dynamic require.

### DIFF
--- a/textDocument/package.json
+++ b/textDocument/package.json
@@ -1,37 +1,38 @@
 {
-	"name": "vscode-languageserver-textdocument",
-	"description": "A simple text document implementation for Node LSP servers",
-	"version": "1.0.10",
-	"author": "Microsoft Corporation",
-	"license": "MIT",
-	"repository": {
-		"type": "git",
-		"url": "https://github.com/Microsoft/vscode-languageserver-node.git",
-		"directory": "textDocument"
-	},
-	"bugs": {
-		"url": "https://github.com/Microsoft/vscode-languageserver-node/issues"
-	},
-	"main": "./lib/umd/main.js",
-	"typings": "./lib/umd/main",
-	"exports": {
-		".": {
-			"import": "./lib/esm/main.js",
-			"default": "./lib/umd/main.js"
-		}
-	},
-	"scripts": {
-		"prepublishOnly": "echo \"⛔ Can only publish from a secure pipeline ⛔\" && node ../build/npm/fail",
-		"prepack": "npm run all:publish",
-		"compile": "node ../build/bin/tsc -b ./tsconfig.json",
-		"clean": "node ../node_modules/.bin/rimraf lib",
-		"watch": "node ../build/bin/tsc -b ./tsconfig.watch.json -w",
-		"lint": "node ../node_modules/eslint/bin/eslint.js --ext ts src",
-		"test": "node ../node_modules/mocha/bin/_mocha",
-		"all": "npm run clean && npm run compile && npm run lint && npm run test",
-		"compile:esm": "node ../build/bin/tsc -b ./tsconfig.esm.publish.json && node ../build/bin/fix-esm",
-		"compile:umd": "node ../build/bin/tsc -b ./tsconfig.umd.publish.json",
-		"all:publish": "git clean -xfd . && npm install && npm run compile:esm && npm run compile:umd && npm run lint && npm run test",
-		"preversion": "npm test"
-	}
+  "name": "vscode-languageserver-textdocument",
+  "description": "A simple text document implementation for Node LSP servers",
+  "version": "1.0.10",
+  "author": "Microsoft Corporation",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Microsoft/vscode-languageserver-node.git",
+    "directory": "textDocument"
+  },
+  "bugs": {
+    "url": "https://github.com/Microsoft/vscode-languageserver-node/issues"
+  },
+  "main": "./lib/umd/main.js",
+  "typings": "./lib/umd/main",
+  "exports": {
+    ".": {
+      "browser": "./lib/esm/main.js",
+      "import": "./lib/esm/main.js",
+      "default": "./lib/umd/main.js"
+    }
+  },
+  "scripts": {
+    "prepublishOnly": "echo \"⛔ Can only publish from a secure pipeline ⛔\" && node ../build/npm/fail",
+    "prepack": "npm run all:publish",
+    "compile": "node ../build/bin/tsc -b ./tsconfig.json",
+    "clean": "node ../node_modules/.bin/rimraf lib",
+    "watch": "node ../build/bin/tsc -b ./tsconfig.watch.json -w",
+    "lint": "node ../node_modules/eslint/bin/eslint.js --ext ts src",
+    "test": "node ../node_modules/mocha/bin/_mocha",
+    "all": "npm run clean && npm run compile && npm run lint && npm run test",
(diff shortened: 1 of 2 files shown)


dbaeumer:
@aeschli Martin, can you please have a look.

AviVahl:
This is easier to review if u ignore the whitespace changes: https://github.com/microsoft/vscode-languageserver-node/pull/1326/files?diff=unified&w=1

aeschli:
The change makes sens, but please revert the indentation changes...

AviVahl:
done. :+1:

remcohaszing:
IMO it’s a good thing there is no distinction between usage in the browser and other platforms. This could open up another set of issues.

I recommend against merging this. However, I do recommend to change the UMD bundle to a regular CJS bundle. That will also make the warning go away.

aeschli (approved):

AviVahl:
Either way works for us. If it won't point to a umd, the warning would go away.

That being said, esm allows for better tree shaking, which might be preferable for browser builds.

alexdima (approved):

AviVahl:
> This change basically says: If the `browser` condition is set, `require('vscode-languageserver-types')` call should resolve to an ESM file. So for example:
> 
> ```shell
> node --conditions browser --test
> ```
> 
> ```js
> // test.js
> // This is an error when the browser condition is used
> const types = require('vscode-languageserver-types')
> ```
> 
> This is just one example. I’m not sure how different bundlers and test runners deal with this.

Both webpack and esbuild set this condition on their resolvers when targeting browser bundles. My team already saw several npm packages using this out there in the wild.

Not being able to run in node with `--conditions browser` set seems fine to me. I mean... it would also break if I use `--conditions types`, as these are used to point to `.d.ts` files.

We use these conditions to address special cases. Forcing those custom conditions in cases that work OOTB (e.g. `node` as-is) doesn't seem logical to me.

If this package was built to both esm and cjs, this "browser" condition could have had nested conditions under it, providing `import`/`require` targets which are bundle-compatible (without any dynamic require usage). That wouldn't necessarily mean we would be able to use `--conditions browser` in node, as many other npm packages don't provide both targets under the "browser" condition.

remcohaszing:
> Both webpack and esbuild set this condition on their resolvers when targeting browser bundles. My team already saw several npm packages using this out there in the wild.

I think it’s very useful to be able to do this. This allows to provide different implementations when targeting a browser or Node.js. For example, the `browser` entry point may be able to use the DOM, or the Node.js might use Node-only APIs. I don’t think it’s a good idea to use this to force `require` to resolve resolve to ESM files.

> Not being able to run in node with `--conditions browser` set seems fine to me. I mean... it would also break if I use `--conditions types`, as these are used to point to `.d.ts` files.
> 
> We use these conditions to address special cases. Forcing those custom conditions in cases that work OOTB (e.g. `node` as-is) doesn't seem logical to me.

The `browser` condition and the `types` condition are not the same thing. The `node --conditions types` deliberately breaks the program. Running `node --conditions browser` has an actual use: To test browser code using Node.js, i.e. leveraring `jsdom` and the Node.js test runner.

> If this package was built to both esm and cjs, this "browser" condition could have had nested conditions under it, providing `import`/`require` targets which are bundle-compatible (without any dynamic require usage). That wouldn't necessarily mean we would be able to use `--conditions browser` in node, as many other npm packages don't provide both targets under the "browser" condition.

Yes, it would be possible to triple-publish ESM for ESM supporting targets, CJS for bundlers, and UMD for Node.js. I don’t think anyone would want to go into that direction though.

remcohaszing:
Note that the current behaviour isn’t broken. The bundler correctly resolves from CJS to the CJS file, which happens to be a UMD file. The bundler now shows a warning, not an error.

How are you pulling in the `vscode-languageserver-types` dependency? Most of the VSCode client libraries have related issues, but solving it starts at the most low-level packages such as this one. https://github.com/microsoft/vscode/issues/192144

AviVahl:
We're directly importing it in our code, using it for integrations with monaco and vscode.

So your suggestion is to keep the new warning? In our case, it is the only warning during build of over 500 deps.

I find that is very odd, just to allow running `node --conditions browser`, which is not really used by anyone nor this project.

I think this should land as-is. The argument for not doing so is not compelling enough IMO.

remcohaszing:
This means you are consuming the imports from a faux ESM context. The solution is to switch to real ESM. Bundlers treat this pretty well nowadays, which is exactly the reason it’s resolving to the CJS entrypoint for you.

If you wish to use ESM, you should either

1. specify `"type": "module"` in your `package.json`; or
2. rename your `.js` files to `.mjs` and your `.ts` files to `.mts`.

Note that using a bundler doesn’t mean the same as targeting a browser. Your use case is a good example of this. You’re bundling a VSCode plugin, which I imagine targets Node.js, not the browser.

AviVahl:
The reality is that we have a commonjs consumer that gets bundled. Changing it to native esm isn't easy. We're talking about **huge** projects with numerous integrations. We're talking node, electron (renderer and main), web worker, worker thread, you name it.

I mean, [we're working](https://github.com/wixplosives/file-services/pull/972) [on it](https://github.com/wixplosives/engine/pull/1970), but both TypeScript and node leave a lot to be desired when migrating large projects.

Using `"type": "module"`, for example, causes typescript to misbehave in commonjs consumers with `moduleResolution: "node16"/"bundler"` set. It warns about packages being esm-only, even though they have a `"require"` condition for their main entrypoint.

aeschli:
We go with 
```
{
	"exports": {
		".": {
			"browser": "./lib/esm/main.js",
			"import": "./lib/esm/main.js",
			"default": "./lib/umd/main.js"
		}
	},
}
```
for now which is equivalent to what we had before when w had the `module` and `browser` entry points.
If `node --conditions browser` becomes popular and we get more requests, we can still consider looking at 2.

remcohaszing:
It sounds like what the OP is really looking for, is to always resolve to ESM, even from CJS. If you want this, the bundler should be configured to use the `import` condition, not the `browser` condition. This isn’t just about `vscode-languageserver-types`, it’s about all of their dual-published dependencies.

> Using `"type": "module"`, for example, causes typescript to misbehave in commonjs consumers with `moduleResolution: "node16"/"bundler"` set. It warns about hybrid packages being esm-only, even though they have a `"require"` condition for their main entrypoint.

TypeScript is really correct about this nowadays. If it warns about hybrid packages being esm-only, that means those packages are misconfigured. I know there are still a lot of misconfigured packages out there, and I’m trying to fix as much as possible all around the ecosystem. That’s one reason why I fixed it for `vscode-languageserver-types` and why I oppose the changes in this PR.

aeschli:
@remcohaszing Thanks so much for your input, Very happy to have your expertise (even if I ignored it this time). To quickly get unblocked we're going with (1.) for now as the existing build scripts for the web extensions that are currently broken. 

Would it have made sense to keep the `browser` entry point?

remcohaszing:
I understand you need to make some pragmatic choices. I fear this change will cause more issues, because it’s technically incorrect. I hope it won’t be as bad, I find this hard to predict.

Dual packaging comes with a lot of quirks. I think a huge amount of issues would already be solved if we can get rid of all UMD bundles.

I think the `browser` field is almost as bad as the `browser` export. Although browser usage pretty much implies a bundler is used, bundling doesn’t imply the code is built for the browser.

PR #1327, opened 2023-09-25
Add final newlines

remcohaszing:
This enables the VSCode option `files.insertFinalNewline` and reformats all files. In most cases this means the final newline is added. In some cases trailing whitespace was stripped as well.

The main reason to prefer inserting a final newline, is because POSIX defines a text file as a file made of lines, where each line ends with a newline character. This includes the final line.

Most tools write final newlines, i.e. npm, tsc. Git and related tools tend to render a missing newline as a sort of warning. This often leads to annoying diffs if the newline is missing.


dbaeumer:
Since the VS Code repository uses the same style I am actually in favor doing the same here. However I usually try to avoid merging huge PRs like this since they are hard to review. @remcohaszing How did you do the file changes?

dbaeumer:
I tried it and this formats a lot of other things as well :-(. Need to think about a better solution.

remcohaszing:
Anything in particular you are concerned about?

For me it only generated the changes in this PR, which are just some whitespace changes. If you see any other changes, they must be due to your own personal VSCode settings. I recommend to make those explicit as well.

dbaeumer:
It started to format json files. Although these are indeed only white spaces I would like to avoid them.

PR #1328, opened 2023-09-25
Convert from UMD to CJS

remcohaszing:
There are essentially 2 valid TypeScript module configurations.

The following is to build a code base using `tsc`. It detects whether it should emit CJS or ESM based on the parent `package.json` file. At the moment of writing `node16` is an alias of `nodenext`, but I’ve seen the TypeScript team promote `nodenext` more.

```jsonc
{
  "compilerOptions": {
    "module": "nodenext"
  }
}
```

The following is to build a code base using a bundler. It always outputs ESM.

```jsonc
{
  "compilerOptions": {
    "module": "esnext",
    "moduleResolution": "bundler."
  }
}
```

Other options are not yet officially deprecated, but are considered legacy. TypeScript on its own does not support dual publishing.

This change replaces all occurrences of `umd` with `cjs`. Next all TypeScript configurations were updated to use the `nodenext` module option, causing them to emit CJS.

The ESM configurations were changed to use the `esnext` module option. This tricks TypeScript to emit ESM, even though all workspaces are marked as a CJS project.

The main reason I’ve seen to emit UMD for some packages in the VSCode ecosystem, is because VSCode itself still uses AMD. However, this is only true for the VSCode core, which doesn’t use any of the packages in this repository.

### DIFF
--- a/.tsconfigrc.js
+++ b/.tsconfigrc.js
@@ -35,8 +35,7 @@ const general = {
 	 * rely on webpack to package everything correctly.
 	 */
 	compilerOptions: {
-		module: 'commonjs',
-		moduleResolution: 'node'
+		module: 'nodenext'
 	}
 };
 
@@ -115,7 +114,7 @@ const textDocument = {
 const textDocument_publish = {
 	name: 'textDocument_publish',
 	path: './textDocument',
-	references: [ './tsconfig.esm.publish.json', './tsconfig.umd.publish.json' ]
+	references: [ './tsconfig.esm.publish.json', './tsconfig.cjs.publish.json' ]
 };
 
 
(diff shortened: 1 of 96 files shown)


aeschli:
AMD is used by the Monaco editor in the development setup.
That said, the plan is to move away from it, but give it more time.

dbaeumer:
@aeschli should we close the PR then for now?

aeschli:
Yes, we close it for now

remcohaszing:
I understand this is closed for now. Is there an ETA when VS Code / Monaco editor will move away from AMD? That’ll make the situation a lot simpler (not just for packages in this repo, but for all VSCode related packages)

Issue #724, opened 2021-01-07
enable willRenameFiles for folder rename


ShuiRuTian:
LSP 3.16 add `willRenameFiles`.
This works well for files change. But how to enable this feature for folder rename?
For my case, the clientOption is something like 
```
   // this is not important, only means I want to log, and here is a good place to set breakpoint.
    const workspace:lc.WorkspaceMiddleware = {
        willRenameFiles:function<P extends vscode.FileRenameEvent,R extends Thenable<vscode.WorkspaceEdit | null | undefined>>(this: void, data: P, next:(data: P) => R ){
            console.log(data);
            return next(data);
        }
    }


    const clientOption = {
            documentSelector: [
                { scheme: 'file', language: 'rust' }
            ],
            ...
            middleware:{
                ...
                workspace
            }
    }

    const client = new lc.LanguageClient(
        'rust-analyzer',
        'Rust Analyzer Language Server',
        serverOptions,
        clientOptions,
    );
    ....
```
I guess the key is documentSelector? But how to select folder?

dbaeumer:
The `willRenameFiles` event is not bound to document selector. When registering you need to provide a pattern. See `FileOperationPattern` here: https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_willCreateFiles

ShuiRuTian:
Whoops, thanks so much for the kind answer. This really helps me a lot!

PR #1343, opened 2023-10-26
Snippet edit feature

MariaSolOs:
Adding support for snippet workspace edits.

Closes [#724](https://github.com/microsoft/language-server-protocol/issues/724)

### DIFF
--- a/client/src/common/protocolConverter.ts
+++ b/client/src/common/protocolConverter.ts
@@ -1088,6 +1088,8 @@ export function createConverter(uriConverter: URIConverter | undefined, trustMar
 					for (const edit of change.edits) {
 						if (ls.AnnotatedTextEdit.is(edit)) {
 							result.replace(uri, asRange(edit.range), edit.newText, asMetadata(edit.annotationId));
+						} else if (ls.SnippetTextEdit.is(edit)) {
+							result.replace(uri, asRange(edit.range), edit.snippet.value, asMetadata(edit.annotationId));
 						} else {
 							result.replace(uri, asRange(edit.range), edit.newText);
 						}
--- a/protocol/src/common/protocol.ts
+++ b/protocol/src/common/protocol.ts
@@ -4094,6 +4094,14 @@ export interface WorkspaceEditClientCapabilities {
 	 * @since 3.16.0
 	 */
 	changeAnnotationSupport?: ChangeAnnotationsSupportOptions;
+
+	/**
+	 * Whether the client supports snippets as text edits.
+	 *
+	 * @since 3.18.0
+	 * @proposed
+	 */
+	snippetEditSupport?: boolean;
 }
 
 /**
@@ -4207,7 +4215,7 @@ export {
 	DidChangeNotebookDocumentNotification, DidSaveNotebookDocumentParams, DidSaveNotebookDocumentNotification, DidCloseNotebookDocumentParams,
 	DidCloseNotebookDocumentNotification,
 	// Inline Completions
-	InlineCompletionClientCapabilities, InlineCompletionOptions, InlineCompletionParams, InlineCompletionRegistrationOptions, InlineCompletionRequest
+	InlineCompletionClientCapabilities, InlineCompletionOptions, InlineCompletionParams, InlineCompletionRegistrationOptions, InlineCompletionRequest,
 };
 
 // To be backwards compatible
--- a/types/src/main.ts
+++ b/types/src/main.ts
@@ -1026,8 +1026,11 @@ export interface TextDocumentEdit {
 	 *
 	 * @since 3.16.0 - support for AnnotatedTextEdit. This is guarded using a
 	 * client capability.
+	 *
+	 * @since 3.18.0 - support for SnippetTextEdit. This is guarded using a
+	 * client capability.
 	 */
-	edits: (TextEdit | AnnotatedTextEdit)[];
+	edits: (TextEdit | AnnotatedTextEdit | SnippetTextEdit)[];
 }
 
 /**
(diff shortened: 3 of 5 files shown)


MariaSolOs:
@dbaeumer while working on this I realized that it would be quite convenient to use [`StringValue`s](https://github.com/microsoft/vscode-languageserver-node/blob/84285312d8d9f22ee0042e709db114e5415dbdde/types/src/main.ts#L4153) in `TextEdit`s to achieve this instead of introducing a new `SnippetTextEdit` type. That is, changing the type of `textEdit.newText` to `string | StringValue`. 

The problem with that is that `StringValue` is still part of the the proposed inline completion feature, and I'm not sure about adding the dependence between the features.

dbaeumer:
@MariaSolOs I think we should stay away from changing `TextEdit`. `TextEdit` is used in many places and not all of them might support snippets. And it will be a big breaking change if clients use `TextEdit`s as results with structural typing. The TS compiler will then emit errors since a `StringValue` will not be assignable to a `string`.

protocol/src/common/protocol.ts
```diff
@@ -4094,6 +4094,13 @@ export interface WorkspaceEditClientCapabilities {
 	 * @since 3.16.0
 	 */
 	changeAnnotationSupport?: ChangeAnnotationsSupportOptions;
+
+	/**
+	 * Whether the client supports snippets as text edits.
+	 *
+	 * @since 3.18.0
```
dbaeumer, line 4101: We should add a @proposed

types/src/main.ts
```diff
@@ -1470,6 +1479,39 @@ class TextEditChangeImpl implements TextEditChange {
...
+
+	/**
+	 * The snippet to be inserted.
+	 */
+	snippet: StringValue;
+
+	/**
+	 * The actual identifier of the snippet edit.
+	 */
+	annotationId: ChangeAnnotationIdentifier;
```
dbaeumer, line 1502: The annotationId should IMO be optional.

dbaeumer (changes requested):

dbaeumer:
As always great work.

MariaSolOs:
@dbaeumer I've addressed your comments. Please let me know if I'm missing anything else :)

types/src/main.ts
```diff
@@ -4293,6 +4335,13 @@ export namespace StringValue {
 	export function createSnippet(value: string): StringValue {
 		return { kind: 'snippet', value };
 	}
+
+	export function is(value: any): value is StringValue {
```
dbaeumer, line 4339: Minor: we should call this: isSnippet in case we get more kinds in the future.

dbaeumer (commented):
As always very good work

protocol/src/common/protocol.ts
```diff
@@ -4094,6 +4094,14 @@ export interface WorkspaceEditClientCapabilities {
...
+	 * @proposed
+	 */
+	snippetEditSupport?: boolean;
```
dbaeumer, line 4104: One additional thing. We need to initialize that capability in the client somewhere here: https://github.com/microsoft/vscode-languageserver-node/blob/ccc901de19515f76521152b0282a90178271e110/client/src/common/client.ts#L1835

dbaeumer (changes requested):

dbaeumer (approved):

alexr00 (approved):

dbaeumer:
@MariaSolOs will you provide a PR for the spec as well?

types/src/main.ts
```diff
@@ -1470,6 +1479,39 @@ class TextEditChangeImpl implements TextEditChange {
...
+	annotationId?: ChangeAnnotationIdentifier;
+}
+
+export namespace SnippetTextEdit {
+	export function is(value: any): value is SnippetTextEdit {
+		const candidate = value as SnippetTextEdit;
+		return Is.objectLiteral(candidate)
+			&& Range.is(candidate.range)
+			&& StringValue.isSnippet(candidate.snippet)
+			&& (ChangeAnnotation.is(candidate.annotationId) || ChangeAnnotationIdentifier.is(candidate.annotationId));
```
DanTup, line 1511: I haven't properly looked at this so apologies if I've misunderstood, but won't this check fail if `annotationId` is undefined? The interface above for `SnippetTextEdit ` suggests it's optional, but these checks seem to require that it's either a `ChangeAnnotation` or `ChangeAnnotationIdentifier`?

client/src/common/protocolConverter.ts
```diff
@@ -1088,6 +1088,8 @@ export function createConverter(uriConverter: URIConverter | undefined, trustMar
 					for (const edit of change.edits) {
 						if (ls.AnnotatedTextEdit.is(edit)) {
 							result.replace(uri, asRange(edit.range), edit.newText, asMetadata(edit.annotationId));
+						} else if (ls.SnippetTextEdit.is(edit)) {
+							result.replace(uri, asRange(edit.range), edit.snippet.value, asMetadata(edit.annotationId));
```
rchl, line 1092: I don't know anything about vscode but wouldn't this need to handle snippet text in some special way? I would imagine that it would otherwise insert a raw, un-expanded snippet value?

MariaSolOs:
> @MariaSolOs will you provide a PR for the spec as well?

Yep, working on it.

types/src/main.ts
MariaSolOs: You're right, good catch! #1408 should fix it.

client/src/common/protocolConverter.ts
MariaSolOs: That's right, such edit should be parsed as a snippet and expanded accordingly by the editor. The capability for this feature should ensure that the editor expects these "special snippet strings".

dbaeumer: @MariaSolOs @rchl is actually correct and I totally missed that as well. This loop https://github.com/microsoft/vscode-languageserver-node/blob/2c7e6238d7fff8198f1bc0872ce46c0d40eaa6e7/client/src/common/protocolConverter.ts#L1088 needs to be recoded. I need to talk for Joh first since we currently can't mix edits with or without meta data when using the set method. 

I would opt that we disable the capability for now until we know how to actually do this correctly. @MariaSolOs I will do that.

dbaeumer: @MariaSolOs : we need to add VS Code SnippetTextEdit to the workspace edit in the snippet cases.

PR #1360, opened 2023-11-14
Fixes #1359: codeAction.resolveSupport does not have command field to detect support, but it supports it

### LINKED ISSUE #1359: `codeAction.resolveSupport` does not have `command` field to detect support, but it supports it
After https://github.com/microsoft/vscode/issues/176705 was implemented, you can async resolve both `edit` and `command` fields in code action. But `context.capabilities.textDocument.codeAction.resolveSupport.properties` array only has `"edit"`, and no `"command"` value, to support both older and newer vscode versions.

dbaeumer:
Fixes #1359

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -51,7 +51,7 @@
 		"./tsconfig-gen"
 	],
 	"editor.codeActionsOnSave": {
-		"source.fixAll.eslint": true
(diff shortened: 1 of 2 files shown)


chrmarti (approved):

RblSb:
Thanks!

PR #1380, opened 2023-12-11
Always delete feature providers in `unregister()`

werat:
The `DynamicFeature.unregister(id)` method is supposed to dispose the specified provider and delete it from the feature. In some implementations the provider is not actually deleted and stays in the storage forever. This PR fixes this.


werat:
@microsoft-github-policy-service agree company="Google"

dbaeumer:
@werat nice catch. Thanks for the PR.

dbaeumer (approved):

jrieken (approved):

PR #1391, opened 2023-12-21
Local

PoeBlu:
local build 
build debugging server


dbaeumer:
I am closing the PR. We use Azure pipelines as a local build and it can be seen here: https://dev.azure.com/ms/vscode-eslint/_build?definitionId=438

PR #1413, opened 2024-01-29
Round progress percentage according to the spec

### LINKED ISSUE #1412: Progress percentage should be an uint according to the spec
Hi folks!

@absynce and I are trying to integrate `elm-language-server` that is based on `vscode-languageserver` with [Zed](https://github.com/zed-industries/zed) and running into the following error:

```
[2024-01-25T20:44:07Z ERROR lsp] error deserializing $/progress request: Error("data did not match any variant of untagged enum ProgressParamsValue", line: 1, column: 127), message: "{\"token\":\"36abce40-63b2-4fa1-8464-030400636f26\",\"value\":{\"kind\":\"report\",\"percentage\":100.00000000000155,\"message\":\"Indexing\"}}"
```

According to [the lsp specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workDoneProgress) the `percentage` should be a uinteger. However, the actual value is a float. This works just fine if the receiving end is not strict about this, but the rust lsp implementation [assumes `percentage` to be a u32](https://docs.rs/lsp-types/latest/lsp_types/struct.WorkDoneProgressReport.html).

Would you be open for a PR that rounds the reported percentage in https://github.com/microsoft/vscode-languageserver-node/blob/c0982bf544bab9600df223146dd75c93c9004dc3/server/src/common/progress.ts#L57 ?

w0rm:
Closes #1412

### DIFF
--- a/client-node-tests/src/servers/testServer.ts
+++ b/client-node-tests/src/servers/testServer.ts
@@ -530,6 +530,14 @@ connection.onRequest(
 		void connection.sendProgress(WorkDoneProgress.type, progressToken, { kind: 'begin', title: 'Test Progress' });
 		void connection.sendProgress(WorkDoneProgress.type, progressToken, { kind: 'report', percentage: 50, message: 'Halfway!' });
 		void connection.sendProgress(WorkDoneProgress.type, progressToken, { kind: 'end', message: 'Completed!' });
+
+		// According to the spec, the reported percentage has to be an integer.
+		// Because JS doesn't have integer support, we have rounding code in place.
+		const progressToken2 = 'TEST-PROGRESS-PERCENTAGE';
+		const progress = connection.window.attachWorkDoneProgress(progressToken2);
+		progress.begin('Test Progress', 0.1);
+		progress.report(49.9, 'Halfway!');
+		progress.done();
 	},
 );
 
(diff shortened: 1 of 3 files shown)


w0rm:

client-node-tests/src/servers/testServer.ts
```diff
@@ -527,9 +527,10 @@ connection.onRequest(
 	async (_, __) => {
 		const progressToken = 'TEST-PROGRESS-TOKEN';
 		await connection.sendRequest(WorkDoneProgressCreateRequest.type, { token: progressToken });
-		void connection.sendProgress(WorkDoneProgress.type, progressToken, { kind: 'begin', title: 'Test Progress' });
```
dbaeumer, line 530: Could we please keep both tests and clarify for the test with fractions that this is outside of the spec but that we have rounding code to tolerate things since JS has not real integer support.

dbaeumer (changes requested):
See my comments about the tests.

dbaeumer:
and thanks a lot for the PR.

w0rm:
@dbaeumer thanks for the review, I've moved it into a separate test and added the comment.

dbaeumer (approved):

aeschli (approved):

PR #1421, opened 2024-02-07
Add Haskell to LanguageKind

michaelpj:
As a maintainer of the Haskell language server, I would like to be able to name my language :D

### DIFF
--- a/types/src/main.ts
+++ b/types/src/main.ts
@@ -1929,6 +1929,7 @@ export namespace LanguageKind {
 	export const Go	= 'go' as const;
 	export const Groovy	= 'groovy' as const;
 	export const Handlebars	= 'handlebars' as const;
+	export const Haskell	= 'haskell' as const;
 	export const HTML = 'html' as const;
 	export const Ini = 'ini' as const;
 	export const Java = 'java' as const;


types/src/main.ts
```diff
@@ -1929,6 +1929,7 @@ export namespace LanguageKind {
 	export const Go	= 'go' as const;
 	export const Groovy	= 'groovy' as const;
 	export const Handlebars	= 'handlebars' as const;
+	export const Haskell	= 'haskell' as const;
```
dbaeumer, line 1932: Needs @since and @proposed. See `D`

dbaeumer (changes requested):
See last comment

dbaeumer:
We also need to update the specification. Can you provide a PR for this as well?

types/src/main.ts
michaelpj: The whole enum has `@since 3.18` and `@proposed`, so are the individual entries not all implicitly the same? It seems a bit odd that only a few of the entries have those annotations: shouldn't all of them have it or none of them?

michaelpj:
Sure/

michaelpj:
The current version of the 3.18 spec doesn't have `LanguageKind` in it at all yet.

dbaeumer:
That is correct. It maintains it as a table here: https://github.com/microsoft/language-server-protocol/blob/9c5689e6f6f93102a67a652e3d9c6e2b95c2b2eb/_specifications/lsp/3.18/types/textDocumentItem.md#L32

types/src/main.ts
dbaeumer: Good point.

dbaeumer (approved):

alexr00 (approved):

PR #1423, opened 2024-02-08
Add SemanticTokenTypes.label

### LINKED ISSUE #1422: Extend predefined semantic token types to include `label`
https://github.com/microsoft/vscode-languageserver-node/blob/main/types/src/main.ts#L3805
has the list of predefined semantic token types. That matches the spec https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/

We noticed the list omits `label` which is supported by vscode.
https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#standard-token-types-and-modifiers

Our extension uses the node client library. Am I understanding it correct that there is no way for us to adjust semantic token client-side capability to customize the SemanticTokensLegend and we should make changes to the types/src/main.ts (and spec change??)

hyangah:
Visual Studio Code has 'label' in its standard token types list. https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#standard-token-types-and-modifiers

Fixes #1422

### DIFF
--- a/client/src/common/semanticTokens.ts
+++ b/client/src/common/semanticTokens.ts
@@ -74,7 +74,8 @@ export class SemanticTokensFeature extends TextDocumentLanguageFeature<boolean |
     		SemanticTokenTypes.number,
     		SemanticTokenTypes.regexp,
     		SemanticTokenTypes.operator,
-			SemanticTokenTypes.decorator
+			SemanticTokenTypes.decorator,
+			SemanticTokenTypes.label,
 		];
 		capability.tokenModifiers = [
     		SemanticTokenModifiers.declaration,
--- a/client/src/common/semanticTokens.ts
+++ b/client/src/common/semanticTokens.ts
@@ -75,7 +75,7 @@ export class SemanticTokensFeature extends TextDocumentLanguageFeature<boolean |
     		SemanticTokenTypes.regexp,
     		SemanticTokenTypes.operator,
 			SemanticTokenTypes.decorator,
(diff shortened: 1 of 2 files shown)


types/src/main.ts
```diff
@@ -3832,7 +3832,12 @@ export enum SemanticTokenTypes {
 	/**
 	 * @since 3.17.0
 	 */
-	decorator = 'decorator'
+	decorator = 'decorator',
+	/**
+	 * @since 3.18.0
+	 */
+	label = 'label'
+
```
dbaeumer, line 3840: Can you please remove the extra new line here.

client/src/common/semanticTokens.ts
```diff
@@ -74,7 +74,8 @@ export class SemanticTokensFeature extends TextDocumentLanguageFeature<boolean |
     		SemanticTokenTypes.number,
     		SemanticTokenTypes.regexp,
     		SemanticTokenTypes.operator,
-			SemanticTokenTypes.decorator
+			SemanticTokenTypes.decorator,
+			SemanticTokenTypes.label,
 		];
 		capability.tokenModifiers = [
     		SemanticTokenModifiers.declaration,
```
dbaeumer, line 81: Can you please remove the extra semicolon here.

dbaeumer (changes requested):
See comments

hyangah:
It needs 1 more collaborator approval(s).
Is there anything else I should do to unblock this issue? Thanks!

PR #1445, opened 2024-03-07
Use client ID as fallback for diagnostics collection

remcohaszing:
This sets the language client ID as a fallback for the diagnostics collection name. If no collection name is provided, VSCode generates one. This name is visible in the diagnostics table view of the problems panel.

Problems panel table view showing several diagnostics

This can be solved for every language server integration separately, but IMO it’s nicer if this package provides a better default. An alternative is to use the name instead of the ID. I went with ID, because `typescript` in the screenshot is lower case.

### DIFF
--- a/client/src/common/client.ts
+++ b/client/src/common/client.ts
@@ -1073,7 +1073,7 @@ export abstract class BaseLanguageClient implements FeatureClient<Middleware, La
 
 		// If we restart then the diagnostics collection is reused.
 		if (this._diagnostics === undefined) {
-			this._diagnostics = Languages.createDiagnosticCollection(this._clientOptions.diagnosticCollectionName || this._id);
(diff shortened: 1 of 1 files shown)


dbaeumer:
Makes sense. When we started we didn't have a table problems view :-)

remcohaszing:
@dbaeumer Friendly ping. I read your comment as you finding this a good idea. I think it makes sense to include this in v10, which is in alpha now.

dbaeumer:
@remcohaszing still agree that this is a good a change. Could you make the code adjustment I asked for.

remcohaszing:
I don’t see any review comments. Did you maybe forget to press Submit? :sweat_smile:

dbaeumer (changes requested):

dbaeumer:
Did not press the button.

dbaeumer (approved):

Tyriar (approved):

PR #1460, opened 2024-04-14
Don't crash when unregistering a feature that doesn't exist

werat:
Note: all other features follow the same practice in their `unregister()` methods -- if the features doesn't exist, do nothing and don't throw errors.


dbaeumer:
Makes sense to make this consistent.

dbaeumer (approved):

jrieken (approved):

Issue #1464, opened 2024-04-19
Diagnostics pulling doesn't work for notebook cells


rchiodo:
I'm attempting to switch Pylance to use a pull method for diagnostics but I'm not getting `textDocument/diagnostic` messages for notebook cells. 

I believe that's because notebook changes aren't tracked here:
https://github.com/microsoft/vscode-languageserver-node/blob/8e625564b531da607859b8cb982abb7cdb2fbe2e/client/src/common/diagnostic.ts#L1024

And notebook opens aren't tracked here:
https://github.com/microsoft/vscode-languageserver-node/blob/8e625564b531da607859b8cb982abb7cdb2fbe2e/client/src/common/diagnostic.ts#L962

dbaeumer:
Diagnostic pull is not available for notebook cells. It would require a special implementation in the notebook.ts file. It should work for the underlying document of the cell but that is not what you are asking for.

rchiodo:
> It should work for the underlying document of the cell but that is not what you are asking for.

Sorry, I'm not sure what you mean. Are you saying there's an alternative method to get the requests? Or are you saying something still needs to be implemented?

rchiodo:
Please take a look at my [PR](https://github.com/microsoft/vscode-languageserver-node/pull/1465). Thanks.

PR #1465, opened 2024-04-22
Support pulling diagnostics for notebooks too

rchiodo:
Fixes https://github.com/microsoft/vscode-languageserver-node/issues/1464

Please let me know if this is acceptable. Wasn't sure if formatting was correct or if I did something weird in the testing.

I also tested this using the Pylance extension to verify I do receive pull diagnostics requests for notebooks in Pylance.

### DIFF
--- a/client/src/common/diagnostic.ts
+++ b/client/src/common/diagnostic.ts
@@ -631,7 +631,7 @@ class DiagnosticRequestor implements Disposable {
 							return { kind: vsdiag.DocumentDiagnosticReportKind.unChanged, resultId: result.resultId };
 						}
 					}, (error) => {
-						return this.client.handleFailedRequest(DocumentDiagnosticRequest.type, token, error, { kind: vsdiag.DocumentDiagnosticReportKind.full, items: [] });
+						return this.client.handleFailedRequest(DocumentDiagnosticRequest.type, token, error, { kind: vsdiag.DocumentDiagnosticReportKind.full, items: [] }, true, true);
 					});
 				};
 				const middleware: DiagnosticProviderMiddleware = this.client.middleware;
--- a/client/src/common/diagnostic.ts
+++ b/client/src/common/diagnostic.ts
@@ -15,7 +15,9 @@ import {
 	ClientCapabilities, ServerCapabilities, DocumentSelector, DidOpenTextDocumentNotification, DidChangeTextDocumentNotification,
 	DidSaveTextDocumentNotification, DidCloseTextDocumentNotification, LinkedMap, Touch, RAL, TextDocumentFilter, PreviousResultId,
 	DiagnosticRegistrationOptions, DiagnosticServerCancellationData, DocumentDiagnosticParams, DocumentDiagnosticRequest, DocumentDiagnosticReportKind,
-	WorkspaceDocumentDiagnosticReport, WorkspaceDiagnosticRequest, WorkspaceDiagnosticParams, DiagnosticOptions, DiagnosticRefreshRequest, DiagnosticTag
+	WorkspaceDocumentDiagnosticReport, WorkspaceDiagnosticRequest, WorkspaceDiagnosticParams, DiagnosticOptions, DiagnosticRefreshRequest, DiagnosticTag,
+	DidChangeNotebookDocumentNotification,
+	NotebookDocumentSyncRegistrationType
 } from 'vscode-languageserver-protocol';
 
 import { generateUuid } from './utils/uuid';
@@ -970,6 +972,15 @@ class DiagnosticFeatureProviderImpl implements DiagnosticProviderShape {
 				this.diagnosticRequestor.pull(textDocument, () => { addToBackgroundIfNeeded(textDocument); });
 			}
 		}));
+		const notebookFeature = client.getFeature(NotebookDocumentSyncRegistrationType.method);
+		disposables.push(notebookFeature.onOpenNotificationSent((event) => {
+			// Send a pull for all opened cells in the notebook.
+			for (const cell of event.getCells()) {
+				if (matches(cell.document)) {
+					this.diagnosticRequestor.pull(cell.document, () => { addToBackgroundIfNeeded(cell.document); });
+				}
+			}
+		}));
 
 		disposables.push(tabs.onOpen((opened) => {
 			for (const resource of opened) {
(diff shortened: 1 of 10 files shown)


rchiodo:
@dbaeumer do you see these automatically? Sorry wasn't sure if you saw this or not.

dbaeumer:
Yes, I do but I was out sick last week :-(

client/src/common/notebook.ts
```diff
@@ -413,12 +413,26 @@ export type NotebookDocumentMiddleware = {
...
+	onChangeNotificationSent: vscode.Event<VNotebookDocumentChangeEvent>;
+	onCloseNotificationSent: vscode.Event<vscode.NotebookDocument>;
+	onSaveNotificationSent: vscode.Event<vscode.NotebookDocument>;
 	sendDidOpenNotebookDocument(notebookDocument: vscode.NotebookDocument): Promise<void>;
 	sendDidSaveNotebookDocument(notebookDocument: vscode.NotebookDocument): Promise<void>;
 	sendDidChangeNotebookDocument(event: VNotebookDocumentChangeEvent): Promise<void>;
 	sendDidCloseNotebookDocument(notebookDocument: vscode.NotebookDocument): Promise<void>;
 }
 
+export interface NotificationNotebookSendEvent<P> {
```
dbaeumer, line 426: Why do we need the two interface? They seem to be unused to me.

client/src/common/notebook.ts
```diff
@@ -427,6 +441,10 @@ class NotebookDocumentSyncFeatureProvider implements NotebookDocumentSyncFeature
 	private readonly notebookDidOpen: Set<string>;
 	private readonly disposables: vscode.Disposable[];
 	private readonly selector: vscode.DocumentSelector;
+	private readonly _onChangeNotificationSent: vscode.EventEmitter<VNotebookDocumentChangeEvent>;
```
dbaeumer, line 444: Do we really need event emitters here. As I understand it and compared to the text document synchronization client will hook the listeners on the NotebookDocumentSyncFeature. Since a NotebookDocumentSyncFeatureProvider belongs to exactly one NotebookDocumentSyncFeature can we not simply use a callback functions that the NotebookDocumentSyncFeature passes into the NotebookDocumentSyncFeatureProvider?

client/src/common/diagnostic.ts
```diff
@@ -8,20 +8,23 @@ import * as minimatch from 'minimatch';
 import {
 	Disposable, languages as Languages, window as Window, workspace as Workspace, CancellationToken, ProviderResult, Diagnostic as VDiagnostic,
 	CancellationTokenSource, TextDocument, CancellationError, Event as VEvent, EventEmitter, DiagnosticCollection, Uri, TabInputText, TabInputTextDiff,
-	TabChangeEvent, Event, TabInputCustom, workspace
+	TabChangeEvent, Event, TabInputCustom, workspace,
+	TabInputNotebook
```
dbaeumer, line 12: Small. Please merge with previous line.

client/src/common/diagnostic.ts
```diff
@@ -268,6 +271,16 @@ class Tabs {
 
 	public isVisible(document: TextDocument | Uri): boolean {
 		const uri = document instanceof Uri ? document : document.uri;
+		if (uri.scheme === NotebookDocumentSyncFeature.CellScheme) {
+			// Notebook cells aren't in the list of tabs, but the notebook should be.
+			return Workspace.notebookDocuments.some(notebook => {
```
dbaeumer, line 276: I am a little worried about that code, especially since it is mainly trigger from a context where we have the notebook itself (e.g. indirectly through the matches method). So it might be worthwhile to introduce a matchNotebookDocument method that takes more information. And then we can call isVisible directly with the notebook document uri.

dbaeumer:
Great work. I highly appreciate the effort. See my comments. I am open for discussion since I might have missed some design choices as well.

dbaeumer (changes requested):

client/src/common/notebook.ts
rchiodo: Callbacks work too. I'll change it.

client/src/common/notebook.ts
rchiodo: Not needed. Leftover code from a different refactoring that I forgot to remove.

client/src/common/diagnostic.ts
rchiodo: It's also used in the diagnostics pull async to verify the diagnostics are for a visible document. If this code isn't here, the diagnostics nevery actually get published.

rchiodo: This code:
```javascript
            if (!this.tabs.isVisible(document)) {
                this.documentStates.unTrack(PullState.document, document);
                return;
            }
```

rchiodo: @dbaeumer do you think there's a different way to handle the isVisible in the pullAsync method?

dbaeumer: Not for pullAsync, unless we change a lot of signatures to pass the cell document together with the notebook document around. But we could add a second optional parameter to isVisible that is the container. If we know it in the context (e.g. for example through match) we could use it. If not and it is a cell document then we can still find the notebook document.

Does that make sense?

rchiodo: Yes makes sense. I'll try that.

rchiodo: Actually I think the way I have match and matchCell makes more sense. matchCell knows it's looking at a cell so it asks isVisible only for the notebookDocument and not the cell itself.

dbaeumer (changes requested):

dbaeumer:
Looks good to me. Only one small request for change.

client/src/common/notebook.ts
```diff
@@ -428,7 +429,14 @@ class NotebookDocumentSyncFeatureProvider implements NotebookDocumentSyncFeature
 	private readonly disposables: vscode.Disposable[];
 	private readonly selector: vscode.DocumentSelector;
 
-	constructor(client: FeatureClient<NotebookDocumentMiddleware, $NotebookDocumentOptions>, options: proto.NotebookDocumentSyncOptions) {
+	constructor(
+		client: FeatureClient<NotebookDocumentMiddleware, $NotebookDocumentOptions>,
+		options: proto.NotebookDocumentSyncOptions,
+		private readonly _onChangeNotificationSent: vscode.EventEmitter<VNotebookDocumentChangeEvent>,
```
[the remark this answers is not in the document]
rchiodo: Certainly, will do.

dbaeumer (approved):

dbaeumer:
@rchiodo thanks again for the great PR.

alexdima (approved):

LLotme (commented):
..

PR #1467, opened 2024-04-24
fix #1451

### LINKED ISSUE #1451: Semantic tokens that are not in ascending order will not be highlighted
I pushed a minimum reproducible [demo](https://github.com/imbant/semantic_highlight_demo)

When the order of semantic tokens provided by the language server is inconsistent with the order in the document character stream, non-ascending semantic tokens will not be highlighted. This will not be the case when using language clients based on the VS Code API, and highlighting will not be affected by token order.

## Phenomenon:

image

The code for providing semantic tokens between the language server and the client is almost exactly the same.
image

## doubt:

No requirement for the order of semantic tokens was found in the [LSP documentation](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/).
Is this a requirement for LSP? Or is it a bug in the `vscode-languageserver/node` npm package?

imbant:
The problem is that when `push()` is called out of order, `_data` does not collect the input token correctly.

So I add a non-delta-encoded data attribute `_dataNonDelta`, which is used when the `push()` function is called in non-character stream order; it is eventually delta-encoded in `build()` or `buildEdits()`.I hope this change doesn't affect any places that previously output `_data`.

If `push()` is called in order, we still use `_data` and it works the same as before.

For the implementation of _sortAndDeltaEncode, I refer to the [vscode source code here](https://github.com/microsoft/vscode/blob/ad0ef25421d4124552932b56e61f26b0d025ad96/src/vs/workbench/api/common/extHostTypes.ts#L3455)

### DIFF
--- a/server/src/common/semanticTokens.ts
+++ b/server/src/common/semanticTokens.ts
@@ -117,7 +117,9 @@ export class SemanticTokensBuilder {
 
 	private _prevLine!: number;
 	private _prevChar!: number;
+	private _dataIsSortedAndDeltaEncoded!: boolean;
 	private _data!: number[];
+	private _dataNonDelta!: number[];
 	private _dataLen!: number;
 
 	private _prevData: number[] | undefined;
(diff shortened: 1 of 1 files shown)


imbant:

server/src/common/semanticTokens.ts
```diff
@@ -159,18 +172,108 @@ export class SemanticTokensBuilder {
...
+			result[dstOffset + 4] = tokenModifiers;
+
+			prevLine = line;
+			prevChar = char;
+		}
+
+		return result;
+	}
+
+	private get _finalDataDelta(): number[] {
```
dbaeumer, line 256: I would make this a get method not a getter

server/src/common/semanticTokens.ts
```diff
@@ -159,18 +172,108 @@ export class SemanticTokensBuilder {
 		return this._id.toString();
 	}
 
+	private static _DeltaDecode(data: number[]): number[] {
```
dbaeumer, line 175: method name should start with lower case

dbaeumer:
Thanks for the PR. Highly appreciated. See my minor review comments.

dbaeumer (changes requested):

imbant:
Thanks for the review. I will resolve it later

imbant:
@dbaeumer Hi. Please see my latest commit. Thanks.

dbaeumer (approved):

chrmarti (approved):

PR #1475, opened 2024-05-17
Save responsePromise before writing message

### LINKED ISSUE #1474: Race condition in sendRequest in jsonrpc
https://github.com/microsoft/vscode-languageserver-node/blob/fc5ed68ccb69737a17b1c3c992b9831457b9dc4a/jsonrpc/src/common/connection.ts#L1449-L1450

In the code linked above, if a response is handled before we get a chance to set the responsePromise on the next line, we will never resolve the request.

The fix should be to set the responsePromise before we await on the message write.

9at8:
Fixes #1474


El-o-heka (approved):
T

PR #1490, opened 2024-06-03
Show the error message from the LSP server

DavyLandman:
I've noticed that if a LSP server reports an error via `ResponseError::message` field, it's a bit of a hit or miss if this is reported to the end user in a popup.

For example [rename reports the error in the dialog](https://github.com/SWAT-engineering/vscode-languageserver-node/blob/1f603f4f73552a30a649667262e5f83532cf223e/client/src/common/rename.ts#L112-L116) while errors in many others parts end up in the [generic handler](https://github.com/SWAT-engineering/vscode-languageserver-node/blob/12f4439aa73c31bfe2d4e44377426c23f2a55c01/client/src/common/client.ts#L2053-L2054)

I've changed this such that both the message from the library is kept, while also reporting the message from the LSP server.

### DIFF
--- a/client/src/common/client.ts
+++ b/client/src/common/client.ts
@@ -1183,12 +1183,15 @@ export abstract class BaseLanguageClient implements FeatureClient<Middleware, La
 			this.outputChannel.appendLine(this.data2String(data));
 		}
 		if (showNotification === 'force' || (showNotification && this._clientOptions.revealOutputChannelOn <= reveal)) {
-			this.showNotificationMessage(type, message);
+			this.showNotificationMessage(type, message, data);
 		}
 	}
 
-	private showNotificationMessage(type: MessageType, message?: string) {
+	private showNotificationMessage(type: MessageType, message?: string, data? : any ) {
 		message = message ?? 'A request has failed. See the output for more information.';
+		if (data?.message) {
+			message += '\n' + data.message;
+		}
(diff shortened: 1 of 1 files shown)


dbaeumer:
Thanks for the PR. Should we use the result from `data2String` which is captured here: https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/client.ts#L1004

That covers more cases than simply an Error with an message.

DavyLandman:
> Thanks for the PR. Should we use the result from `data2String` which is captured here: https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/client.ts#L1004
> 
> That covers more cases than simply an Error with an message.

I was thinking about that as well, but that might show a bit more than needed (as it also prints the jsonRPC/lsp error code)?

But if you prefer, I rewrite it to that.

DavyLandman:
Having reviewed the function again, yes I agree, it's better to reuse that function. The PR has been updated.

DavyLandman:
hi @dbaeumer is there something I can do to help get this merged? It would be a pitty if it misses an upcoming release window, as this helps us better report to users when something crashed, what is the issue, without having them dig through the output window.

dbaeumer (approved):

alexdima (approved):

PR #1527, opened 2024-07-31
Fix lint errors

dbaeumer:


Tyriar (approved):

textDocument/src/test/textdocument.test.ts
```diff
@@ -104,7 +104,7 @@ suite('Text Document Lines Model Validator', () => {
 	});
```
PhantomRwcwride2Zero, line 104: <[  Android-X-Developer.com:/
     android / android test  all of this with ANDROID Object script/ android-X-azure /
Virtual studio/ Powershell/Ajax Powershell and everything else with the test ]>

<[ I've been trying to do the test but it won't download and run for me and I need to get this done if you can Thank you

PhantomRwcwride2Zero (changes requested):

Issue #1555, opened 2024-09-06
Support `isAI` property on `CodeAction`


StellaHuang95:
The Typescript team has [added Copilot suggestions](https://github.com/microsoft/vscode/pull/192602) to various refactors in VS Code. To bring similar functionality to python code, Pylance has enabled the [implement inherited abstract classes with Copilot](https://github.com/microsoft/pyrx/issues/5533) code action through the Github Copilot extension for Python files.

To display the sparkle icon for the code action instead of the regular light bulb icon, the [`isAI` property](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.codeActionAI.d.ts) needs to be set to `true` on the `codeAction`.

![Image](https://github.com/user-attachments/assets/ae7959a4-8632-47c4-95ae-df001258c392)

LSP doesn't support this API yet. Would it be possible for LSP to support the [codeActionAI](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.codeActionAI.d.ts) proposed API?

Is it as simple as adding the conversion logic to `asCodeAction` in [`protocolConverter.js`](https://github.com/microsoft/vscode-languageserver-node/blob/f58f4dff16ad2760028bb1cb95e882de30a1000f/client/src/common/protocolConverter.ts#L1013)? I'd be happy to send a PR if that's all it takes. Thanks!

dbaeumer:
There is actually more to it:

- we need to add this to the spec (https://github.com/microsoft/language-server-protocol)
- find a better name than `isAI`. May may be `llmGenerated`
- make this proposed in the spec since it is still proposed in VS Code

Happy to help if you want to work on a PR for here and https://github.com/microsoft/language-server-protocol

StellaHuang95:
@dbaeumer I sent both PRs, let me know if you have any feedback. Thanks for the help😊

PRs are at:
https://github.com/microsoft/vscode-languageserver-node/pull/1557
https://github.com/microsoft/language-server-protocol/pull/2020

PR #1557, opened 2024-09-13
Support  `llmGenerated` property on `CodeAction`

StellaHuang95:
Addresses https://github.com/microsoft/vscode-languageserver-node/issues/1555

- Added `llmGeneratedSupport` to `CodeActionClientCapabilities`.
- Added `llmGenerated` to `CodeAction`.
- Added conversion logics in `protocolConverter` and `codeConverter`.
- Added the proposed api to the `enabledApiProposals` in `package.json` and the corresponding `.d.ts` file.


dbaeumer:
Looking at the implementation now another boolean here feels wrong. I think we should follow the same approach as we did with other literals and add a `tags` field with `CodeActionTag` (like SymbolTag or CompletionItemTag). This make it homogenous with approaches we use somewhere else.

dbaeumer (changes requested):

types/src/main.ts
```diff
@@ -3446,6 +3446,17 @@ export interface CodeAction {
...
 	data?: LSPAny;
+
+	/**
+ 	 * A boolean flag indicating whether the code action is generated by an
+ 	 * LLM.
+	 *
+	 * This property helps distinguish between code actions suggested by an AI
+	 * model and those generated by static analysis.
+	 *
+	 * @since 3.17.0 - proposed
```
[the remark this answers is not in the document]
StellaHuang95: Sure, updated to use the `CodeActionTag` and the `@since` tag. Please let me know if there's anything else you'd like me to address.

types/src/main.ts
```diff
@@ -3372,6 +3372,26 @@ export type CodeActionDisabled = {
...
 };
 
+/**
+ * Code action tags are extra annotations that tweak the rendering of a code action.
+ *
+ * @since 3.18.0 - proposed
+ */
+export namespace CodeActionTag {
+	/**
+	 * Render a code action as LLM-generated, usually using a sparkle icon.
```
dbaeumer, line 3382: We should not enforce rendering. I would simply say: Marks the code action as LLM-generated.

types/src/main.ts
```diff
@@ -3372,6 +3372,26 @@ export type CodeActionDisabled = {
 	reason: string;
 };
 
+/**
+ * Code action tags are extra annotations that tweak the rendering of a code action.
```
dbaeumer, line 3376: I know we did this with other tags but we should not make the concept rendering specific. So I would say:

Code action tags are extra annotations that tweak the behavior of a code action.


For example a client could decide that LLM generated code actions must always go through a review (diff editor).

client/src/common/client.ts
```diff
@@ -2,6 +2,8 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License. See License.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
+/// <reference path="../../typings/vscode.proposed.codeActionAI.d.ts" />
```
dbaeumer, line 5: Please add this to both convertes and not to the client. It is IMO not necessary here:

dbaeumer (changes requested):
Great work. Only some minor wording tweaks.

dbaeumer:
Let's wait until we have the spec right and then we finish up the coding. Is that OK for you?

StellaHuang95:
I've added the `tagSupport` client capability, let me know if there's anything else. Thanks!

karthiknadig:
As far as I can tell this is enabled for the extension. Marking as verified.

dbaeumer (approved):

dbaeumer:
@StellaHuang95 nice work!

TylerLeonhardt (approved):

PR #1561, opened 2024-09-27
Allow resultReporter to be substituted in provideWorkspaceDiagnostics middleware.

JoeRobich:
By not accepting a `resultReporter` parameter in `provideDiagnostics` the `provideWorkspaceDiagnostics` middleware always used the reporter it was initially passed. This means that diagnostics are always reported without the ability to manipulate them first.

### DIFF
--- a/client-node-tests/src/integration.test.ts
+++ b/client-node-tests/src/integration.test.ts
@@ -1387,13 +1387,18 @@ suite('Client integration', () => {
 		});
 
 		let middlewareCalled: boolean = false;
+		let reporterCalled: boolean = false;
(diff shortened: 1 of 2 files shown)


JoeRobich:
@dbaeumer You were suggested as someone who might could review this. We are looking at allowing C# users to optionally turn information diagnostics into hint diagnostics and this is blocking us.

dbaeumer (approved):

dbaeumer:
@JoeRobich great catch.

mjbvz (approved):

JoeRobich:
@dbaeumer Would it be possible to get a new .next release with this change in it?

Issue #1568, opened 2024-10-10
`ReadableStreamMessageReader.listen` should register `data` handler once per readeable


karthiknadig:
The issue is that if the `listen()` is called multiple times. The `callback` gets overwritten but , we now have two `onData` events registered.

The proposal to change is this, so that `onData` is registered on the reader when the instance is created, and not on `listen`. Also, the dispose removes callback, and not the `onData` handler.

For efficiency, we don't need to process data unless there is a listener on. So, adding a `isEmpty` to the `Emitter`, and using emitter to manage callbacks and cleanup.

dbaeumer:
I actually prefer to not allowing this and throw like coded here: dbaeumer/probable-muskox-peach

Can you explain why you want to call listen n times.

karthiknadig:
I don’t want to call listen multiple times. My worry is since we allowed it. Throwing now may cause crashes. If that is not a risk then, exception option is better.

karthiknadig:
If there is no concern with potentially crashing consumers of that API, then I like direction this implementation is going: https://github.com/microsoft/vscode-languageserver-node/blob/a5a067bd8f1001e2e9a978057b41e331be6062ca/jsonrpc/src/common/messageReader.ts#L202-L217

I do want to suggest that, if the reader is only going to be bound to one readable, then all the events should be hooked up in the constructor. The only thing that the listen needs to do is this:

```ts
	public listen(callback: DataCallback): Disposable {
		if (this.callback !== undefined) {
			throw new Error('Reader can only listen once.');
		}
		this.callback = callback;
		return new Disposable(()=>({this.callback = undefined}));
	}
```

the constructor should be doing this:
```ts
        private disposables = new DisposableStore();
	public constructor(readable: RAL.ReadableStream, options?: RAL.MessageBufferEncoding | MessageReaderOptions) {
		super();
		this.readable = readable;
		this.options = ResolvedMessageReaderOptions.fromOptions(options);
		this.buffer = RAL().messageBuffer.create(this.options.charset);
		this._partialMessageTimeout = 10000;
                this.partialMessageTimer = undefined;
		this.nextMessageLength = -1;
		this.messageToken = 0;
		this.readSemaphore = new Semaphore(1);

                this.disposables.add(this.readable.onData((data: Uint8Array) => {
			if(this.callback){this.onData(data);}
		}));
		this.disposables.add(this.readable.onError((error: any) => {if(this.callback){this.fireError(error);}}));
[...]

PR #1569, opened 2024-10-10
Ensure `ReadableStreamMessageReader` registers `onData` only once

karthiknadig:
Fixes https://github.com/microsoft/vscode-languageserver-node/issues/1568

### DIFF
--- a/jsonrpc/src/common/events.ts
+++ b/jsonrpc/src/common/events.ts
@@ -168,4 +168,8 @@ export class Emitter<T> {
 			this._callbacks = undefined;
 		}
 	}
+
+	public isEmpty(): boolean {
+		return this._callbacks?.isEmpty() ?? true;
+	}
 }
--- a/jsonrpc/src/common/messageReader.ts
+++ b/jsonrpc/src/common/messageReader.ts
@@ -170,7 +170,8 @@ export class ReadableStreamMessageReader extends AbstractMessageReader {
 
 	private readable: RAL.ReadableStream;
 	private options: ResolvedMessageReaderOptions;
-	private callback!: DataCallback;
+	private dataEmitter: Emitter<Message>;
+	private onDataDispose: Disposable;
 
 	private nextMessageLength: number;
 	private messageToken: number;
@@ -188,6 +189,10 @@ export class ReadableStreamMessageReader extends AbstractMessageReader {
 		this.nextMessageLength = -1;
 		this.messageToken = 0;
 		this.readSemaphore = new Semaphore(1);
+		this.dataEmitter = new Emitter();
+		this.onDataDispose = this.readable.onData((data: Uint8Array) => {
+			this.onData(data);
+		});
 	}
 
 	public set partialMessageTimeout(timeout: number) {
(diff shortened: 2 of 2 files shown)


jsonrpc/src/common/messageReader.ts
```diff
@@ -202,13 +215,9 @@ export class ReadableStreamMessageReader extends AbstractMessageReader {
 		this.nextMessageLength = -1;
 		this.messageToken = 0;
 		this.partialMessageTimer = undefined;
-		this.callback = callback;
-		const result = this.readable.onData((data: Uint8Array) => {
-			this.onData(data);
-		});
 		this.readable.onError((error: any) => this.fireError(error));
```
dbaeumer, line 218: What do we do with these. The leaked from the beginning but my internal model was always that you only use a reader once. Since this is a wrapper around a readable stream would an easier fix not simply be to create a new reader and dispose the old one instead of making this a full event emitter.

Can you explain why you want to have two listen calls?

jsonrpc/src/common/messageReader.ts
```diff
@@ -202,13 +215,9 @@ export class ReadableStreamMessageReader extends AbstractMessageReader {
 		this.nextMessageLength = -1;
```
dbaeumer, line 215: This might break the whole data processing on the second listen call. If a partial message is already read that had a full header reseting this will break the message processing.

dbaeumer (changes requested):
See comments.

jsonrpc/src/common/messageReader.ts
karthiknadig: You are correct, this needs to be moved out of the listen. Even if we go with the exception approach we may need to clean up this class regardless.

jsonrpc/src/common/messageReader.ts
karthiknadig: See my comment here for alternative (throw exception option) https://github.com/microsoft/vscode-languageserver-node/issues/1568#issuecomment-2408004113

jsonrpc/src/common/events.ts
```diff
@@ -168,4 +168,8 @@ export class Emitter<T> {
 			this._callbacks = undefined;
 		}
 	}
+
+	public isEmpty(): boolean {
+		return this._callbacks?.isEmpty() ?? true;
+	}
```
Benz1993com, line 174: Ok

jsonrpc/src/common/events.ts
```diff
@@ -168,4 +168,8 @@ export class Emitter<T> {
 			this._callbacks = undefined;
 		}
 	}
+
+	public isEmpty(): boolean {
+		return this._callbacks?.isEmpty() ?? true;
```
Benz1993com, line 173: Ok

jsonrpc/src/common/events.ts
```diff
@@ -168,4 +168,8 @@ export class Emitter<T> {
 			this._callbacks = undefined;
 		}
 	}
+
+	public isEmpty(): boolean {
```
Benz1993com, line 172: Ok

Benz1993com (commented):
Ok

jsonrpc/src/common/events.ts
Benz1993com: Ok

Benz1993com (commented):
แก้ไข

Benz1993com (commented):
ตรวจสอบ

Benz1993com (commented):
ทั้งหมด

PR #1591, opened 2024-11-30
Add capability information to the metamodel

MariaSolOs:
Closes https://github.com/microsoft/language-server-protocol/issues/2059.

### DIFF
--- a/protocol/src/common/protocol.callHierarchy.ts
+++ b/protocol/src/common/protocol.callHierarchy.ts
@@ -7,9 +7,10 @@ import { RequestHandler } from 'vscode-jsonrpc';
 import { CallHierarchyItem, CallHierarchyIncomingCall, CallHierarchyOutgoingCall } from 'vscode-languageserver-types';
 
 import { MessageDirection, ProtocolRequestType } from './messages';
-import type {
-	TextDocumentRegistrationOptions, StaticRegistrationOptions, TextDocumentPositionParams, PartialResultParams,
-	WorkDoneProgressParams, WorkDoneProgressOptions
+import {
+	type TextDocumentRegistrationOptions, type StaticRegistrationOptions, type TextDocumentPositionParams, type PartialResultParams,
+	type WorkDoneProgressParams, type WorkDoneProgressOptions,
+	CM
 } from './protocol';
 
 /**
@@ -59,6 +60,7 @@ export namespace CallHierarchyPrepareRequest {
 	export const messageDirection: MessageDirection = MessageDirection.clientToServer;
 	export const type = new ProtocolRequestType<CallHierarchyPrepareParams, CallHierarchyItem[] | null, never, void, CallHierarchyRegistrationOptions>(method);
 	export type HandlerSignature = RequestHandler<CallHierarchyPrepareParams, CallHierarchyItem[] | null, void>;
+	export const capabilities = CM.create('textDocument.callHierarchy', 'callHierarchyProvider');
 }
 
 /**
--- a/protocol/src/common/protocol.configuration.ts
+++ b/protocol/src/common/protocol.configuration.ts
@@ -7,6 +7,7 @@ import { RequestHandler, HandlerResult, CancellationToken } from 'vscode-jsonrpc
 import { LSPAny, URI } from 'vscode-languageserver-types';
 
 import { MessageDirection, ProtocolRequestType } from './messages';
+import { CM } from './protocol';
 
 //---- Get Configuration request ----
 
@@ -25,6 +26,7 @@ export namespace ConfigurationRequest {
 	export const type = new ProtocolRequestType<ConfigurationParams, LSPAny[], never, void, void>(method);
 	export type HandlerSignature = RequestHandler<ConfigurationParams, LSPAny[], void>;
 	export type MiddlewareSignature = (params: ConfigurationParams, token: CancellationToken, next: HandlerSignature) => HandlerResult<LSPAny[], void>;
+	export const capabilities = CM.create('workspace.configuration', undefined);
 }
 
 
--- a/protocol/metaModel.json
+++ b/protocol/metaModel.json
@@ -27,6 +27,8 @@
 				]
 			},
 			"messageDirection": "clientToServer",
+			"clientCapability": "textDocument.implementation",
+			"serverCapability": "implementationProvider",
 			"params": {
 				"kind": "reference",
 				"name": "ImplementationParams"
@@ -80,6 +82,8 @@
 				]
 			},
 			"messageDirection": "clientToServer",
+			"clientCapability": "textDocument.typeDefinition",
+			"serverCapability": "typeDefinitionProvider",
 			"params": {
 				"kind": "reference",
 				"name": "TypeDefinitionParams"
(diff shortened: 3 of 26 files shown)


protocol/src/common/protocol.configuration.ts
```diff
@@ -25,6 +25,7 @@ export namespace ConfigurationRequest {
 	export const type = new ProtocolRequestType<ConfigurationParams, LSPAny[], never, void, void>(method);
 	export type HandlerSignature = RequestHandler<ConfigurationParams, LSPAny[], void>;
 	export type MiddlewareSignature = (params: ConfigurationParams, token: CancellationToken, next: HandlerSignature) => HandlerResult<LSPAny[], void>;
+	export const capabilities: { client: 'workspace.configuration'; server: 'workspace' } = { client: 'workspace.configuration', server: 'workspace' };
```
MariaSolOs, line 28: @dbaeumer The spec doesn't mention a server capability for the `workspace/configuration` request so I'm unsure about this. Since this is requested by the server it seems like there's no server capability needed here?

protocol/src/common/protocol.callHierarchy.ts
```diff
@@ -59,6 +59,7 @@ export namespace CallHierarchyPrepareRequest {
 	export const messageDirection: MessageDirection = MessageDirection.clientToServer;
 	export const type = new ProtocolRequestType<CallHierarchyPrepareParams, CallHierarchyItem[] | null, never, void, CallHierarchyRegistrationOptions>(method);
 	export type HandlerSignature = RequestHandler<CallHierarchyPrepareParams, CallHierarchyItem[] | null, void>;
+	export const capabilities: { client: 'textDocument.callHierarchy'; server: 'callHierarchyProvider' } = { client: 'textDocument.callHierarchy', server: 'callHierarchyProvider' };
```
MariaSolOs, line 62: @dbaeumer Do you think we should add capability information for requests like `callHierarchy/incomingCalls` that don't define their own capabilities and instead depend on whether the server supports other methods (in this case `textDocument/prepareCallHierarchy`)?

protocol/src/common/protocol.notebook.ts
```diff
@@ -375,6 +375,7 @@ export namespace NotebookDocumentSyncRegistrationType {
 	export const method: 'notebookDocument/sync' = 'notebookDocument/sync';
 	export const messageDirection: MessageDirection = MessageDirection.clientToServer;
 	export const type = new RegistrationType<NotebookDocumentSyncRegistrationOptions>(method);
+	export const capabilities: { client: 'notebookDocument.synchronization'; server: 'notebookDocumentSync' } = { client: 'notebookDocument.synchronization', server: 'notebookDocumentSync' };
```
MariaSolOs, line 378: I need some help defining the rest of the notebook capabilities. It isn't very clear from the spec how those are defined.

MariaSolOs:

dbaeumer:
@MariaSolOs just a heads up. We have house keeping month in Dezember. New features have to wait until next year.

MariaSolOs:
> @MariaSolOs just a heads up. We have house keeping month in Dezember. New features have to wait until next year.

Thanks for the heads-up. Although this isn't user facing so I wouldn't consider it a new feature per se, I understand. I'll wait until January then!

MariaSolOs:
@dbaeumer happy new year!

Are you available to take a look at this PR? :)

dbaeumer:
Yes. Will look into it this week.

dbaeumer:
The change looks good to me. The only thing I am wondering is that we have a lot of repetition and if we want to change something we need to use find / replace.  Alternatively we could investigate something like this:

Current
```
export const capabilities: { client: 'textDocument.selectionRange'; server: 'selectionRangeProvider' } = { client: 'textDocument.selectionRange', server: 'selectionRangeProvider' }
```

New
```
type CM<C extends string, S extends string> = { client: C; server: S };
namespace CM {
   export function create<C extends string, S extends string>(client: C, server: S): CM<C, S> {
       return { client, server };
   }
}

export const capabilities: CM<'textDocument.selectionRange', 'selectionRangeProvider'> = CM.create('textDocument.selectionRange', 'selectionRangeProvider');
```

Then if we every want to change something we can either do that inside the type definition / or function or at least can use find all references to see where it is used.

What do you think?

MariaSolOs:
@dbaeumer that does feel like a better design. My only question is where should we define this new `CM` type?

MariaSolOs:
> The change looks good to me. The only thing I am wondering is that we have a lot of repetition and if we want to change something we need to use find / replace. Alternatively we could investigate something like this:
> 
> Current
> 
> ```ts
> export const capabilities: { client: 'textDocument.selectionRange'; server: 'selectionRangeProvider' } = { client: 'textDocument.selectionRange', server: 'selectionRangeProvider' }
> ```
> 
> New
> 
> ```ts
> type CM<C extends string, S extends string> = { client: C; server: S };
> namespace CM {
>    export function create<C extends string, S extends string>(client: C, server: S): CM<C, S> {
>        return { client, server };
>    }
> }
> 
> export const capabilities: CM<'textDocument.selectionRange', 'selectionRangeProvider'> = CM.create('textDocument.selectionRange', 'selectionRangeProvider');
> ```
> 
> Then if we every want to change something we can either do that inside the type definition / or function or at least can use find all references to see where it is used.
> 
> What do you think?

That sounds like a good idea. I'll update the PR.

MariaSolOs:
Hmm @dbaeumer how do you want this solution to handle the cases were only a single capability has been documented? Take the `workspace/configuration` request as an example: There's only the client capability (`workspace.configuration`)

dbaeumer:
We could define the type parameters as `C extends string | undefined` and then use `undefined` to indicate there is no server or client capability. But I have not tested it if it works :-)

MariaSolOs:
> We could define the type parameters as `C extends string | undefined` and then use `undefined` to indicate there is no server or client capability. But I have not tested it if it works :-)

Yeah the thing is that if we do that I think we would lose the literal typing. But I'll play with it.

MariaSolOs:
@dbaeumer I just tried your suggested approach of using `CM.create` to generate these capability objects, but then this complicates the generator script because the initializer node [here](https://github.com/microsoft/vscode-languageserver-node/blob/9dce9a9c3e90a1f379a4f33b1e8fca290ab9e4e4/tools/src/visitor.ts#L600) is now a call expression (and I don't even think we could invoke it to get the output object literal).

dbaeumer:
You don't need to invoke the call expression. You need to look at the arguments of the call expression and then take the arguments and use them in the meta model.

MariaSolOs:
> You don't need to invoke the call expression. You need to look at the arguments of the call expression and then take the arguments and use them in the meta model.

Thank you for the hint. I've updated the code to follow your suggested approach :)

dbaeumer:
Nice work.

dbaeumer (approved):

alexdima (approved):

dbaeumer:
Actually, this broke the build due to a change I did yesterday. The code introduced a cyclic dependency which throws in the browser. Fixed it.

MariaSolOs:
> Actually, this broke the build due to a change I did yesterday. The code introduced a cyclic dependency which throws in the browser. Fixed it.

Oh no, sorry about that! Thank you for the quick fix.

dbaeumer:
@MariaSolOs  Do you know if there is a TS version for this. Cyclic dependencies with type imports are OK just no code imports.

MariaSolOs:
> @MariaSolOs Do you know if there is a TS version for this. Cyclic dependencies with type imports are OK just no code imports.

The rule is compatible with TypeScript, it ignores type-only imports.

Issue #1487, opened 2024-05-29
Extra semanticTokens requests in a multiroot workspace


andyw8:
:wave:

I'm a maintainer on [Ruby LSP](https://github.com/Shopify/ruby-lsp/).

I have a multiroot repo with two apps, `app1` and `app2`. https://github.com/andyw8/multiroot_monorepo

While a file from `app1` is open, the logs show all requests going to `app1`, as expected, **except** for one type of request, `semanticTokens`:

<img width="1027" alt="Screenshot 2024-05-29 at 12 06 39 PM" src="https://github.com/microsoft/vscode-languageserver-node/assets/13400/b8ab4b73-9d1f-475c-b24b-c803b4bd8209">

This happens even if I have no files from `app2` open.

Is there something special about `semanticTokens` that might cause it to behave that way?

(I can provide further information to help troubleshoot, but wanted to ask this up-front in case there's a simple answer).

cc @vinistock

vinistock:
Worth pointing out that the Ruby LSP spawns separate language servers for each configured workspace. Each one of the clients has their document selector configured to only handle requests for the specific workspace they are related to.

All other requests work as expected: each language server only receives requests for files that are inside their workspace. Except for semantic highlighting, which seems to be sent to all servers despite the document selector.

dbaeumer:
Is this still an issue? If so, what document selectors are you using for the different workspace folders?

vinistock:
Yes, it still happens. In fact, we started noticing that code snippets in Copilot Chat also trigger a semantic highlighting request before ever sending a `textDocument/didOpen` notification. It kept erroring for us because the server wasn't aware of that document's existence.

Just recently, we tried to limit our document selector to only the `file` and `git` schemes ([PR](https://github.com/Shopify/ruby-lsp/pull/2889)), but we still see the Copilot Chat window firing semantic highlighting requests.

Our document selector is configured like this (simplified):

```typescript
const selector = [
  // Selectors for documents inside the workspace for `ruby` and `erb`
  { scheme: "file", language: "ruby", pattern: `${workspaceFolder.uri.fsPath}/**/*` },
  { scheme: "git", language: "ruby", pattern: `${workspaceFolder.uri.fsPath}/**/*` },
  { scheme: "file", language: "erb", pattern: `${workspaceFolder.uri.fsPath}/**/*` },
  { scheme: "git", language: "erb", pattern: `${workspaceFolder.uri.fsPath}/**/*` },

[...]

PR #1602, opened 2025-01-27
Log message when overriding client's document selector

vinistock:
Based on this discussion https://github.com/microsoft/vscode-languageserver-node/issues/1487#issuecomment-2601661175

This PR starts logging a message when the client's document selector gets overridden for a feature, in hopes to make this behaviour easier to trace when it occurs.

For the test, I redefined the `info` method so that I could assert that the log is being invoked as expected. Let me know if that pattern is okay.

### DIFF
--- a/client/src/common/features.ts
+++ b/client/src/common/features.ts
@@ -552,6 +552,12 @@ export abstract class TextDocumentLanguageFeature<PO, RO extends TextDocumentReg
 		if (!documentSelector || !capability) {
 			return undefined;
 		}
+
+		const capabilitySelector = capability ?? (capability as any).documentSelector;
+		if (documentSelector && capabilitySelector && JSON.stringify(documentSelector) !== JSON.stringify(capabilitySelector)) {
+			this._client.info(`Overriding client document selector for ${this._registrationType.method}`, capability);
+		}
+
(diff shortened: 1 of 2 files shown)


client/src/common/features.ts
```diff
@@ -552,6 +552,11 @@ export abstract class TextDocumentLanguageFeature<PO, RO extends TextDocumentReg
 		if (!documentSelector || !capability) {
 			return undefined;
 		}
+
+		if (documentSelector && capability) {
```
dbaeumer, line 556: We should only log something if the document selectors are different. Specifying capabilities is fine since there are no client default once. In addition we should somehow log for which feature this is happening.

dbaeumer (changes requested):

client/src/common/features.ts
vinistock: Since both `documentSelector` and `capability` are objects, I ended up going with comparing their JSON stringified versions. I don't really like that approach, so let me know if you have a preferred way of comparing the selectors.

There are a few possible types for selectors. If you prefer, we can definitely add checks for each possible combination.

I also added the method name through `this._registrationType.method`.

PR #1603, opened 2025-01-28
Detached fix

nkomonen-amazon:
## Background Info

- In Unix-like environments, the child process would sometimes not exit even when the parent did.
- The exit/watchdog timers were created due to address this behavior. See the docstring of setupExitTimer().
- The 'detached' flag would not be honored in Unix-like environments, and the process to stay running in some cases.

## Problem

If a language server was set as `detached` in the ServerOptions, and we wanted it to not exit on parent exit, it would
still exit.

## Solution

Commits in this PR have been split up in to smaller concerns, see their comments for specific changes.

Summary is that when the detached flag is set in the ServerOptions, it will now ensure that the Server actually stays alive when the parent exits.

## Testing

- Wasn't able to find an easy way to write a test for behavior when the parent (client) terminates and that the child continues to run.
- E2E test that the server gets the "detached" flag forwarded to it by the client
- Did manual tests by doing the following:
  - Have the test bed client log the ExtHost PID: `console.log('Client PID: ${process.pid}')`
  - Have the test bed server log its PID, such as a through a notification to the client
  - Run the test bed
  - Configure the `detached` flag in the `ServerOptions`
  - Kill the client PID
  - Observe that the child PID continues running if `detached: true`, otherwise it should not exist if `detached: false`

Signed-off-by: nkomonen-amazon <nkomonen@amazon.com>


nkomonen-amazon:
@microsoft-github-policy-service agree company="Amazon"

server/src/node/main.ts
```diff
@@ -26,8 +26,53 @@ export namespace Files {
...
+	}
+
+	if(process.argv.includes('--detached')) {
+		_isDetached = true;
+
+		// Ignores the default behavior when then parent/child communication
+		// channel (e.g. IPC) disconnects, the child terminates.
+		process.on('disconnect', () => {});
+
+		// Keeps the server alive since node may terminate this process if
```
dbaeumer, line 55: I would not hardcode this here. Could we pass that as a argument to the --detach param.

dbaeumer (changes requested):

server/src/node/main.ts
nkomonen-amazon: Good point, I've added an optional value for `--detached NNNN` to set the timeout in milliseconds.

nkomonen-amazon:
I was able to get a Windows machine, I'm going to do some testing on it

nkomonen-amazon:
Hey @dbaeumer would you be able to give this another look

server/src/node/main.ts
```diff
@@ -26,8 +27,60 @@ export namespace Files {
...
+ *
+ * @description
+ * When a process is detached, it will continue running even if it's parent process
+ * terminates. **But EXIT notifications from the client will be honored regardless.**
+ *
+ * @remarks
+ * This function assumes the parent process has been properly configured with
+ * necessary settings (e.g., using `child_process.unref()`).
+ */
+function isDetached() {
```
dbaeumer, line 43: Small things: declare as return a `boolean`

server/src/node/main.ts
```diff
@@ -26,8 +27,60 @@ export namespace Files {
...
+		const timeoutValue = args['detached'];
+		const timeoutMillis = Number(timeoutValue);
+		if (timeoutValue !== undefined && isNaN(timeoutMillis)) {
+			throw Error(`Cli value for flag '--detached' was not a number: ${timeoutMillis}`);
+		}
+		// Keeps the server alive since node may terminate this process if
+		// the parent terminates + this process has nothing in the event loop.
+		setInterval(() => {
+			// Check if the detached server has reached the user-specified lifetime
+			if (_protocolConnectionEndTime && timeoutMillis && Date.now() - _protocolConnectionEndTime >= timeoutMillis) {
```
dbaeumer, line 69: Although not happening here I usually try to make undefined checks explicit especially if the variable as a or type that can falsify. 

`_protocolConnectionEndTime !== undefined`

server/package.json
```diff
@@ -24,6 +24,10 @@
 		"./browser": {
 			"types": "./lib/browser/main.d.ts",
 			"browser": "./lib/browser/main.js"
+		},
+		"./utils": {
```
dbaeumer, line 28: Why do we need this. It will export this from the npm package and will force use to keep it stable as API.

dbaeumer (changes requested):
Great work. Only small things.

PR #1614, opened 2025-02-17
fix: avoid dispose unmatched handlers

yf-yang:
When multiple handlers with the same type is registered for json rpc request/notification/progress, only the last handler will be kept. However, calls of previous registration's returned disposable will unexpectedly remove the registered handler.


dbaeumer (dismissed):

dbaeumer:
Nice work

alexr00 (dismissed):

yf-yang:
Forgot to have star handlers included

yf-yang:

dbaeumer (approved):

alexr00 (approved):

PR #1627, opened 2025-05-02
Add client capability to `workspace/didChangeWorkspaceFolders` notification

MariaSolOs:
Closes https://github.com/microsoft/language-server-protocol/issues/2133


dbaeumer:
@MariaSolOs can you explain why you think this is necessary. The assumption was always that if a client has support for workspace folders it also supports change notifications. If we add the client capability it might break existing servers since they instead of a `boolean` receive a object literal.

MariaSolOs:
> @MariaSolOs can you explain why you think this is necessary. The assumption was always that if a client has support for workspace folders it also supports change notifications. If we add the client capability it might break existing servers since they instead of a `boolean` receive a object literal.

My main motivation here is being able to announce client support for dynamically registering to receive `workspace/didChangeWorkspaceFolders` notifications. This aligns with the client capabilities of other requests (e.g. [`CodeLensClientCapabilities`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeLensClientCapabilities)).

dbaeumer:
I think we should still be able to allow for registering without breaking servers. We could offer a different route for this if it is necessary.

Dozamecik (approved):
gh pr checkout 1627

PR #1644, opened 2025-06-18
Use `MapIterator` return types in `LinkedMap` methods.

DanielRosenwasser:
If you try to consume (or build) this library with the latest `lib`, you'll get the following error from TypeScript:

```
node_modules/.pnpm/vscode-jsonrpc@8.2.0/node_modules/vscode-jsonrpc/lib/common/linkedMap.d.ts:29:5 - error TS2416: Property 'keys' in type 'LinkedMap<K, V>' is not assignable to the same property in base type 'Map<K, V>'.
  Type '() => IterableIterator<K>' is not assignable to type '() => MapIterator<K>'.
    Type 'IterableIterator<K>' is missing the following properties from type 'MapIterator<K>': [Symbol.dispose], map, filter, take, and 9 more.

29     keys(): IterableIterator<K>;
       ~~~~
```

This is because `Map` now returns a ["richer" iterator object](https://github.com/tc39/proposal-iterator-helpers) with several methods, but `LinkedMap` just returns its own bespoke `IterableIterator`s.

The workaround I've made here is to just return an initialized generator (which currently has the same methods as any iterator from a `Map`). The declared type is a bit of a lie, but for most people this is ideal for future-proofing.

This does cause some divergences in behavior - namely, `next()` only throws once if the underlying collection has changed.

### DIFF
--- a/jsonrpc/src/common/linkedMap.ts
+++ b/jsonrpc/src/common/linkedMap.ts
@@ -162,53 +162,51 @@ export class LinkedMap<K, V> implements Map<K, V> {
 		}
 	}
 
-	public keys(): IterableIterator<K> {
-		const state = this._state;
-		let current = this._head;
-		const iterator: IterableIterator<K> = {
-			[Symbol.iterator]: () => {
-				return iterator;
-			},
-			next: (): IteratorResult<K> => {
-				if (this._state !== state) {
-					throw new Error(`LinkedMap got modified during iteration.`);
-				}
-				if (current) {
-					const result = { value: current.key, done: false };
-					current = current.next;
-					return result;
-				} else {
-					return { value: undefined, done: true };
-				}
-			}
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
Alternatively, we can just remove `implements Map`.

dbaeumer:
@DanielRosenwasser thanks for the changes. I would like to keep the `implements Map`. Removing it might break clients that assign it to a Map.

tamird:
Would it be possible to cut a release of vscode-jsonrpc containing this fix?

dbaeumer:
I was OOO and I want to see if I can not implement Map and see what the consequences are.

dbaeumer:
I decided to remove Map as an implement clause due to comment from @rbuckton. We have a major version change anyways so it is fine to break.

PR #1652, opened 2025-07-12
protocol: Fix typo in WorkspaceCapabilities.diagnostic field name

osiewicz:
This commit fixes a typo in the name of `diagnostic` field on client's workspace capabilities. Put another way, it removes the trailing `s` from the name of `diagnostic` field to align it with the LSP spec.

This turned out to be an issue with tsgo, which is generating it's JSON deserialization based on metaModel.json. This then led to it being incompatible with Zed, which is a non-VSC client.

I do not know the policy of this package with regards to breaking changes, but there's at least one language server (rust-analyzer) that is already expecting the name of this field to be `diagnostic`. 

Related:
- https://github.com/microsoft/typescript-go/issues/1396
- https://github.com/zed-extensions/tsgo/issues/6

### DIFF
--- a/client/src/common/diagnostic.ts
+++ b/client/src/common/diagnostic.ts
@@ -1066,7 +1066,7 @@ export class DiagnosticFeature extends TextDocumentLanguageFeature<DiagnosticOpt
 		// the active editor.
 		capability.relatedDocumentSupport = false;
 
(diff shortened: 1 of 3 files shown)


osiewicz:
@microsoft-github-policy-service agree company="Zed Industries"

PR #1660, opened 2025-07-27
Add capability information to `textDocument/colorPresentation`

MariaSolOs:
Adding the same capability information as `textDocument/documentColor` as this is the corresponding resolve request.


MariaSolOs:
@dbaeumer could you please take a quick look at this when you get a minute? :)

dbaeumer (approved):

PR #1663, opened 2025-08-08
Remove implements Map from LinkedMap.

dbaeumer:


MariaSolOs:
@dbaeumer when do you expect to release the next major version of `vscode-jsonrpc` that contains these changes? Right now only `next` releases are available but it would be nice for this to hit the "stable" track.

dbaeumer:
Since it is breaking from a typing perspective it is hard to backport to a stable release.

Issue #1681, opened 2025-10-27
delayOpenNotifications results in `didClose` notifications with no matching `didOpen` notifications


DanTup:
I was testing out vNext of the LSP client with `delayOpenNotifications: true`, however I found that it's suppressing the `didOpen` notification without also suppressing the `didClose` notification.

Eg. if I ctrl+hover something but don't click it, my log looks like:

```text
[3:20:20 PM] [Analyzer] [Info] ==> {"jsonrpc":"2.0","id":16,"method":"textDocument/hover","params":{"textDocument":{"uri":"file:///D%3A/Dev/Test%20Projects/lsp_vnext/lib/lsp_vnext.dart"},"position":{"line":2,"character":4}},"clientRequestTime":1761578420910}
[3:20:20 PM] [Analyzer] [Info] <== {"id":16,"jsonrpc":"2.0","result":{"contents":{"kind":"markdown","value":"(SNIP)"},"range":{"end":{"character":8,"line":2},"start":{"character":0,"line":2}}}}
[3:20:22 PM] [Analyzer] [Info] ==> {"jsonrpc":"2.0","id":17,"method":"textDocument/definition","params":{"textDocument":{"uri":"file:///D%3A/Dev/Test%20Projects/lsp_vnext/lib/lsp_vnext.dart"},"position":{"line":2,"character":4}},"clientRequestTime":1761578422073}
[3:20:22 PM] [Analyzer] [Info] <== {"id":17,"jsonrpc":"2.0","result":[{"originSelectionRange":{"end":{"character":8,"line":2},"start":{"character":0,"line":2}},"targetRange":{"end":{"character":1,"line":45},"start":{"character":0,"line":39}},"targetSelectionRange":{"end":{"character":23,"line":39},"start":{"character":15,"line":39}},"targetUri":"file:///D:/Tools/Dart/Latest%20(Bleeding%20Edge)/lib/async/future.dart"}]}
# This is unexpected
[3:20:22 PM] [Analyzer] [Info] ==> {"jsonrpc":"2.0","method":"textDocument/didClose","params":{"textDocument":{"uri":"file:///D%3A/Tools/Dart/Latest%20%28Bleeding%20Edge%29/lib/async/future.dart"}},"clientRequestTime":1761578422092}
```

The spec says:

> A close notification requires a previous open notification to be sent

DanTup:
I've opened a PR for this at https://github.com/microsoft/vscode-languageserver-node/pull/1682

PR #1682, opened 2025-10-28
Drop textDocument/didClose if dropping textDocument/didOpen

DanTup:
When using delayOpenNotifications=true, the open notifications are queued and dropped if a close notification immediately follows. However, the original implementation still sent the close notification. This does not match the spec:

> A close notification requires a previous open notification to be sent

This change also drops the close notification if the open notification was dropped.

Fixes https://github.com/microsoft/vscode-languageserver-node/issues/1681

cc @dbaeumer


chrmarti (approved):

dbaeumer:
Yes, will ship a new next version end of this week.

DanTup:
Awesome, thank you! 🙂

PR #1691, opened 2025-11-20
Use NoInfer for better typing

### LINKED ISSUE #1685: Using NoInfer for better typing
There are some generic methods that accept parameter named `type`, which type defines types of request handler, including parameter types and result type.

Here is an example of such method:
```ts
onRequest<P, R, PR, E, RO>(type: ProtocolRequestType<P, R, PR, E, RO>, handler: RequestHandler<P, R, E>): Disposable
```

TypeScript compiler infers generic arguments from both the `type` and `handler` parameters. This can result in the inferred types being wider than what is allowed by `ProtocolRequestType`.

For example, the following code compiles but should not, because the return type of the handler, `ls.Hover | undefined | string`, is not assignable to the type `Hover | null` specified in `ls.HoverRequest.type`.
```ts
const connection: Connection = null as any;
const handler: (params: HoverParams) => Promise<Hover | undefined | string> = null as any;
connection.onRequest(HoverRequest.type, handler); // no error
```

If we wrap type of the `handler` parameter in `NoInfer`, TypeScript will not infer `onRequest` method generic arguments from `handler` parameter and will correctly report an error.

Here is another example:
```ts
const params: HoverParams | string = null as any;
await connection.sendRequest(HoverRequest.type, params); // no error
```

It’s also necessary to adjust the `HandlerResult` type to allow returning `undefined` wherever returning `null` is permitted, since the library code automatically replaces all `undefined` values with `null`.
This can be done as follows:
```ts
export type HandlerResult<R, E, R_ = R extends null ? (R | undefined) : R> = R_ | ResponseError<E> | Thenable<R_> | Thenable<ResponseError<E>> | Thenable<R_ | ResponseError<E>>
```

After applying these changes locally, I was able to find several incorrect return types in my request handlers.

If you agree with the proposed changes, I would like to make a PR implementing them.

itrapashko:
Fixes #1685

After adding `NoInfer` to types of parameters, several compilation errors appeared in the `Connection` interface in the methods that register request handlers for `DocumentLinkResolve`, `DocumentColor`, and `ColorPresentation`. The reason for them is that the return type of the request specified in those methods did not match the return type specified in the corresponding `type` variables and in the protocol specification. The return type does not allow `null` and `undefined`, so I removed those from the types.


itrapashko:

dbaeumer:
To keep things consistent with other requests that can return an array I would suggest the following:

- keep `undefined | null` on the handler registration. The server implementation converts `undefined` to `null` for the wire.
- change the `type` definition to include `null`. `undefined` needs to be left out since it is not a valid JSON type.

dbaeumer:
I like the change of the `RequestParam<P> = P extends null ? P | undefined : P;`

dbaeumer (approved):

PR #1706, opened 2025-12-31
Docs: Fix broken security vulnerability definition link

anzulaqeel-anunzio:
Addresses a broken link in \SECURITY.md\. The previous link to the Technet archive was returning a 404 error. I have updated it to the current MSRC Coordinated Vulnerability Disclosure (CVD) URL (https://www.microsoft.com/en-us/msrc/cvd), which is the authoritative source referenced in the same document. Verified the new link resolves correctly.


dbaeumer:
See https://github.com/microsoft/vscode-languageserver-node/pull/1708

PR #1748, opened 2026-03-28
Fix LanguageClient.prepareRename() support for `{ defaultBehavior: true }`

ndonfris:
* /client behavior no longer converts `{ defaultBehavior: true }` to `null`, to now match >= 3.16.0 specification

  - converts `{ defaultBehavior: true }` into the editor's default word range, allowing `textDocument/rename` requests without widening the VS Code provider type
  - continues to reject `{ defaultBehavior: false }`, preventing `textDocument/rename` requests

* /client-node-tests new related `prepareRename` tests

  - adds test server coverage for a controlled `{ defaultBehavior: true }` prepareRename response
  - adds a dedicated integration test asserting the client converts that server result into a vscode.Range while keeping the existing Rename test focused on the standard flow

### DIFF
--- a/client-node-tests/src/integration.test.ts
+++ b/client-node-tests/src/integration.test.ts
@@ -714,6 +714,20 @@ suite('Client integration', () => {
 		assert.strictEqual(middlewareCalled, 2);
 	});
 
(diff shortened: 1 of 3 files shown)


ndonfris:

PR #1750, opened 2026-04-01
add client and jsonrpc esm targets

alexander-heimbuch:
Adds esm target to `jsonrpc` and `client`.


alexander-heimbuch:
@microsoft-github-policy-service agree company="osdyne"

dbaeumer:
We are currently discussing what to do here and the current plan is to make one last official version with CommonJS and then switch completely over to ESM. Having to suppport both is simply not somehing we want to do.

PR #1755, opened 2026-04-22
Snapshot TextDocuments when delayOpenNotifications=true to avoid serving updated info to middleware/server

### LINKED ISSUE #1695: delayOpenNotifications=true results in the wrong document version and content being sent in didOpen notification
I enabled `delayOpenNotifications` in the pre-release version of my extension but [a bug](https://github.com/dart-lang/sdk/issues/62209) came up that it was corrupting the servers overlays for open files in some cases.

I was able to reproduce this, and it seems to happen when code actions create new files with contents. The delayed `didOpen` notification gets sent with the "updated" content/version of the file instead of what it would've been had the notification not been delayed.

For example given a code action that has:

```js
"edit": {
    "documentChanges": [
        {
            "kind": "create",
            "uri": "file:///D:/foo/foo.dart"
        },
        {
            "edits": [
                {
                    "newText": "part of 'sort_declarations.dart';\r\n\r\n",
                    "range": { "end": { "character": 0, "line": 0 }, "start": { "character": 0, "line": 0 } }
                }
            ],
            "textDocument": { "uri": "file:///D:/foo/foo.dart", "version": null }
        }
    ]
},
```

Without delayed notifications, this results in the following LSP traffic, which opens the file (empty) and then applies the edit:

```
==> {"method":"textDocument/didOpen","params":{"textDocument":{"uri":"file:///D%3A/foo/foo.dart","languageId":"dart","version":1,"text":""}}}
==> {"method":"textDocument/didChange","params":{"textDocument":{"uri":"file:///D%3A/foo/foo.dart","version":2},"contentChanges":[{"range":{"start":{"line":0,"character":0},"end":{"line":0,"character":0}},"rangeLength":0,"text":"part of 'sort_declarations.dart';\r\n\r\n"}]}}
```

However, with delayed notifications enabled, we get a `didOpen` that has the modified content in it, and then the `didChange` notification re-inserts the same text, resulting in duplication of content in the servers version of the file:

```
==> {"method":"textDocument/didOpen","params":{"textDocument":{"uri":"file:///D%3A/foo/foo.dart","languageId":"dart","version":2,"text":"part of 'sort_declarations.dart';\r\n\r\n"}}}
==> {"method":"textDocument/didChange","params":{"textDocument":{"uri":"file:///D%3A/foo/foo.dart","version":2},"contentChanges":[{"range":{"start":{"line":0,"character":0},"end":{"line":0,"character":0}},"rangeLength":0,"text":"part of 'sort_declarations.dart';\r\n\r\n"}]}}
```

Image

DanTup:
When delayOpenNotifications is enabled and an open notification is delayed, we cannot use the live VS Code TextDocument to pass to middleware or to the server because it may have updated version/content info.

Instead, capture a snapshot of the document at the time the original notification would've been sent, and then when the notification will be sent, provide that snapshot to the middleware and build the server parameters from it.

Fixes https://github.com/microsoft/vscode-languageserver-node/issues/1695


client/src/common/textSynchronization.ts
```diff
@@ -632,4 +641,175 @@ export class DidSaveTextDocumentFeature extends TextDocumentEventFeature<DidSave
 	protected getTextDocument(data: TextDocument): TextDocument {
 		return data;
 	}
-}
\ No newline at end of file
+}
+
+class TextDocumentSnapshot implements TextDocument {
```
DanTup, line 646: @dbaeumer the implementation here was written by GPT-5.4 with some tidying up by me. I don't like that so much is repeated here from VS Code (and I haven't tried to verify the behaviour matches perfectly, because I'm hoping you might have a better idea :-))

I did add 'vscode-languageserver-textdocument' but it only provided a small number of the methods we needed.

I also noticed there is a very similar class here, but it uses some additional types from VS Code so I couldn't just lift it:

https://github.com/microsoft/vscode/blob/a14d0306921f937f5644bb89bfbb66c699e66409/extensions/copilot/src/platform/editing/common/textDocumentSnapshot.ts

I wonder if it'd be better adding a good implementation of this to vscode-languageserver-textdocument and then using it both in Copilot and here?

client/src/common/textSynchronization.ts
```diff
@@ -632,4 +641,175 @@ export class DidSaveTextDocumentFeature extends TextDocumentEventFeature<DidSave
...
+
+	private static readonly DefaultWordRegExp: RegExp = /(-?\d*\.\d\w*)|([^\s`~!@#%^&*()=+[\]{}\\|;:'",.<>/?-]+)/g;
```
dbaeumer, line 648: Question: did you copy this from somewhere?

client/src/common/textSynchronization.ts
```diff
@@ -632,4 +641,175 @@ export class DidSaveTextDocumentFeature extends TextDocumentEventFeature<DidSave
...
+	public get languageId(): string {
+		return this.snapshotDocument.languageId;
+	}
+
+	public get version(): number {
+		return this.snapshotDocument.version;
+	}
+
+	public save(): Thenable<boolean> {
+		return this.liveVsCodeDocument.save();
```
dbaeumer: I would only forward this to the live document if both documents have the same version. If not I would return false, indicating that the save didn't happen

dbaeumer (changes requested):

dbaeumer:
@DanTup nice work.

client/src/common/textSynchronization.ts
DanTup: No, as noted above - most of the implementation here was written by GPT 5.4. Although TypeScript has a very similar class here:

https://github.com/microsoft/vscode/blob/a14d0306921f937f5644bb89bfbb66c699e66409/extensions/copilot/src/platform/editing/common/textDocumentSnapshot.ts

It's possible that this was in its training (it uses `DEFAULT_WORD_REGEXP` imported from the VS Code source).

I don't like having all this code here, and since TS has a version too, but would really be nice if they could be shared in some way. This feels a bit fragile (and I have not fully reviewed the generated code here yet, because I was hoping you had a better idea 😄)

DanTup: I think this might also be incorrect depending on the language?

dbaeumer:
@DanTup I am planning to do a new stable LSP release soon and I would like to include this PR since the implementation makes sense to me. Could you address the remaining issues I flaged?

DanTup:
@dbaeumer yes, I'll fix that - but do you have any concerns about the generated code or hard-coded regex noted above?

client/src/common/textSynchronization.ts
DanTup: Done!

dbaeumer:
Mo concerns about the code or the regexp. I only wanted to know where the RegExp is coming from. If copied from somehwere would be good to left a pointer

DanTup:
Ok, sgtm as long as you're happy with that generated class. I do think it'd be nice to have a central "official" implementation (since TS is doing it too), but that could always be done in future.

I realise I didn't add any test for the `save()` change I made - let me know if you want me to add that before you merge this (I don't know if you'll need to restart the bots, I already pushed that one change since you started it).

dbaeumer:
@danTup I created https://github.com/microsoft/vscode-languageserver-node/pull/1773

It adds some basic method to the npm module to avoid code duplication at least on the LSP side.

DanTup:
@dbaeumer sgtm, thanks!

PR #1773, opened 2026-05-22
Implement TextDocumentSnapshot for delay open notifications

dbaeumer:
Introduce `TextDocumentSnapshot` to enhance the delay open notification functionality, ensuring accurate handling of document states and content.

### DIFF
--- a/build/bin/symlink.js
+++ b/build/bin/symlink.js
@@ -33,6 +33,7 @@ const root = path.dirname(path.dirname(__dirname));
 	await ln.tryLinkJsonRpc(clientFolder);
 	await ln.tryLinkTypes(clientFolder);
 	await ln.tryLinkProtocol(clientFolder);
+	await ln.tryLinkTextDocuments(clientFolder);
 
 	// test-extension
 	let extensionFolder = path.join(root, 'client-node-tests');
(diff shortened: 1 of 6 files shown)


roblourens (approved):

DanTup:
@dbaeumer FYI, this issue came up this week:

https://github.com/Dart-Code/Dart-Code/issues/6088

The issue is that `didChange` and `didOpen` (in response to a `workspace/applyEdit` from the server) were sent in the wrong order (and the Dart server errors if it gets `didChange` for a file that's not open).

I _suspect_ that it's related to my recent re-enabling or `delayOpenNotifications` (using the latest next release), but I am not certain.

I will let you know if I can track anything down, but wanted to give you a heads up.

I'm starting to wonder if this flag is worth the effort. I really wish it could just be handled in VS Code instead of trying to workaround it here.

dbaeumer:
> I really wish it could just be handled in VS Code instead of trying to workaround it here.

My gut feeling tells me that with all the cases in VS Code itself the code would get equally complicated. The problem is that changing something like this so late is absolutely not easy.

Please keep me in the loop if you find something. And BTW I release 3.18 so you don't need to use next.

DanTup:
> Please keep me in the loop if you find something.

Will do!

> And BTW I release 3.18 so you don't need to use next.

Thanks - I noticed this last week and changed over (and have started on updating the server too) :-)

DanTup:
I haven't managed to repro that issue yet, so I'm going to revert enabling delayOpenNotifications for the upcoming Dart extension release, and I'll try to do some more debugging next month.

PR #1794, opened 2026-06-12
Fix partial result type for textDocument/diagnostic request

dbaeumer:
Update the handling of partial results in the document diagnostic request to allow for both DocumentDiagnosticReport and DocumentDiagnosticReportPartialResult types. This change improves the flexibility of the diagnostic response structure.


deepak1556 (approved):

protocol/metaModel.json
adonovan, line 3: Is it correct that this field still reads 3.17.0 after the 3.18 tag?

PR #1819, opened 2026-07-08
Port metamodel generator to TypeScript 7 API

andrewbranch:
This was done as an exercise/example—we probably don't want to merge it until the TS 7 API is actually stable.

Produces 100% identical output in almost exactly the same amount of time as the TS 6 API, with less setup code.

### DIFF
--- a/tools/src/visitor.ts
+++ b/tools/src/visitor.ts
@@ -3,59 +3,57 @@
  * Licensed under the MIT License. See License.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
 
-import * as ts from 'typescript-7/unstable/sync';
 import {
-	__String,
-	SourceFile,
-	SyntaxKind,
-	Node,
-	NodeArray,
-	ModuleDeclaration,
-	TypeAliasDeclaration,
-	TypeNode,
-	Identifier,
-	TypeReferenceNode,
-	IntersectionTypeNode,
-	TypeLiteralNode,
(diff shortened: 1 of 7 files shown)


tools/src/visitor.ts
```diff
@@ -875,17 +921,17 @@ export default class Visitor {
...
+			const declaration = this.getDeclaration(symbol, SyntaxKind.TypeAliasDeclaration);
+			if (declaration === undefined ||!isTypeAliasDeclaration(declaration)) {
+				throw new Error (`No declaration found for type alias ${symbol.name }`);
 			}
-			if (ts.isTypeLiteralNode(declaration.type)) {
+			if (isTypeLiteralNode(declaration.type)) {
 				// We have a single type literal node. So treat it as a structure
 				const result: Structure = { name: name, properties: [] };
-				this.fillProperties(result, this.typeChecker.getTypeAtLocation(declaration.type).symbol);
+				this.fillProperties(result, this.typeChecker.getTypeAtLocation(declaration.type)!.getSymbol()!);
```
andrewbranch, line 931: At least one of these `!` will go away; the client typings are wrong—I think this is the last thing here I need to fix.

dbaeumer:
@andrewbranch thanks a lot for the PR. I merged it into the https://github.com/microsoft/vscode-languageserver-node/tree/dbaeumer/esm branch since this one convert the LSP code to the new TS compiler and ESM.

PR #1825, opened 2026-07-22
Add support for workspace read requests

msujew:
Related to https://github.com/microsoft/language-server-protocol/issues/1264 (does not fully resolve it, since this PR does not include features to write into a file system - only read from it).

Adds support for the server to read files/directories/stat info from the client.

As indicated by https://github.com/microsoft/language-server-protocol/issues/1264#issuecomment-1493373396, I also thought it'd be best to start with read-only access to the file system. However, the new interfaces/types should be extendable enough to also add write requests if required later on.

Some questions/considerations:
* Right now, if the read/stat fails, the LSP returns `null`. Should it return a response error instead?
* I'm not entirely sure about the properties in `FileSystemClientCapabilities`. Does it make sense to expose each individual request type as an opt-in flag? Or is it enough to provide a `read?: boolean` flag (maybe extend this with a `write?: boolean` flag later on)?
* The `FileType.unknown` value results from the fact that vscode offers `vscode.FileType.unknown` as a possible value to return for `FileStat.type`. Should this be included in the protocol, or should stats/directory entries with this type simply be omitted?
* The LSP has been using *folder* instead of *directory* for the most part. Should this change align to this? I've simply used the same nomenclature as vscode does.

### DIFF
--- a/client/src/common/fileSystem.ts
+++ b/client/src/common/fileSystem.ts
@@ -6,13 +6,14 @@
 import * as vscode from 'vscode';
 
 import {
-	ClientCapabilities, StatRequest, ReadFileRequest, ReadDirectoryRequest
+	ClientCapabilities, StatRequest, ReadFileRequest, ReadDirectoryRequest,
+	ReadFileParamKind
 } from 'vscode-languageserver-protocol';
 
 import { StaticFeature, FeatureClient, FeatureState } from './features';
 
 export interface FileSystemReadFileSignature {
-	(this: void, uri: vscode.Uri, encoding?: string): Promise<string | null>;
+	(this: void, kind: ReadFileParamKind, uri: vscode.Uri, encoding?: string): Promise<string | null>;
 }
 
 export interface FileSystemStatSignature {
@@ -31,7 +32,7 @@ export interface FileSystemReadDirectorySignature {
 export interface FileSystemMiddleware {
 	fs?: {
 		stat?: (this: void, uri: vscode.Uri, next: FileSystemStatSignature) => vscode.ProviderResult<vscode.FileStat | null>;
-		readFile?: (this: void, uri: vscode.Uri, encoding: string | undefined, next: FileSystemReadFileSignature) => vscode.ProviderResult<string | null>;
+		readFile?: (this: void, kind: ReadFileParamKind, uri: vscode.Uri, encoding: string | undefined, next: FileSystemReadFileSignature) => vscode.ProviderResult<string | null>;
 		readDirectory?: (this: void, uri: vscode.Uri, next: FileSystemReadDirectorySignature) => vscode.ProviderResult<[string, vscode.FileType][] | null>;
 	};
 }
@@ -92,24 +93,31 @@ export class FileSystemFeature implements StaticFeature {
 				: null;
 		});
 		client.onRequest(ReadFileRequest.type, async (params) => {
+			const encoding = params.kind === 'text' ? params.encoding : undefined;
 			const paramsUri = this._client.protocol2CodeConverter.asUri(params.uri);
-			const fileRead: FileSystemReadFileSignature = async (uri, encoding) => {
+			const fileRead: FileSystemReadFileSignature = async (kind, uri, encoding) => {
 				try {
 					const bytes = await vscode.workspace.fs.readFile(uri);
-					const decoder = new TextDecoder(encoding || 'utf-8');
-					return decoder.decode(bytes);
+					if (kind === 'binary') {
+						return toBase64(bytes);
+					} else if (kind === 'text') {
+						const decoder = new TextDecoder(encoding || 'utf-8');
+						return decoder.decode(bytes);
+					} else {
+						return null;
+					}
 				} catch {
 					return null;
 				}
 			};
 			const middleware = client.middleware.workspace;
 			const result = await (middleware?.fs?.readFile
-				? middleware.fs.readFile(paramsUri, params.encoding, fileRead)
-				: fileRead(paramsUri, params.encoding));
+				? middleware.fs.readFile(params.kind, paramsUri, encoding, fileRead)
+				: fileRead(params.kind, paramsUri, encoding));
 			if (result === undefined || result === null) {
 				return null;
 			}
-			return { text: result };
+			return { content: result };
 		});
 		client.onRequest(ReadDirectoryRequest.type, async (params) => {
 			const paramsUri = this._client.protocol2CodeConverter.asUri(params.uri);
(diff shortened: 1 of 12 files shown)


msujew:

protocol/src/common/protocol.fileSystem.ts
```diff
@@ -0,0 +1,202 @@
...
+ */
+export interface FileStat {
+	/**
+	 * The type of the file, e.g. is a regular file or a directory.
+	 */
+	type: FileType;
+	/**
+	 * Whether the file is a symbolic link.
+	 */
+	isSymlink: boolean;
```
dbaeumer: Can we make this a flag property with a bit wise implementation. Makes it easier to expand in the future.

protocol/src/common/protocol.fileSystem.ts
```diff
@@ -0,0 +1,202 @@
...
+	 */
+	name: string;
+	/**
+	 * The type of the entry.
+	 */
+	type: FileType;
+	/**
+	 * Whether the entry is a symbolic link.
+	 */
+	isSymlink: boolean;
```
dbaeumer: See above

dbaeumer (changes requested):
Thanks. Very nice PR.

protocol/src/common/protocol.fileSystem.ts
msujew: I thought about combining it with the `type` property, similar to how its done in vscode, but decided against it. Is this what you had in mind?

Bakker-Martijn:
What if we read a file that is binary? I have some library files and we need to obfuscate the source for end-users. 
Because of the way the fileread is implemented it is always decoded using textdecoder. 
```typescript
const fileRead: FileSystemReadFileSignature = async (uri, encoding) => {
	try {
		const bytes = await vscode.workspace.fs.readFile(uri);
		const decoder = new TextDecoder(encoding || 'utf-8');
		return decoder.decode(bytes);
	} catch {
		return null;
	}
};
```
Does it make sense when `encoding === 'buffer'` we return the bytes? Or some other mechanism for this use-case? 

I currently am already testing with this pull request and all works fine. But I have used middleware to overwrite the implementation. (Which is also fine for me)

protocol/src/common/protocol.fileSystem.ts
dbaeumer: @msujew no, I like that they are separate but I would rename isSymlink to flags and have a SymLink flag. If we have more flags in the future it is easier to extend.

msujew: Ok, I believe that's what I've done in https://github.com/microsoft/vscode-languageserver-node/pull/1825/commits/b356b60cc8e00bb6674be8172e4b6e5029db0285 already 👍

msujew:
> Does it make sense when encoding === 'buffer' we return the bytes? Or some other mechanism for this use-case?

@Bakker-Martijn WDYT about `encoding: "base64"` for this and then get the buffer as a base64 string? Otherwise we would need to type the result value as `string | number[]`, with the `number[]` taking 3-4 bytes per byte in the original file, which we would need to pipe through the jsonrpc protocol (a >200% increase). base64 only increases the size by 33%. I'd be open to integrate this into the PR. 

@dbaeumer do you have an opinion on this?

Bakker-Martijn:
@msujew I had the same thought process. But this is not allowed: https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API/Encodings

I believe the `textEncoder` only wants to convert to plain text (not base64 or similar). So this would not resolve my current issue. 

I also notices the size increase. Which is not ideal... 
I could also handle the decoding in the middleware. This way I can still send plain text through the pipe. However, End-users could see the text when looking at the verbose log of the language server. Not sure if I think this is an real issue though.

Just wanted to point this out. It also depends if binary files will be read by other languages servers or not (perhaps I am one of the few). Can imagine not implementing this, having the middleware as fallback is also OK for me. 

--Edit: 
Or do you mean skipping the `textEncoder` when `encoding === "base64"` and converting the `uint8array` to a `base64` string? 
That is also completely fine. I would say that this would indeed make more sense than sending `number[]` (looking at the size increase).

msujew:
> Or do you mean skipping the textEncoder when encoding === "base64" and converting the uint8array to a base64 string?

Yes, exactly. Essentially just special casing the client code and documenting this into the protocol. I.e. servers can request binary file content via `encoding: "base64"`, which clients should respect.

Bakker-Martijn:
That would be perfect for my use-case. I think that is an excellent idea :)

dbaeumer:
I think returning `base64` for binary files is the right way to go. However overloading the encoding property for it feels strange since it normally talks about the encoding of the file on disk not of the encoding of the resulting string. So I would rather do something like this:

```typescript
export enum ReadFileParamKind {
   Text = 'text,
   Binary = 'binary'
}

export interface TextReadFileParams {

 	kind: ReadFileParamKind.Text;

	/**
	 * A URI for the location of the file.
	 */
	uri: DocumentUri;
	/**
	 * The encoding of the file content. If not specified, the content is assumed to be UTF-8.
	 */
	encoding?: string;
}

export interface BinaryReadFileParams {
 	kind: ReadFileParamKind.Binary;

	/**
	 * A URI for the location of the file.
	 */
	uri: DocumentUri;
}

export type ReadFileParams = TextReadFileParams | BinaryReadFileParams;
```

msujew:
> So I would rather do something like this:

Yep, much better. I've replaced the enum with string literal types, since that works better with the meta model generator script. Since there was no base64 conversion yet, I've tried my best with what was available. That should probably do.

dbaeumer:
@msujew the `@since 3.19.0` should also have a `@proposed` tag since we usually give the community time for additional feedback.  Would you also update the specification here: https://github.com/microsoft/language-server-protocol Then I create a 3.19 version as well.

client/src/common/fileSystem.ts
```diff
@@ -0,0 +1,176 @@
...
+			return result
+				? this._client.code2ProtocolConverter.asFileStat(result)
+				: null;
+		});
+		client.onRequest(ReadFileRequest.type, async (params) => {
+			const encoding = params.kind === 'text' ? params.encoding : undefined;
+			const paramsUri = this._client.protocol2CodeConverter.asUri(params.uri);
+			const fileRead: FileSystemReadFileSignature = async (kind, uri, encoding) => {
+				try {
+					const bytes = await vscode.workspace.fs.readFile(uri);
```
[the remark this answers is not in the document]
msujew: I believe this can be ignored? Language servers already can read the whole file system in desktop environments. Using the new LSP API should do no (further) harm.

msujew:
> Would you also update the specification here: https://github.com/microsoft/language-server-protocol Then I create a 3.19 version as well.

Will do, thanks!

PR #1827, opened 2026-07-28
Expose null progress reporter detection

### LINKED ISSUE #847: Allow to determine if `NullProgressMonitor` is used.
See https://github.com/microsoft/pyright/pull/2546

volksec:
## Summary

- export `isNullProgressReporter` from `vscode-languageserver`
- let consumers distinguish a no-op work-done reporter without inspecting its constructor
- replace the server test's `constructor.name` check with the public helper
- cover both real and null work-done reporters

## Motivation

Consumers currently need to rely on implementation details to determine whether a request supplied a real work-done reporter. For example, Pyright compared the reporter constructor with a reporter created by `attachWorkDone`. A public type guard makes this check explicit and keeps `NullProgressReporter` itself private.

This avoids adding a required property to `WorkDoneProgressReporter`, which would break external implementations of that interface.

Fixes #847.

## Validation

- `npm --prefix server run all`
  - TypeScript compilation passed
  - ESLint passed
  - 13 tests passed
- `git diff --check`


volksec:

server/src/common/progress.ts
```diff
@@ -143,6 +143,13 @@ class NullProgressServerReporter extends NullProgressReporter implements WorkDon
 	}
 }
 
+/**
+ * Tests whether the given work done progress reporter discards progress.
+ */
+export function isNullProgressReporter(reporter: WorkDoneProgressReporter): boolean {
```
dbaeumer: I do avoid free floating function. I would prefer having this in an exported namespace WorkDoneProgressReporter right after the interface and name the function isNullInstance

dbaeumer (changes requested):

server/src/common/progress.ts
volksec: Addressed in b360c82f. The helper now lives in an exported `WorkDoneProgressReporter` namespace immediately after the interface and is named `isNullInstance`. I also updated the public export and test usages. The server TypeScript compilation, ESLint, and all 13 server tests pass locally.

server/src/common/progress.ts
```diff
@@ -27,6 +27,15 @@ export interface WorkDoneProgressReporter {
 	done(): void;
 }
 
+export namespace WorkDoneProgressReporter {
+	/**
+	 * Tests whether the given work done progress reporter discards progress.
+	 */
+	export function isNullInstance(reporter: WorkDoneProgressReporter): boolean {
+		return reporter instanceof NullProgressReporter;
```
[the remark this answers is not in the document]
dbaeumer: The change was requested that wat from the code owner

dbaeumer (approved):

amunger (approved):

benibenj (approved):

PR #1846, opened 2026-08-31
Fix socket transport when TransportKind.socket uses port 0

### LINKED ISSUE #1845: `TransportKind.socket` with `port: 0` does not hand the bound port to the server
When configured with `transport: { kind: TransportKind.socket, port: 0 }`, the client spawns the server with `--socket=0`. Passing 0 to `server.listen` makes the OS pick a free port, but the client never reads that bound port back, it just forwards the requested 0 to the server on the command line. The server has no valid port to connect to, so `client.start()` never resolves.

Proposed fix:
Expose a `port()` accessor on `SocketTransport` that returns the actually-bound port (via `server.address()`), and defer pushing `--socket=${transport.port()}` until after `createClientSocketTransport` resolves. Callers passing a specific non-zero port are unaffected, `port()` returns that same port.

sratz:
When the client is configured with `transport: { kind: TransportKind.socket, port: 0 }`, Node's `server.listen(0, ...)` picks a random port. The client was pushing `--socket=0` to the server args before the transport was created, so the server never learned the actual bound port and was unable to connect.

Expose the actually-bound port on SocketTransport via `port(): number`, and defer pushing `--socket=<port>` until after `createClientSocketTransport` resolves in all three spawn paths (runtime, fork, executable command).

Resolves #1845.

### DIFF
--- a/jsonrpc/src/node/main.ts
+++ b/jsonrpc/src/node/main.ts
@@ -229,6 +229,7 @@ export function createServerPipeTransport(pipeName: string, encoding: RAL.Messag
 }
 
 export interface SocketTransport {
+	port(): number;
 	onConnected(): Promise<[MessageReader, MessageWriter]>;
 }
 
@@ -248,7 +249,14 @@ export function createClientSocketTransport(port: number, encoding: RAL.MessageB
 		server.on('error', reject);
 		server.listen(port, '127.0.0.1', () => {
 			server.removeListener('error', reject);
+			const address = server.address();
+			if (address === null || typeof address === 'string') {
+				reject(new Error(`Unexpected server address: ${address}`));
+				return;
+			}
+			const boundPort = address.port;
 			resolve({
+				port: () => boundPort,
 				onConnected: () => { return connected; }
 			});
 		});
--- a/client-node-tests/src/integration.test.ts
+++ b/client-node-tests/src/integration.test.ts
@@ -242,6 +242,23 @@ suite('Server output', () => {
 	});
 });
 
+suite('Socket transport', () => {
+
+	test('Uses an OS-assigned port when transport.port is 0', async () => {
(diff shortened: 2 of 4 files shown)


sratz:
@microsoft-github-policy-service agree company="SAP SE"

jsonrpc/src/node/main.ts
```diff
@@ -248,7 +249,9 @@ export function createClientSocketTransport(port: number, encoding: RAL.MessageB
 		server.on('error', reject);
 		server.listen(port, '127.0.0.1', () => {
 			server.removeListener('error', reject);
+			const boundPort = (server.address() as AddressInfo).port;
```
dbaeumer: Why is this cast necessary. The typing is:

Image

dbaeumer (dismissed):

dbaeumer (changes requested):
See latest comment

jsonrpc/src/node/main.ts
sratz: The cast is for `server.address()`, which according to
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/127fd9871abcf3ff2a66bdc92fc85c77e725a466/types/node/net.d.ts#L680-L706
returns `<AddressInfo> | <string> | <null>`.
`null` only if called before listening, `string` in case of a unix domain socket, and `AddressInfo` in the IP socket case we have here.

dbaeumer: But we should still not cast it without a type check of `!== null && typeof ... !== 'string'` in which case TS infers AddressInfo

sratz: Updated to clearly `reject()` in the unexpected `null | string` case.

dbaeumer (approved):

benibenj (approved):

ulugbekna (approved):

PR #1853, opened 2026-09-08
Route jsonrpc connection logging to the output channel instead of console

### LINKED ISSUE #1613: VSCode language client should configure logger pointing to the output channel rather than console
I recently received a user report of a bug in my language server, but the expected error log did not show in the configured output channel as I expected. Typically I direct users to look into this output channel when reporting issues.

I happened to find the error after I opened `Developer: Toggle Developer Tools` and saw the error log in the developer console, rather than the output channel:

Image

The logged error was a `Received response message without id`, and looking at the code this comes from a call to `logger.error`:

https://github.com/microsoft/vscode-languageserver-node/blob/d810d51297c667bd3a3f46912eb849055beb8b6b/jsonrpc/src/common/connection.ts#L909

Logger is configurable, and after looking around I found that the client configures the logger as a `ConsoleLogger`, which is why it ends up going to the developer console:

https://github.com/microsoft/vscode-languageserver-node/blob/d810d51297c667bd3a3f46912eb849055beb8b6b/client/src/common/client.ts#L2409

What do you think about changing this to log into the vscode output channel rather than the console? I believe this will make the logs more visible for debugging

pujitha24:
Motivation:
BaseLanguageClient's internal jsonrpc Connection was created with a
ConsoleLogger, so protocol-level diagnostics such as "Received response
message without id" (jsonrpc/src/common/connection.ts) only showed up in
VS Code's Developer Tools console. Extensions typically point users at
their own Output channel for troubleshooting, so these messages were
effectively invisible to end users and to extension authors reproducing
bug reports.

Approach:
Add an OutputChannelLogger that implements the jsonrpc Logger interface
(error/warn/info/log) by forwarding to a LogOutputChannel, mapping the
generic log() to outputChannel.info() since LogOutputChannel has no
generic log method. The module-level createConnection() now accepts an
optional lazy accessor (() => LogOutputChannel) and uses
OutputChannelLogger when provided, falling back to the previous
ConsoleLogger otherwise. BaseLanguageClient.createConnection() passes
"() => this.outputChannel" rather than a resolved channel so the output
channel keeps its existing lazy-creation semantics: a channel is still
only created the first time something is actually logged, matching the
behavior of the pre-existing this.error/warn/info/debug client methods
and shouldLogToOutputChannel().

Validation:
- npm run compile:client (tsc -b ./client/tsconfig.json) passes with no
  errors, type-checked against the real @types/vscode LogOutputChannel
  shape.
- cd client && npm run lint (eslint) passes clean.
- Traced call sites in jsonrpc/src/common/connection.ts to confirm
  logger.error/warn/log are only invoked from message-processing paths,
  never from connection setup, so the lazy output-channel accessor is
  never forced eagerly and a session that never logs still creates no
  output channel, preserving prior behavior for that case.
- Could not run this repo's client-node-tests integration suite (real
  VS Code Insiders launched via @vscode/test-electron, as used in
  build/azure-pipelines/darwin/build.yml): the sandbox this change was
  developed in has no free disk space to download the ~300MB VS Code
  build (ENOSPC), an environment constraint unrelated to this change.
  There is no lighter-weight unit-test harness for client.ts in this
  repo since everything under client-node-tests only runs inside a real
  VS Code Electron process.

Report: https://github.com/microsoft/vscode-languageserver-node/issues/1613

Fixes #1613


client/src/common/client.ts
```diff
@@ -2488,8 +2505,8 @@ interface ConnectionCloseHandler {
 	(): void;
 }
 
-function createConnection(input: MessageReader, output: MessageWriter, errorHandler: ConnectionErrorHandler, closeHandler: ConnectionCloseHandler, options?: ConnectionOptions): Connection {
-	const logger = new ConsoleLogger();
+function createConnection(input: MessageReader, output: MessageWriter, errorHandler: ConnectionErrorHandler, closeHandler: ConnectionCloseHandler, getOutputChannel: (() => LogOutputChannel) | undefined, options?: ConnectionOptions): Connection {
```
dbaeumer: Can we make a type / interface ChannelProvider like 

interface ChannelProvider {
     outputChannel: LogOutputChanneln
}

And then pass the client instead of a function instance.

dbaeumer (changes requested):

pujitha24:
@dbaeumer I've pushed changes addressing your review — the branch is now at `93a5c6e` and CI is green. Could you take another look when you have a moment? Happy to keep iterating if anything is still off.

dbaeumer:
@ghost see the Copilot review comments.

pujitha24:
@dbaeumer I've pushed changes addressing your review — the branch is now at `f739ee8` and CI is green. Could you take another look when you have a moment? Happy to keep iterating if anything is still off.

pujitha24:
@dbaeumer I've pushed changes addressing your review — the branch is now at `dbff547` and CI is green. Could you take another look when you have a moment? Happy to keep iterating if anything is still off.

dbaeumer (approved):

pujitha24:
This is approved with green CI and no outstanding comments as far as I can tell — ready whenever you have a moment. Happy to rebase first if you'd like it freshened.

sandy081 (approved):

TylerLeonhardt (approved):

roblourens (approved):

PR #1854, opened 2026-09-08
Include error code and message in the 'Sending response' trace log

### LINKED ISSUE #1604: Trace logs should clearly indicate an error has happened
```
[Trace - 10:09:25 AM] Received request 'client/registerCapability - (1)'.
Params: [
    {
        "id": "abc",
        "method": "textDocument/formatting",
        "registerOptions": {
            "documentSelector": [
                {
                    "language": "helloworld"
                }
            ]
        }
    }
]


[Trace - 10:16:18 AM] Sending response 'client/registerCapability - (1)'. Processing request took 413323ms
```
The trace message makes no indication that an error was sent back and I could not figure out why the `client/registerCapability` was not even being hit. Since the messages were getting logged I knew they were not totally "lost" so I added breakpoints to the `vscode-jsonrpc` code and realized that VS Code did not like how I had formatted my `client/registerCapability` request.
https://github.com/microsoft/vscode-languageserver-node/blob/dadd73f7fc283b4d0adb602adadcf4be16ef3a7b/jsonrpc/src/common/connection.ts#L868-L872
We can then later clearly see that `message.error` is present and bad but the trace message in the console does not include the `-32602` error.
https://github.com/microsoft/vscode-languageserver-node/blob/dadd73f7fc283b4d0adb602adadcf4be16ef3a7b/jsonrpc/src/common/connection.ts#L1076-L1098
Image

It seems to me that we should check for more than just `message.error && message.error.data`. If `message.error.code` and/or `message.error.message` is available then we should log those in the trace too.
```JSON
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32602,
        "message": "Request client/registerCapability defines parameters by name but received parameters by position"
    }
}
```

pujitha24:
Motivation:
When a server sends back a JSON-RPC error response, the
'Sending response' trace line only surfaced error.data (when
present) but silently dropped error.code and error.message. A user
debugging a malformed request spent significant time investigating
because the trace output gave no indication that an error had even
occurred, even though the underlying response clearly carried a
-32602 error with a descriptive message.

Approach:
traceReceivedResponse (the client-side counterpart in the same file)
already appends "Request failed: <message> (<code>)." to its main
log line when message.error is set. Apply the identical pattern to
traceSendingResponse so the server-side trace line carries the same
information. This is a pure logging/string-formatting change with no
behavioral or API impact.

Validation:
- `npx tsc -b ./jsonrpc/tsconfig.json` and `npx tsc -b tsconfig.json`
  (full monorepo) both compile cleanly.
- `npm run lint` (eslint) in jsonrpc/ passes with no new warnings.
- Added a test, "Trace includes error code and message when sending
  an error response", to jsonrpc/src/node/test/connection.test.ts.
  Ran `npm run test:node` in jsonrpc/: 68 passing, including the new
  test. Verified the test fails (times out, since the assertion runs
  inside a rejection-handler `.then()` callback, consistent with this
  file's existing test style) against the pre-fix code and passes
  against the fix.
- Did not run the jsonrpc browser test suite (npm run test:browser)
  in this environment: it requires downloading a Playwright Chrome
  binary, which failed here due to local disk space constraints. The
  browser tests exercise the same shared
  jsonrpc/src/common/connection.ts trace code path, so this is a
  disclosed gap rather than a skipped check.

Report: https://github.com/microsoft/vscode-languageserver-node/issues/1604

Fixes #1604


pujitha24:
Good catch, added the missing period so it now reads `...ms.${error}`, matching `traceReceivedResponse`. Ran the jsonrpc node test suite again (68 passing, including the new trace test) to confirm nothing broke.

dbaeumer (approved):

pujitha24:
This is approved with green CI and no outstanding comments as far as I can tell — ready whenever you have a moment. Happy to rebase first if you'd like it freshened.

sandy081 (approved):

ntrogh (approved):

Issue #1860, opened 2026-09-20
markupMessageSupport is hard-coded to false, what would it take to enable it?


Marvin-Brouwer:
`DiagnosticFeature.fillClientCapabilities` sets:

```ts
// VS Code has no support for markup content in diagnostic messages.
capability.markupMessageSupport = false;
```

(`client/src/common/diagnostic.ts`, [~L1141 on main](https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/diagnostic.ts#L1141))

I want to request functionally to support `MarkupContent` diagnostic messages with unified-language-server (remark-based markdown linting). 
It's messages are authored as markdown at the source, so today things like ``Don't add a trailing `:` to headings`` reach the user with the backticks intact.

Two questions:

1. Is that flag waiting purely on microsoft/vscode#214051 / microsoft/vscode#54272
   landing markdown support in `vscode.Diagnostic.message`, or is there
   something else blocking it on the client side?

2. What should a **push-only** server read? 
   The 3.18 guard is `textDocument.diagnostic.markupMessageSupport`, which is a pull-model
   capability and `PublishDiagnosticsClientCapabilities` has no equivalent. 
   But `asDiagnostic` (`client/src/common/protocolConverter.ts`, [~L329](https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/protocolConverter.ts#L329)) is shared
   between the push and pull paths and switches on the value type rather than the negotiated capability. 
   Zed hit the same thing in zed-industries/zed#61030 and advertises the capability only when pull
   diagnostics are enabled, while handling markup on both paths.

Happy to open a spec issue instead if question 2 belongs upstream.

dbaeumer:
Yes, that flag is waiting on the support in VS Code. 

For push diagnostics we never had client capabilities. I would argue that we should update the spec and say that the property is for both pull and push.

Would you be willing to provide a PR?

Marvin-Brouwer:
Or course, I'm happy to help. 

Should push-only clients set `textDocument.diagnostic.markupMessageSupport`, or would you rather mirror the property under `publishDiagnostics`?

dbaeumer:
I would actually mirror it. Makes it easier to explain an evolve.

Marvin-Brouwer:
I've given it a go, you can see them here:

- Implementation: https://github.com/microsoft/vscode-languageserver-node/pull/1861
[...]

PR #1861, opened 2026-09-21
Markup message support push

Marvin-Brouwer:
Fix for #1860.

3.18 added `MarkupContent` diagnostic messages, guarded by `textDocument.diagnostic.markupMessageSupport`. 
The guarded `message` field is on `Diagnostic`, which both the push and pull models use, but the capability only exists on `DiagnosticClientCapabilities`.
So a push-only server has no way to negotiate it. 

In #1860 @dbaeumer's preference was to mirror the property rather than reword the existing one: 
> "I would actually mirror it. Makes it easier to explain and evolve."

### Notes

- I also added `@proposed` to the pull capability.
  `pullDiagnostics.md` in [the specification](https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/lsp/3.19/language/pullDiagnostics.md?plain=1#L61) has carried it since 3.18, but it was missing here and in the generated model. Happy to drop it if you'd rather keep that change separate.
- The `client.ts` line is not a behaviour change
  The property is optional, so omitting it already means unsupported. It's there so that both capabilities carry the same comment and turn up together when the flag is eventually enabled; otherwise it's easy to flip the pull one and leave push behind.
- `contributing.md` calls a `vscode-languageclient` reference implementation desirable. 
  There isn't one here because VS Code can't render markdown diagnostic messages yet (microsoft/vscode#54272, microsoft/vscode#214051), which is why `markupMessageSupport` is hardcoded `false` on both paths.


Marvin-Brouwer:

protocol/src/common/protocol.diagnostic.ts
```diff
@@ -32,9 +32,11 @@ export type DiagnosticClientCapabilities = DiagnosticsCapabilities & {
 	relatedDocumentSupport?: boolean;
 
 	/**
-	 * Whether the client supports `MarkupContent` in diagnostic messages.
+	 * Whether the client supports `MarkupContent`
+	 * in pull diagnostic messages.
 	 *
 	 * @since 3.18.0
+	 * @proposed
```
dbaeumer: 3.18.0 shipped. We can't mark anything as propsed for 3.18

protocol/metaModel.json
```diff
@@ -9064,8 +9064,12 @@
 							}
 						]
 					},
-					"documentation": "The diagnostic's message. It usually appears in the user interface.\n\n@since 3.18.0 - support for MarkupContent. This is guarded by the client\ncapability `textDocument.diagnostic.markupMessageSupport`.",
-					"since": "3.18.0 - support for MarkupContent. This is guarded by the client\ncapability `textDocument.diagnostic.markupMessageSupport`."
+					"documentation": "The diagnostic's message. It usually appears in the user interface.\n\n@since 3.18.0 - support for MarkupContent in pull diagnostics.\nThis is guarded by the client capability\n`textDocument.diagnostic.markupMessageSupport`.\n\n@since 3.19.0 - support for MarkupContent in push diagnostics.\nThis is guarded by the client capability\n`textDocument.publishDiagnostics.markupMessageSupport`.",
```
dbaeumer: I assume you generate the file using npm run generate:metaModel. Right?

dbaeumer (changes requested):

protocol/src/common/protocol.diagnostic.ts
Marvin-Brouwer: That is absolutely fair, I figured since https://github.com/microsoft/language-server-protocol does have a proposed marking someone might have missed it.
But if we can't do it after the fact I will revert it.
Are you sure though? Since it does really say `@proposed` here:
https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/lsp/3.19/language/pullDiagnostics.md?plain=1#L61

protocol/metaModel.json
Marvin-Brouwer: Yes, I did

protocol/src/common/protocol.diagnostic.ts
Marvin-Brouwer: I see you answered in that PR, I will remove the proposed tag

Marvin-Brouwer: Removed it

PR #1863, opened 2026-09-22
Close pipe/socket transport server when the language server process fails to start

### LINKED ISSUE #1641: Server is not closed when child process not started
https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/node/main.ts#L389
https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/node/main.ts#L402
When error happened on child process start, the server created by `createClientPipeTransport` and `createClientSocketTransport` is not closed.

pujitha24:
Motivation:
When TransportKind.pipe or TransportKind.socket is used, the client first
creates a net.Server via createClientPipeTransport/createClientSocketTransport
(jsonrpc/src/node/main.ts) and starts it listening, then spawns the language
server process, which is expected to connect back to that server. If
cp.spawn() fails synchronously (the returned ChildProcess has no pid), the
client rejects via handleChildProcessStartError but never told the listening
net.Server to close. The server (and, for named pipes, the backing pipe/
socket file) was leaked for the lifetime of the extension host process.

Approach:
Added a dispose() method to the PipeTransport and SocketTransport interfaces
in jsonrpc/src/node/main.ts that closes the underlying net.Server. Call
transport.dispose() at the 4 call sites in client/src/node/main.ts where
cp.spawn's result has no pid (both the NodeModule+runtime branch and the
Executable/command branch, for both pipe and socket transports) before
delegating to handleChildProcessStartError. The 2 cp.fork-based call sites
are intentionally left unchanged: cp.fork always spawns the current Node
executable, which cannot fail in the same synchronous, pid-less way, so
this leak does not apply there.

This is a resource leak fix only (a listening net.Server, and for pipes a
stale named pipe/socket file, being left behind); it does not change any
observable client behavior on the success path, and it is unrelated to the
separate issue of the server process exiting cleanly after spawning but
before connecting.

Validation:
- Added two tests to jsonrpc/src/node/test/connection.test.ts,
  "createClientSocketTransport dispose closes the listening socket before a
  client connects" and "createClientPipeTransport dispose closes the
  listening pipe before a client connects", which assert that a connection
  attempt after dispose() is refused (proving the server actually stopped
  listening, not just that the method exists).
- Ran `npx tsc -b ./tsconfig.json && npm run test:node` in jsonrpc/: 71
  passing, including the 2 new tests.
- Confirmed the new tests fail before the fix: temporarily replaced the
  dispose() implementation with a no-op, reran the suite, and got exactly
  69 passing / 2 failing (the 2 new tests); restored the fix and reran to
  confirm 71 passing again.
- Ran `node ./build/bin/symlink.js` (this repo's documented dev-setup step)
  then `npx tsc -b ./tsconfig.json --force` from the repo root: clean
  across all packages (jsonrpc, protocol, types, textDocument, client,
  server, client-node-tests).
- Ran `npx eslint src/node/main.ts src/node/test/connection.test.ts` in
  jsonrpc/ and `npx eslint src/node/main.ts` in client/: no new warnings
  or errors.
- Did not run client-node-tests' own integration suite (`npm test` there),
  which launches a real VS Code/Electron instance via @vscode/test-electron
  and needs a network download of VS Code; not feasible in this
  environment. This does not exercise the specific spawn-failure code path
  changed here.

Report: https://github.com/microsoft/vscode-languageserver-node/issues/1641

Fixes #1641


jsonrpc/src/node/main.ts
```diff
@@ -257,7 +272,8 @@ export function createClientSocketTransport(port: number, encoding: RAL.MessageB
 			const boundPort = address.port;
 			resolve({
 				port: () => boundPort,
-				onConnected: () => { return connected; }
+				onConnected: () => { return connected; },
+				dispose: () => { server.close(); }
```
dbaeumer: I would call this `close` not `dispose`. It is not a disposable in the classical sense.

dbaeumer (changes requested):

pujitha24:
Makes sense, it just wraps `server.close()`. Renamed it to `close()` on both `PipeTransport` and `SocketTransport` in d94b3905 and updated the call sites in `client/src/node/main.ts` and the tests.

